PR #1, opened 2023-05-23
CASSANDRA-18545: Provide a SecretsProvider interface to abstract the secret provisioning

frankgh:
This commit introduces the SecretsProvider interface that abstracts the secrets provisioning. This way different implementations of the SecretsProvider can be used to provide SSL secrets for the Analytics job. We provide an implementation, SslConficSecretsProvider, which provides secrets based on the configuration for the job.


dineshjoshi (approved):
+1

frankgh:
Merged in https://github.com/apache/cassandra-analytics/commit/b87b0edd310d1ef93c507bbbb1ae51e1b0b319c6

PR #2, opened 2023-05-24
CASSANDRA-18548: Add the .asf.yaml file

frankgh:
This commit adds the .asf.yaml file to control notifications and github settings for the Cassandra Analytics project.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,0 +1,8 @@
+github:
+  enabled_merge_buttons:
+    squash:  false
+    merge:   false
+    rebase:  true
+
+notifications:
+  jira_options: link worklog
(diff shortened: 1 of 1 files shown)


.asf.yaml
```diff
@@ -0,0 +1,8 @@
+github:
+  enabled_merge_buttons:
+    squash:  false
```
yifan-c: If ever wants to allow merge directly from PR, can we enable squash? Because it is a step we do anyway when committing manually.

frankgh: Makes sense, I've updated the PR

yifan-c:
Test notification

PR #7, opened 2023-06-15
CASSANDRA-18605 Adding support for TTL & Timestamps for bulk writes

jyothsnakonisa:
This commit introduces a new feature in Spark Bulk Writer to support writes with constant/per_row based TTL & Timestamps.

### DIFF
--- a/cassandra-analytics-core-example/src/main/java/org/apache/cassandra/spark/example/SampleCassandraJob.java
+++ b/cassandra-analytics-core-example/src/main/java/org/apache/cassandra/spark/example/SampleCassandraJob.java
@@ -141,9 +141,10 @@ public final class SampleCassandraJob
           .option("local_dc", "datacenter1")
           .option("bulk_writer_cl", "LOCAL_QUORUM")
           .option("number_splits", "-1")
-//          .option(WriterOptions.TTL.name(), TTLOption.constant(20))
+          // A constant timestamp and TTL can be used by setting the following options.
+          // .option(WriterOptions.TIMESTAMP.name(), TimestampOption.constant(System.currentTimeMillis() * 1000))
+          // .option(WriterOptions.TTL.name(), TTLOption.constant(20))
           .option(WriterOptions.TTL.name(), TTLOption.perRow("ttl"))
-//          .option(WriterOptions.TIMESTAMP.name(), TimestampOption.constant(System.currentTimeMillis() * 1000))
           .option(WriterOptions.TIMESTAMP.name(), TimestampOption.perRow("timestamp"))
           .mode("append")
           .save();
--- a/cassandra-analytics-core-example/src/main/java/org/apache/cassandra/spark/example/SampleCassandraJob.java
+++ b/cassandra-analytics-core-example/src/main/java/org/apache/cassandra/spark/example/SampleCassandraJob.java
@@ -28,6 +28,9 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.LongStream;
 
+import org.apache.cassandra.spark.bulkwriter.TTLOption;
+import org.apache.cassandra.spark.bulkwriter.TimestampOption;
+import org.apache.cassandra.spark.bulkwriter.WriterOptions;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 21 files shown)


cassandra-analytics-core-example/src/main/java/org/apache/cassandra/spark/example/SampleCassandraJob.java
```diff
@@ -140,6 +141,10 @@ private static Dataset<Row> write(long rowCount, SparkConf sparkConf, SQLContext
           .option("local_dc", "datacenter1")
           .option("bulk_writer_cl", "LOCAL_QUORUM")
           .option("number_splits", "-1")
+//          .option(WriterOptions.TTL.name(), TTLOption.constant(20))
```
yifan-c: Should the comments (and the one below) be removed? If not, can you add comments for the commented code to explain why they are commented.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -168,18 +169,9 @@ private void persist(@NotNull JavaPairRDD<DecoratedKey, Object[]> sortedRDD)
         }
     }
 
-    private static class WriteIterator implements Serializable
+    private static VoidFunction<Iterator<Tuple2<DecoratedKey, Object[]>>> writeRowsInPartition(Broadcast<BulkWriterContext> broadcastContext,
```
yifan-c, line 174: Can you add a comment of declaring `static`? It is to ensure that the lambda does not capture `this` reference, which is not serializable. The comment is useful for maintainers. 
(I think it works w/o `static`, but it is more explicit to have it in the method signature)

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TimestampOption.java
```diff
@@ -0,0 +1,125 @@
...
+
+package org.apache.cassandra.spark.bulkwriter;
+
+import java.io.Serializable;
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
+
+public final class TimestampOption implements Serializable
+{
+    private static final TimestampOption NOW = new TimestampOption(System.nanoTime() / 1000);
```
yifan-c: I do not think you should use it to get the absolute time `now`. See the comment of the method. 

> This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time. 

It should use `System#currentTimeMillis()`, which is used in C* codebase to generate timestamp for cells too.

cassandra-analytics-core-example/src/main/java/org/apache/cassandra/spark/example/SampleCassandraJob.java
jyothsnakonisa: @yifan-c I left them as comments to leave an example of how to use constant Timestamp/ TTL.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TimestampOption.java
jyothsnakonisa: We need to have timestamp in `microseconds` to emulate the Timestamp option in spark connector. I thought Sytem.nanoTime() /1000 would be a more accurate value than System.currentTimeMillis() * 1000.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TTLOption.java
```diff
@@ -0,0 +1,126 @@
...
+     * @return TTLOption
+     */
+    public static String constant(Duration duration)
+    {
+        return String.valueOf(duration.getSeconds());
+    }
+
+    /**
+     * TTL option for writes with TTL per Row. When different TTL has to be used for different rows in a bulk write
+     * call use this option. The RDD should have additional column with TTL values in seconds for each row.
```
yifan-c:
```suggestion
     * call use this option. It expects the input RDD to supply the TTL values as an additional column at each row of the RDD. The TTL value provider column is selected by {@param ttlColumnName}
```

Hopefully, the suggestion makes it a bit more clear.

frankgh (approved):
Nothing to add here. Thanks for the patch!

PR #9, opened 2023-06-27
[CASSANDRA-18633] Added Caching of Node Settings to Improve Efficiency

5:
Currently, `CassandraClusterInfo` does not reuse results of a call to `allNodeSettingsBlocking` from `Sidecar`.
We should improve efficiency by caching these results in order to prevent multiple expensive cluster-wide calls.

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
@@ -23,6 +23,7 @@ import java.io.Closeable;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -64,11 +65,18 @@ public class CassandraClusterInfo implements ClusterInfo, Closeable
     protected transient GossipInfoResponse gossipInfo;
     protected transient CassandraContext cassandraContext;
     protected transient NodeSettings nodeSettings;
+    protected final transient CompletableFuture<List<NodeSettings>> allNodeSettings;
 
     public CassandraClusterInfo(BulkSparkConf conf)
     {
         this.conf = conf;
         this.cassandraContext = buildCassandraContext();
+        this.allNodeSettings = CompletableFuture.supplyAsync(() -> {
+            LOGGER.info("Getting Cassandra versions from all nodes");
+            return Sidecar.allNodeSettingsBlocking(conf,
+                                                   cassandraContext.getSidecarClient(),
+                                                   cassandraContext.clusterConfig);
+        });
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -338,11 +339,23 @@ public String getVersionFromFeature()
...
+        synchronized (this)
+        {
+            if (allNodeSettings == null)
+            {
+                LOGGER.info("Getting Cassandra versions from all nodes");
+                allNodeSettings = Sidecar.allNodeSettingsBlocking(conf,
+                                                                  cassandraContext.getSidecarClient(),
+                                                                  cassandraContext.clusterConfig);
+            }
+            return getLowestVersion(allNodeSettings);
```
yifan-c: I think this line could be moved out of the critical section. 
And further more, it can combine with line#345 and making it the classic double checked locking pattern...

5: Refactored.
(I was trying to keep style consistent with other methods within the same class.)

frankgh: To me, the [original implementation provided in the PR](https://github.com/apache/cassandra-analytics/commit/10725fb9f29076637eb2e9a9206d97287b239bc7) is the correct* classic double checked locking pattern. By some definition of correct

frankgh: I agree that getLowestVersion should be out of the critical section

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -64,11 +65,18 @@ public class CassandraClusterInfo implements ClusterInfo, Closeable
     protected transient GossipInfoResponse gossipInfo;
     protected transient CassandraContext cassandraContext;
     protected transient NodeSettings nodeSettings;
+    protected final transient CompletableFuture<List<NodeSettings>> allNodeSettings;
 
     public CassandraClusterInfo(BulkSparkConf conf)
     {
         this.conf = conf;
         this.cassandraContext = buildCassandraContext();
+        this.allNodeSettings = CompletableFuture.supplyAsync(() -> {
```
yifan-c, line 75: I think it is wasting one thread that is only used to wait for the blocking code to complete. The blocking code is actually a future underlying. `Sidecar.allNodeSettingsBlocking` could be revised to return the future directly, and rename the method.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -338,11 +346,14 @@ public String getVersionFromFeature()
...
-        LOGGER.info("Getting Cassandra versions from all nodes");
-        List<NodeSettings> allNodeSettings = Sidecar.allNodeSettingsBlocking(conf,
-                                                                             cassandraContext.getSidecarClient(),
-                                                                             cassandraContext.clusterConfig);
-        return getLowestVersion(allNodeSettings);
+        try
+        {
+            return getLowestVersion(allNodeSettings.get());
+        }
+        catch (InterruptedException | ExecutionException exception)
```
yifan-c: for `InterruptedException`, it is better to restore the interrupted state for the thread. It might not be a big issue as the actual effect here is to exit the job, which makes the comment just a nit.

dineshjoshi (approved):
+1

PR #13, opened 2023-07-25
CASSANDRA-18692 Fix bulk writes with Buffered RowBufferMode

frankgh:
When setting Buffered RowBufferMode as part of the `WriterOption`s, `org.apache.cassandra.spark.bulkwriter.RecordWriter` ignores that configuration and instead uses the batch size to determine when to finalize an SSTable and start writing a new SSTable, if more rows are available.

In this commit, we fix `org.apache.cassandra.spark.bulkwriter.RecordWriter#checkBatchSize` to take into account the configured `RowBufferMode`. And in specific to the case of the `UNBUFFERED` RowBufferMode, we check then the batchSize of the SSTable during writes, and for the case of `BUFFERED` that check will take no effect.

Co-authored-by: Doug Rohrer <doug@therohrers.org>

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
@@ -41,6 +41,7 @@ import com.google.common.collect.Range;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.cassandra.bridge.RowBufferMode;
 import org.apache.cassandra.sidecar.common.data.TimeSkewResponse;
 import org.apache.spark.InterruptibleIterator;
 import org.apache.spark.TaskContext;
@@ -93,7 +94,7 @@ public class RecordWriter implements Serializable
         TaskContext taskContext = taskContextSupplier.get();
         LOGGER.info("[{}]: Processing Bulk Writer partition", taskContext.partitionId());
         scala.collection.Iterator<scala.Tuple2<DecoratedKey, Object[]>> dataIterator =
-                new InterruptibleIterator<>(taskContext, asScalaIterator(sourceIterator));
+        new InterruptibleIterator<>(taskContext, asScalaIterator(sourceIterator));
         StreamSession streamSession = createStreamSession(taskContext);
         validateAcceptableTimeSkewOrThrow(streamSession.replicas);
         int partitionId = taskContext.partitionId();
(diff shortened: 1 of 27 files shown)


dineshjoshi:
+1 LGTM.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -114,16 +115,21 @@ public StreamResult write(Iterator<Tuple2<DecoratedKey, Object[]>> sourceIterato
                 checkBatchSize(streamSession, partitionId, job);
             }
 
-            if (batchSize != 0)
+            if (sstableWriter != null)
             {
-               finalizeSSTable(streamSession, partitionId, sstableWriter, batchNumber, batchSize);
+                finalizeSSTable(streamSession, partitionId, sstableWriter, batchNumber, batchSize);
```
yifan-c, line 120: nit: reset `sstableWriter` to `null` after `finalizeSSTable`

cassandra-four-zero/src/main/java/org/apache/cassandra/bridge/SSTableWriterImplementation.java
```diff
@@ -80,4 +79,28 @@ public void close() throws IOException
...
+                                                           .forTable(createStatement)
+                                                           .withPartitioner(cassPartitioner)
+                                                           .using(insertStatement);
+        if (rowBufferMode == RowBufferMode.UNBUFFERED)
+        {
+            builder.sorted();
+        }
+        else if (rowBufferMode == RowBufferMode.BUFFERED)
+        {
+            builder.withBufferSizeInMB(bufferSizeMB);
```
yifan-c, line 102: nit: maybe decide a valid value range for `bufferSizeMB` and validate. `CQLSSTableWriter` accepts whatever int value. 
For the upper bound, I think 1/2 of the `spark.executor.memory` is a good limit.

yifan-c (commented):
some nits. Looks good to me.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
JeetKunDoug: Given this is essentially the last step before we're done with the sstablewriter, we don't really need to set the sstablewriter to null here - this RecordWriter instance should be collected as soon as the upload/commits finish, and we close the sstablewriter on finalize so we should be good w/o nulling it out.

cassandra-four-zero/src/main/java/org/apache/cassandra/bridge/SSTableWriterImplementation.java
JeetKunDoug: I think, for now, we leave this just a configuration option... Validating it given the Spark environment and picking a reasonable upper-bound would take some experimentation (and would likely involve more than just the executor.memory setting, as there are other config settings that deal w/ memory like memory overhead).

PR #17, opened 2023-10-12
Cassandra 18852: Make bulk writer resilient to cluster resize events

arjunashok:
Bulk-write resiliency changes to analytics library. 

### Changes:
- Replaces dependency on Sidecar `ring` endpoint to use the `token-ranges` endpoint
- Queries sidecar `token-ranges` endpoint to get the instance metadata, token-ranges and write replica-sets which include the pending ranges for transient nodes.
- Data is written to RF replicas calculated including the transient nodes. 
- RingMonitor and other checks to fail-fast on detecting resizing cluster are removed
-  Ring changes within the scope of a task are checked to allow the task to retry.

### Testing
- Added unit tests and in-jvm-dtest based integration tests for cluster expansion, shrink, node-movement and replacement scenarios
- Validates partitioned data on each recipient instance of the write, including transitioning nodes.

Note: Integration test parameterization for Consistency Levels and Cassandra versions is currently out of scope from this PR. Will be tracked and patched separately. The tests do validate different CL levels currently.

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
@@ -34,6 +34,8 @@ import com.google.common.annotations.VisibleForTesting;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.cassandra.bridge.CassandraBridge;
+import org.apache.cassandra.bridge.CassandraBridgeFactory;
 import org.apache.cassandra.bridge.CassandraVersionFeatures;
 import org.apache.cassandra.clients.Sidecar;
 import org.apache.cassandra.sidecar.client.SidecarInstance;
@@ -52,6 +54,8 @@ import org.apache.cassandra.spark.data.partitioner.Partitioner;
 import org.apache.cassandra.spark.utils.CqlUtils;
 import org.apache.cassandra.spark.utils.FutureUtils;
 import org.jetbrains.annotations.NotNull;
+
+import static org.apache.cassandra.bridge.CassandraBridgeFactory.maybeQuotedIdentifier;
 
 public class CassandraClusterInfo implements ClusterInfo, Closeable
 {
@@ -233,7 +237,9 @@ public class CassandraClusterInfo implements ClusterInfo, Closeable
 
     protected String getCurrentKeyspaceSchema() throws Exception
     {
-        SchemaResponse schemaResponse = getCassandraContext().getSidecarClient().schema(conf.keyspace).get();
+        SchemaResponse schemaResponse = getCassandraContext().getSidecarClient()
+                                                             .schema(maybeQuotedIdentifier(bridge(), conf.quoteIdentifiers, conf.keyspace))
+                                                             .get();
         return schemaResponse.schema();
     }
 
@@ -243,7 +249,11 @@ public class CassandraClusterInfo implements ClusterInfo, Closeable
         RingResponse ringResponse = getCurrentRingResponse();
         List<RingInstance> instances = getSerializableInstances(ringResponse);
         ReplicationFactor replicationFactor = getReplicationFactor();
-        return new CassandraRing<>(getPartitioner(), conf.keyspace, replicationFactor, instances);
+
+        return new CassandraRing<>(getPartitioner(),
+                                   maybeQuotedIdentifier(bridge(), conf.quoteIdentifiers, conf.keyspace),
+                                   replicationFactor,
+                                   instances);
     }
 
     @NotNull
@@ -403,7 +413,9 @@ public class CassandraClusterInfo implements ClusterInfo, Closeable
 
     private RingResponse getCurrentRingResponse() throws Exception
     {
-        return getCassandraContext().getSidecarClient().ring(conf.keyspace).get();
+        return getCassandraContext().getSidecarClient()
+                                    .ring(maybeQuotedIdentifier(bridge(), conf.quoteIdentifiers, conf.keyspace))
+                                    .get();
     }
 
     private static List<RingInstance> getSerializableInstances(RingResponse ringResponse)
@@ -533,6 +545,11 @@ public class CassandraClusterInfo implements ClusterInfo, Closeable
             }
         }
         return false;
+    }
+
+    protected CassandraBridge bridge()
+    {
+        return CassandraBridgeFactory.get(getLowestCassandraVersion());
     }
 
     // Startup Validation
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
@@ -29,12 +29,16 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import com.google.common.base.Preconditions;
-import org.apache.cassandra.spark.data.CqlField;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.cassandra.bridge.CassandraBridge;
+import org.apache.cassandra.bridge.CassandraBridgeFactory;
 import org.apache.cassandra.spark.common.schema.ColumnType;
+import org.apache.cassandra.spark.data.CqlField;
 import org.apache.spark.sql.types.StructType;
+
+import static org.apache.cassandra.bridge.CassandraBridgeFactory.maybeQuotedIdentifier;
 
 public class TableSchema implements Serializable
 {
@@ -49,13 +53,22 @@ public class TableSchema implements Serializable
     private final WriteMode writeMode;
     private final TTLOption ttlOption;
     private final TimestampOption timestampOption;
+    private final String lowestCassandraVersion;
+    private final boolean quoteIdentifiers;
 
-    public TableSchema(StructType dfSchema, TableInfoProvider tableInfo, WriteMode writeMode,
-                       TTLOption ttlOption, TimestampOption timestampOption)
+    public TableSchema(StructType dfSchema,
+                       TableInfoProvider tableInfo,
+                       WriteMode writeMode,
+                       TTLOption ttlOption,
+                       TimestampOption timestampOption,
+                       String lowestCassandraVersion,
+                       boolean quoteIdentifiers)
     {
         this.writeMode = writeMode;
         this.ttlOption = ttlOption;
         this.timestampOption = timestampOption;
+        this.lowestCassandraVersion = lowestCassandraVersion;
+        this.quoteIdentifiers = quoteIdentifiers;
 
         validateDataFrameCompatibility(dfSchema, tableInfo);
         validateNoSecondaryIndexes(tableInfo);
@@ -123,7 +136,7 @@ public class TableSchema implements Serializable
                          CqlField.CqlType cqlType = tableInfo.getColumnType(fieldName);
                          return SqlToCqlTypeConverter.getConverter(cqlType);
                      })
-                    .collect(Collectors.toList());
+                     .collect(Collectors.toList());
     }
 
     private static List<ColumnType<?>> getPartitionKeyColumnTypes(TableInfoProvider tableInfo)
@@ -156,35 +169,44 @@ public class TableSchema implements Serializable
         }
     }
 
-    private static String getInsertStatement(StructType dfSchema, TableInfoProvider tableInfo,
-                                             TTLOption ttlOption, TimestampOption timestampOption)
+    private String getInsertStatement(StructType dfSchema,
+                                      TableInfoProvider tableInfo,
+                                      TTLOption ttlOption,
+                                      TimestampOption timestampOption)
     {
+        CassandraBridge bridge = CassandraBridgeFactory.get(lowestCassandraVersion);
         List<String> columnNames = Arrays.stream(dfSchema.fieldNames())
                                          .filter(fieldName -> !fieldName.equals(ttlOption.columnName()))
                                          .filter(fieldName -> !fieldName.equals(timestampOption.columnName()))
                                          .collect(Collectors.toList());
         StringBuilder stringBuilder = new StringBuilder("INSERT INTO ")
-                .append(tableInfo.getKeyspaceName())
-                .append(".").append(tableInfo.getName())
-                .append(columnNames.stream().collect(Collectors.joining(",", " (", ") ")))
-                .append("VALUES")
-                .append(columnNames.stream().map(columnName -> ":" + columnName).collect(Collectors.joining(",", " (", ")")));
+                                      .append(maybeQuotedIdentifier(bridge, quoteIdentifiers, tableInfo.getKeyspaceName()))
+                                      .append(".")
+                                      .append(maybeQuotedIdentifier(bridge, quoteIdentifiers, tableInfo.getName()))
+                                      .append(columnNames.stream()
+                                                         .map(columnName -> maybeQuotedIdentifier(bridge, quoteIdentifiers, columnName))
+                                                         .collect(Collectors.joining(",", " (", ") ")));
+
+        stringBuilder.append("VALUES")
+                     .append(columnNames.stream()
+                                        .map(columnName -> ":" + maybeQuotedIdentifier(bridge, quoteIdentifiers, columnName))
+                                        .collect(Collectors.joining(",", " (", ")")));
         if (ttlOption.withTTl() && timestampOption.withTimestamp())
         {
             stringBuilder.append(" USING TIMESTAMP ")
-                         .append(timestampOption)
+                         .append(timestampOption.toCQLString(columnName -> maybeQuotedIdentifier(bridge, quoteIdentifiers, columnName)))
                          .append(" AND TTL ")
-                         .append(ttlOption);
+                         .append(ttlOption.toCQLString(columnName -> maybeQuotedIdentifier(bridge, quoteIdentifiers, columnName)));
         }
         else if (timestampOption.withTimestamp())
         {
             stringBuilder.append(" USING TIMESTAMP ")
-                         .append(timestampOption);
+                         .append(timestampOption.toCQLString(columnName -> maybeQuotedIdentifier(bridge, quoteIdentifiers, columnName)));
         }
         else if (ttlOption.withTTl())
         {
             stringBuilder.append(" USING TTL ")
-                    .append(ttlOption);
+                         .append(ttlOption.toCQLString(columnName -> maybeQuotedIdentifier(bridge, quoteIdentifiers, columnName)));
         }
         stringBuilder.append(";");
         String insertStatement = stringBuilder.toString();
@@ -195,10 +217,11 @@ public class TableSchema implements Serializable
 
     private String getDeleteStatement(StructType dfSchema, TableInfoProvider tableInfo)
     {
-        Stream<String> fieldEqualityStatements = Arrays.stream(dfSchema.fieldNames()).map(key -> key + "=?");
+        CassandraBridge bridge = CassandraBridgeFactory.get(lowestCassandraVersion);
+        Stream<String> fieldEqualityStatements = Arrays.stream(dfSchema.fieldNames()).map(key -> maybeQuotedIdentifier(bridge, quoteIdentifiers, key) + "=?");
         String deleteStatement = String.format("DELETE FROM %s.%s where %s;",
-                                               tableInfo.getKeyspaceName(),
-                                               tableInfo.getName(),
+                                               maybeQuotedIdentifier(bridge, quoteIdentifiers, tableInfo.getKeyspaceName()),
+                                               maybeQuotedIdentifier(bridge, quoteIdentifiers, tableInfo.getName()),
                                                fieldEqualityStatements.collect(Collectors.joining(" AND ")));
 
         LOGGER.info("CQL delete statement for the RDD {}", deleteStatement);
--- a/cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/RecordWriterTest.java
+++ b/cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/RecordWriterTest.java
@@ -29,7 +29,9 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Range;
+import org.apache.commons.lang3.tuple.Pair;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
@@ -37,9 +39,19 @@ import org.junit.jupiter.api.io.TempDir;
 
 import org.apache.cassandra.bridge.RowBufferMode;
 import org.apache.cassandra.spark.bulkwriter.token.CassandraRing;
+import org.apache.cassandra.spark.bulkwriter.token.ConsistencyLevel;
 import org.apache.cassandra.spark.common.model.CassandraInstance;
+import org.apache.cassandra.spark.data.CqlField;
+import org.apache.spark.sql.types.DataType;
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.StructType;
 import scala.Tuple2;
 
+import static org.apache.cassandra.spark.bulkwriter.MockBulkWriterContext.DEFAULT_CASSANDRA_VERSION;
+import static org.apache.cassandra.spark.bulkwriter.SqlToCqlTypeConverter.DATE;
+import static org.apache.cassandra.spark.bulkwriter.SqlToCqlTypeConverter.INT;
+import static org.apache.cassandra.spark.bulkwriter.SqlToCqlTypeConverter.VARCHAR;
+import static org.apache.cassandra.spark.bulkwriter.TableSchemaTestCommon.mockCqlType;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.matchesPattern;
@@ -85,20 +97,41 @@ public class RecordWriterTest
     }
 
     @Test
+    public void testWriteWithMixedCaseColumnNames()
+    {
+        boolean quoteIdentifiers = true;
+        String[] pk = {"ID", "date"};
+        String[] columnNames = {"ID", "date", "course", "limit"};
+
+        Pair<StructType, ImmutableMap<String, CqlField.CqlType>> validPair = TableSchemaTestCommon.buildMatchedDataframeAndCqlColumns(
+        columnNames,
+        new DataType[]{DataTypes.IntegerType, DataTypes.DateType, DataTypes.StringType, DataTypes.IntegerType},
+        new CqlField.CqlType[]{mockCqlType(INT), mockCqlType(DATE), mockCqlType(VARCHAR), mockCqlType(INT)});
+
+        MockBulkWriterContext writerContext = new MockBulkWriterContext(ring,
+                                                                        DEFAULT_CASSANDRA_VERSION,
+                                                                        ConsistencyLevel.CL.LOCAL_QUORUM,
+                                                                        validPair,
+                                                                        pk,
+                                                                        pk,
+                                                                        quoteIdentifiers);
+        Iterator<Tuple2<DecoratedKey, Object[]>> data = generateData(5, true);
+        validateSuccessfulWrite(writerContext, data, columnNames);
+    }
+
+    @Test
     public void testWriteWithConstantTTL()
     {
-        MockBulkWriterContext bulkWriterContext = new MockBulkWriterContext(ring);
         Iterator<Tuple2<DecoratedKey, Object[]>> data = generateData(5, true, false, false);
-        validateSuccessfulWrite(bulkWriterContext, data, COLUMN_NAMES);
+        validateSuccessfulWrite(writerContext, data, COLUMN_NAMES);
     }
 
     @Test
     public void testWriteWithTTLColumn()
     {
-        MockBulkWriterContext bulkWriterContext = new MockBulkWriterContext(ring);
         Iterator<Tuple2<DecoratedKey, Object[]>> data = generateData(5, true, true, false);
         String[] columnNamesWithTtl = {"id", "date", "course", "marks", "ttl"};
-        validateSuccessfulWrite(bulkWriterContext, data, columnNamesWithTtl);
+        validateSuccessfulWrite(writerContext, data, columnNamesWithTtl);
     }
 
     @Test
--- a/cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/IntegrationTestBase.java
+++ b/cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/IntegrationTestBase.java
@@ -39,7 +39,6 @@ import org.junit.jupiter.api.BeforeEach;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.datastax.driver.core.KeyspaceMetadata;
 import com.datastax.driver.core.Session;
 import com.google.inject.Guice;
 import com.google.inject.Injector;
@@ -49,6 +48,7 @@ import io.vertx.core.eventbus.Message;
 import io.vertx.core.eventbus.MessageConsumer;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.web.client.WebClient;
+import io.vertx.ext.web.client.predicate.ResponsePredicate;
 import io.vertx.junit5.VertxTestContext;
 import org.apache.cassandra.sidecar.cluster.CassandraAdapterDelegate;
 import org.apache.cassandra.sidecar.cluster.instance.InstanceMetadata;
@@ -68,7 +68,7 @@ import static org.apache.cassandra.sidecar.server.SidecarServerEvents.ON_CASSAND
 public abstract class IntegrationTestBase
 {
     private static final int MAX_KEYSPACE_TABLE_WAIT_ATTEMPTS = 100;
-    private static final long MAX_KEYSPACE_TABLE_TIME = 100L;
+    private static final long MAX_KEYSPACE_TABLE_TIME = 200L;
     protected Logger logger = LoggerFactory.getLogger(this.getClass());
     protected Vertx vertx;
     protected Server server;
@@ -229,41 +229,37 @@ public abstract class IntegrationTestBase
     }
 
     /**
-     * Waits for the specified keyspace/table to be available.
+     * Waits for the specified keyspace to be available in Sidecar.
      * Empirically, this loop usually executes either zero or one time before completing.
      * However, we set a fairly high number of retries to account for variability in build machines.
      *
-     * @param keyspaceName the keyspace for which to wait
-     * @param tableName    the table in the keyspace for which to wait
+     * @param keyspace the keyspace for which to wait
      */
-    protected void waitForKeyspaceAndTable(String keyspaceName, String tableName)
+    protected void waitUntilSidecarPicksUpSchemaChange(String keyspace)
     {
-        int numInstances = sidecarTestContext.instancesConfig().instances().size();
         int retries = MAX_KEYSPACE_TABLE_WAIT_ATTEMPTS;
-        boolean sidecarMissingSchema = true;
-        while (sidecarMissingSchema && retries-- > 0)
+        WebClient client = WebClient.create(vertx);
+        while (retries-- > 0)
         {
-            sidecarMissingSchema = false;
-            for (int i = 0; i < numInstances; i++)
+            try
             {
-                KeyspaceMetadata keyspace = sidecarTestContext.session(i)
-                                                              .getCluster()
-                                                              .getMetadata()
-                                                              .getKeyspace(keyspaceName);
-                sidecarMissingSchema |= (keyspace == null || keyspace.getTable(tableName) == null);
-            }
-            if (sidecarMissingSchema)
-            {
-                logger.info("Keyspace/table {}/{} not yet available - waiting...", keyspaceName, tableName);
-                Uninterruptibles.sleepUninterruptibly(MAX_KEYSPACE_TABLE_TIME, TimeUnit.MILLISECONDS);
-            }
-            else
-            {
+                client.get(server.actualPort(), "localhost", "/api/v1/keyspaces/" + keyspace + "/schema")
+                      .expect(ResponsePredicate.SC_OK)
+                      .send()
+                      .toCompletionStage()
+                      .toCompletableFuture()
+                      .get(MAX_KEYSPACE_TABLE_TIME, TimeUnit.MILLISECONDS);
+                logger.info("Schema is ready in Sidecar");
+                client.close();
                 return;
             }
+            catch (Exception exception)
+            {
+                logger.info("Waiting for schema to propagate to Sidecar");
+                Uninterruptibles.sleepUninterruptibly(MAX_KEYSPACE_TABLE_TIME, TimeUnit.MILLISECONDS);
+            }
         }
-        throw new RuntimeException(
-        String.format("Keyspace/table %s/%s did not become visible on all sidecar instances",
-                      keyspaceName, tableName));
+        client.close();
+        throw new RuntimeException(String.format("Keyspace %s did not become visible in Sidecar", keyspace));
     }
 }
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,6 +58,7 @@ jobs:
             SPARK_VERSION: "2"
             SCALA_VERSION: "2.11"
             JDK_VERSION: "1.8"
+            INTEGRATION_MAX_PARALLEL_FORKS: 2
 
       - store_artifacts:
           path: build/test-reports
@@ -85,6 +86,7 @@ jobs:
             SPARK_VERSION: "2"
             SCALA_VERSION: "2.12"
             JDK_VERSION: "1.8"
+            INTEGRATION_MAX_PARALLEL_FORKS: 2
 
       - store_artifacts:
           path: build/test-reports
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 1.0.0
+ * Fix bulk writing when using identifiers that need quotes (CASSANDRA-19031)
  * Fix bulk reading when using identifiers that need quotes (CASSANDRA-19024)
  * Remove unused dead code (CASSANDRA-19148)
  * Get Sidecar port through CassandraContext for more flexibility (CASSANDRA-19903)
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/bridge/CassandraBridgeFactory.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/bridge/CassandraBridgeFactory.java
@@ -124,12 +124,24 @@ public final class CassandraBridgeFactory
             Class<CassandraBridge> bridge = (Class<CassandraBridge>) loader.loadClass("org.apache.cassandra.bridge.CassandraBridgeImplementation");
             Constructor<CassandraBridge> constructor = bridge.getConstructor();
             return constructor.newInstance();
-
         }
         catch (IOException | ClassNotFoundException | NoSuchMethodException | InstantiationException
              | IllegalAccessException | InvocationTargetException exception)
         {
             throw new RuntimeException("Failed to create Cassandra bridge for label " + label, exception);
         }
+    }
+
+    /***
+     * Returns the quoted name when the {@code quoteIdentifiers} parameter is {@code true} <i>AND</i> the
+     * {@code unquotedName} needs to be quoted (i.e. it uses mixed case, or it is a Cassandra reserved word).
+     * @param bridge the Cassandra bridge
+     * @param quoteIdentifiers whether identifiers should be quoted
+     * @param unquotedName the unquoted name to maybe be quoted
+     * @return the quoted name when the conditions are met
+     */
+    public static String maybeQuotedIdentifier(CassandraBridge bridge, boolean quoteIdentifiers, String unquotedName)
+    {
+        return quoteIdentifiers ? bridge.maybeQuoteIdentifier(unquotedName) : unquotedName;
     }
 }
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
@@ -126,6 +126,7 @@ public class BulkSparkConf implements Serializable
     protected final SparkConf conf;
     public final boolean validateSSTables;
     public final int commitThreadsPerInstance;
+    public boolean quoteIdentifiers;
     protected final int effectiveSidecarPort;
     protected final int userProvidedSidecarPort;
     protected boolean useOpenSsl;
@@ -166,6 +167,7 @@ public class BulkSparkConf implements Serializable
         this.ringRetryCount = getInt(RING_RETRY_COUNT, DEFAULT_RING_RETRY_COUNT);
         this.ttl = MapUtils.getOrDefault(options, WriterOptions.TTL.name(), null);
         this.timestamp = MapUtils.getOrDefault(options, WriterOptions.TIMESTAMP.name(), null);
+        this.quoteIdentifiers = MapUtils.getBoolean(options, WriterOptions.QUOTE_IDENTIFIERS.name(), false, "quote identifiers");
         validateEnvironment();
     }
 
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
@@ -63,6 +63,9 @@ public class CassandraBulkWriterContext implements BulkWriterContext, KryoSerial
     {
         this.conf = conf;
         this.clusterInfo = clusterInfo;
+        String lowestCassandraVersion = clusterInfo.getLowestCassandraVersion();
+        CassandraBridge bridge = CassandraBridgeFactory.get(lowestCassandraVersion);
+
         CassandraRing<RingInstance> ring = clusterInfo.getRing(true);
         jobInfo = new CassandraJobInfo(conf,
                                        new TokenPartitioner(ring, conf.numberSplits, sparkContext.defaultParallelism(), conf.getCores()));
@@ -71,20 +74,20 @@ public class CassandraBulkWriterContext implements BulkWriterContext, KryoSerial
                                     String.format("Keyspace %s is not replicated on datacenter %s",
                                                   conf.keyspace, conf.localDC));
 
-        String keyspace = conf.keyspace;
-        String table = conf.table;
+        String keyspace = jobInfo.keyspace();
+        String table = jobInfo.tableName();
 
         String keyspaceSchema = clusterInfo.getKeyspaceSchema(true);
-        CassandraBridge bridge = CassandraBridgeFactory.get(clusterInfo.getLowestCassandraVersion());
         Partitioner partitioner = clusterInfo.getPartitioner();
-        String tableSchema = CqlUtils.extractTableSchema(keyspaceSchema, keyspace, table);
+        String createTableSchema = CqlUtils.extractTableSchema(keyspaceSchema, keyspace, table);
         Set<String> udts = CqlUtils.extractUdts(keyspaceSchema, keyspace);
         ReplicationFactor replicationFactor = CqlUtils.extractReplicationFactor(keyspaceSchema, keyspace);
         int indexCount = CqlUtils.extractIndexCount(keyspaceSchema, keyspace, table);
-        CqlTable cqlTable = bridge.buildSchema(tableSchema, keyspace, replicationFactor, partitioner, udts, null, indexCount);
+        CqlTable cqlTable = bridge.buildSchema(createTableSchema, keyspace, replicationFactor, partitioner, udts, null, indexCount);
 
-        TableInfoProvider tableInfoProvider = new CqlTableInfoProvider(tableSchema, cqlTable);
-        schemaInfo = new CassandraSchemaInfo(new TableSchema(dfSchema, tableInfoProvider, conf.writeMode, conf.getTTLOptions(), conf.getTimestampOptions()));
+        TableInfoProvider tableInfoProvider = new CqlTableInfoProvider(createTableSchema, cqlTable);
+        TableSchema tableSchema = initializeTableSchema(conf, dfSchema, tableInfoProvider, lowestCassandraVersion);
+        schemaInfo = new CassandraSchemaInfo(tableSchema);
     }
 
     public static BulkWriterContext fromOptions(@NotNull SparkContext sparkContext,
@@ -180,8 +183,27 @@ public class CassandraBulkWriterContext implements BulkWriterContext, KryoSerial
     {
         if (dataTransferApi == null)
         {
-            dataTransferApi = new SidecarDataTransferApi(clusterInfo.getCassandraContext(), jobInfo, conf);
+            CassandraBridge bridge = CassandraBridgeFactory.get(clusterInfo.getLowestCassandraVersion());
+            dataTransferApi = new SidecarDataTransferApi(clusterInfo.getCassandraContext(),
+                                                         bridge,
+                                                         jobInfo,
+                                                         conf);
         }
         return dataTransferApi;
+    }
+
+    @NotNull
+    protected TableSchema initializeTableSchema(@NotNull BulkSparkConf conf,
+                                                @NotNull StructType dfSchema,
+                                                TableInfoProvider tableInfoProvider,
+                                                String lowestCassandraVersion)
+    {
+        return new TableSchema(dfSchema,
+                               tableInfoProvider,
+                               conf.writeMode,
+                               conf.getTTLOptions(),
+                               conf.getTimestampOptions(),
+                               lowestCassandraVersion,
+                               conf.quoteIdentifiers);
     }
 }
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraJobInfo.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraJobInfo.java
@@ -113,9 +113,20 @@ public class CassandraJobInfo implements JobInfo
     }
 
     @Override
-    @NotNull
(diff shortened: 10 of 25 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
```diff
@@ -20,18 +20,43 @@
...
+     * - ONE and TWO consistency guarantees are maintained by ensuring that the failures leave us with at-least the corresponding healthy
+     *   (and non-pending) nodes.
+     *
+     *   For both the above cases, blocked instances are also considered as failures while performing consistency checks.
+     *   Write replicas are adjusted to exclude replacement nodes for consistency checks, if we have replacement nodes that are not among the failed instances.
+     *   This is to ensure that we are writing to sufficient non-replacement nodes as replacements can potentially fail leaving us with fewer nodes.
+     * </pre>
+     */
+    boolean checkConsistency(CassandraRing ring,
+                             TokenRangeMapping<? extends CassandraInstance> tokenRangeMapping,
```
yifan-c: It is unclear to having `ring` and `tokenRangeMapping` as parameter for the consistency level check API. 
What do you think about passing the effective replica set. So the check api look like this

```
boolean checkConsistency(Collection<? extends CassandraInstance> writeReplicaSet, Collection<? extends CassandraInstance> failedInsts, String localDC);
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
```diff
@@ -82,19 +119,25 @@ public boolean checkConsistency(Collection<? extends CassandraInstance> failedIn
...
+                                            final Collection<? extends CassandraInstance> failedInsts,
+                                            final String localDC)
             {
-                int rf = replicationFactor.getTotalReplicationFactor();
-                return failedInsts.size() <= (rf - (rf / 2 + 1));
+                Set<String> replacingInstances = tokenRangeMapping.getReplacementInstances();
+                Set<String> failedInstanceIPs = failedInsts.stream().map(CassandraInstance::getIpAddress).collect(Collectors.toSet());
+                final long writeReplicaCount = maybeUpdateWriteReplicasForReplacements(tokenRangeMapping.getWriteReplicas(),
+                                                                                       replacingInstances,
+                                                                                       failedInstanceIPs);
```
yifan-c: Similar to the API change comment. It is not the consistency level variants' responsibility to calculate the effective write replica set.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
arjunashok: Fair point. The reason I pulled this in is because some of these values are CL specific. For instance, we use local_dc for LOCAL_QUORUM and LOCAL_ONE. Also, for multi-DC, we cycle through all DCs to find consistency failures.

What you're suggesting will require some specific checks up-front (before `checkConsistency`) for the cases I mentioned, but it will isolate CL functionality to some extent. Will update

arjunashok: Addressed

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
arjunashok: Addressed

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -136,35 +185,122 @@ public StreamResult write(Iterator<Tuple2<DecoratedKey, Object[]>> sourceIterato
...
+    private boolean haveTokenRangeMappingsChanged(TokenRangeMapping<RingInstance> startTaskMapping, TaskContext taskContext)
+    {
+        Range<BigInteger> taskTokenRange = getTokenRange(taskContext);
+        // Get the uncached, current view of the ring to compare with initial ring
+        TokenRangeMapping<RingInstance> endTaskMapping = writerContext.cluster().getTokenRangeMapping(false);
+        Map<Range<BigInteger>, List<RingInstance>> startMapping = taskTokenRangeMapping(startTaskMapping, taskTokenRange);
+        Map<Range<BigInteger>, List<RingInstance>> endMapping = taskTokenRangeMapping(endTaskMapping, taskTokenRange);
+
+        return !(startMapping.keySet().equals(endMapping.keySet()) &&
+               instancesFromMapping(startMapping).equals(instancesFromMapping(endMapping)));
```
yifan-c: nit: I think this is easier to read with less parenthesis. 
```suggestion
        return !Objects.equals(startMapping.keySet(), endMapping.keySet()) ||
               !Objects.equals(instancesFromMapping(startMapping), instancesFromMapping(endMapping));
```

A follow up question. In the scenario that a node is joining at the beginning of a task and finishes joining at the end, the `haveTokenRangeMappingsChanged` would return false. Am I right?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -110,20 +132,47 @@ public StreamResult write(Iterator<Tuple2<DecoratedKey, Object[]>> sourceIterato
...
+                sessions.add(streamSession);
                 maybeCreateTableWriter(partitionId, baseDir);
-                writeRow(valueMap, dataIterator, partitionId, range);
+                writeRow(rowData, valueMap, partitionId, streamSession.getTokenRange());
                 checkBatchSize(streamSession, partitionId, job);
             }
 
-            if (sstableWriter != null)
+            // Finalize SSTable for the last StreamSession
+            if (sstableWriter != null || (streamSession != null && batchSize != 0))
```
yifan-c: `streamSession` cannot be null. The check is redundant. When `sstableWriter` is not null, the `batchSize` cannot be 0. The condition `batchSize != 0` is redundant too.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -110,20 +132,47 @@ public StreamResult write(Iterator<Tuple2<DecoratedKey, Object[]>> sourceIterato
         Map<String, Object> valueMap = new HashMap<>();
         try
         {
+            List<RingInstance> exclusions = failureHandler.getFailedInstances();
+            Set<Range<BigInteger>> newRanges = initialTokenRangeMapping.getRangeMap().asMapOfRanges().entrySet()
+                                                                       .stream()
+                                                                       .filter(e -> !exclusions.contains(e.getValue()))
```
yifan-c: `exclusions` is a list, and `e.getValue()` returns a list too. `list.contains(list)` is always evaluated to be `false`.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -136,35 +185,122 @@ public StreamResult write(Iterator<Tuple2<DecoratedKey, Object[]>> sourceIterato
...
+                Preconditions.checkState(matchingRange != null,
+                                         String.format("Received Token %s outside of expected range %s", token, matchingRange));
+
+                if (streamSession != null && streamSession.getTokenRange() == matchingRange)
+                {
+                    return streamSession;
+                }
+                else
+                {
+                    LOGGER.debug(String.format("[{}] Creating stream session for range: %s", matchingRange), taskContext.partitionId());
```
yifan-c: `String.format` is redundant

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -136,35 +185,122 @@ public StreamResult write(Iterator<Tuple2<DecoratedKey, Object[]>> sourceIterato
...
+                                                   ReplicaAwareFailureHandler<RingInstance> failureHandler) throws IOException
+    {
+        BigInteger token = rowData._1().getToken();
+        Range<BigInteger> tokenRange = getTokenRange(taskContext);
+
+        Preconditions.checkState(tokenRange.contains(token),
+                                 String.format("Received Token %s outside of expected range %s", token, tokenRange));
+
+        // token range for this partition is not among the write-replica-set ranges
+        if (!newRanges.contains(tokenRange))
```
yifan-c: This check is performed for _every_ row. Is it really needed here? The `newRanges` and `tokenRange` remain the same while writing (the `while` loop)

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -110,20 +132,47 @@ public StreamResult write(Iterator<Tuple2<DecoratedKey, Object[]>> sourceIterato
...
+                                                                       .filter(e -> !exclusions.contains(e.getValue()))
+                                                                       .map(Map.Entry::getKey)
+                                                                       .collect(Collectors.toSet());
+
             while (dataIterator.hasNext())
             {
+                Tuple2<DecoratedKey, Object[]> rowData = dataIterator.next();
+                streamSession = maybeCreateStreamSession(taskContext, streamSession, rowData, newRanges, failureHandler);
+
+                sessions.add(streamSession);
```
yifan-c: There is one active session at a time. Why collecting the sessions rather than close the previous session immediately?

yifan-c (commented):
Not a complete review. Submitting the comments so far.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
arjunashok: Thanks for the input.

We're checking if either the token-ranges or the set of instances differ between the start and end of the task to determine if it should fail the task (and retry).

In your example, if the the node is joining at the beginning of the task and has completed joining before the task ends, the above condition is still good as nothing would have changed and it would return `false`.

It is in the alternate case i.e. when the transitioning nodes appear during task execution that we want to detect and fail-fast.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
arjunashok: This was done to separate the session closures instead of having it scattered with checks for non-existent sessions. 

However, on looking further it seems like we do not really need to eagerly create the session for the partition's token range. Will update for it to be created lazily.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
arjunashok: Good catch. Evaluating if we even need this filter anymore. Will update.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
arjunashok: Good point, removing redundant checks.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
arjunashok: Makes sense. Pulled this out of the loop.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -132,37 +133,32 @@ public List<StreamResult> write(Iterator<Tuple2<DecoratedKey, Object[]>> sourceI
...
-                sessions.add(streamSession);
+                streamSession = maybeCreateStreamSession(taskContext, streamSession, rowData, subRanges, failureHandler, results);
                 maybeCreateTableWriter(partitionId, baseDir);
                 writeRow(rowData, valueMap, partitionId, streamSession.getTokenRange());
                 checkBatchSize(streamSession, partitionId, job);
             }
 
-            // Finalize SSTable for the last StreamSession
-            if (sstableWriter != null || (streamSession != null && batchSize != 0))
+            // Cleanup SSTable writer and schedule the last stream
```
yifan-c: "Cleanup SSTable writer" reads wrong to me. I would stick with "Finalize". The code is to flush any data to sstable by closing the writer. Cleanup leads me to think it remove files.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -207,46 +203,62 @@ private Set<RingInstance> instancesFromMapping(Map<Range<BigInteger>, List<RingI
...
-            if (subRanges.size() > 1)
-            {
-                // Create session using sub-range that contains the token from current row
-                Range<BigInteger> matchingRange = subRanges.stream().filter(r -> r.contains(token)).findFirst().get();
-                Preconditions.checkState(matchingRange != null,
-                                         String.format("Received Token %s outside of expected range %s", token, matchingRange));
+            // Create session using sub-range that contains the token from current row
+            Range<BigInteger> matchingSubRange = subRanges.stream().filter(r -> r.contains(token)).findFirst().get();
+            Preconditions.checkState(matchingSubRange != null,
+                                     String.format("Received Token %s outside of expected range %s", token, matchingSubRange));
```
yifan-c: The `checkState` will not ever see `matchingSubRange == null`. The reason is that at line#222, if the value is null, the `get()` operation throws exception already. 
If the intent to provide a more user friendly error message, can you not call `get()` and use `Optional<Range<BigInteger>> matchingSubRangeOpt` to capture the result and run `checkState` on the optional.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/RecordWriterTest.java
```diff
@@ -346,19 +366,22 @@ void writeBuffered()
...
     private void validateSuccessfulWrite(MockBulkWriterContext writerContext,
                                          Iterator<Tuple2<DecoratedKey, Object[]>> data,
                                          String[] columnNames,
-                                         int uploadedTables)
+                                         int uploadedTables) throws InterruptedException
     {
         RecordWriter rw = new RecordWriter(writerContext, columnNames, () -> tc, SSTableWriter::new);
         rw.write(data);
+        // Wait for uploads to finish
+        Thread.sleep(500);
```
yifan-c: using `sleep` is prone to flaky test. It depends a lot on the test runtime. We should do the best to avoid the usage.
The `rw.write` is already blocking. It blocks until the sstables are uploaded. I am not sure why sleep is needed.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/RecordWriterTest.java
```diff
@@ -346,19 +366,22 @@ void writeBuffered()
...
         Map<CassandraInstance, List<UploadRequest>> uploads = writerContext.getUploads();
+        System.out.println("Uploads in test: " + uploads.values().stream().mapToInt(List::size).sum());
```
yifan-c: remove it or use logger.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/TokenRangeMappingUtils.java
```diff
@@ -82,6 +106,7 @@ public static TokenRangeMapping<RingInstance> buildTokenRangeMapping(final int i
                                        Collections.emptySet());
     }
 
+
```
yifan-c: nit: remove duplicated empty lines

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/TokenRangeMappingUtils.java
```diff
@@ -50,13 +51,36 @@ private TokenRangeMappingUtils()
...
+        RingInstance newInstance = new RingInstance(newEntry);
+        instances.add(0, newInstance);
         ReplicationFactor replicationFactor = getReplicationFactor(rfByDC);
         Map<String, Set<String>> writeReplicas =
-        instances.stream()
-                 .collect(Collectors.groupingBy(RingInstance::getDataCenter,
-                                                Collectors.mapping(RingInstance::getNodeName, Collectors.toSet())));
+        instances.stream().collect(Collectors.groupingBy(RingInstance::getDataCenter,
+                                                         Collectors.mapping(RingInstance::getNodeName,
+                                                                            Collectors.toSet())));
```
yifan-c: nit: prefer to not reformatting if there is no actual code change.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriteValidator.java
```diff
@@ -23,103 +23,120 @@
...
+    }
+
     private void updateInstanceAvailability()
     {
         cluster.refreshClusterInfo();
         Map<RingInstance, InstanceAvailability> availability = cluster.getInstanceAvailability();
         availability.forEach(this::checkInstance);
     }
 
+    private void updateFailureHandler(RingInstance instance, InstanceAvailability availability)
```
yifan-c: This method is only called in `private void checkInstance(RingInstance instance, InstanceAvailability availability)`, which throws exception when `availability == InstanceAvailability.INVALID_STATE`. It voids the need to check the condition again in this method.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -98,39 +95,19 @@ public void insert(@NotNull Dataset<Row> data, boolean overwrite)
...
-                .map(Row::toSeq)
-                .map(seq -> JavaConverters.seqAsJavaListConverter(seq).asJava().toArray())
-                .map(tableSchema::normalize)
-                .keyBy(tokenizer::getDecoratedKey)
-                .repartitionAndSortWithinPartitions(broadcastContext.getValue().job().getTokenPartitioner());
+                                                            .map(Row::toSeq)
+                                                            .map(seq -> JavaConverters.seqAsJavaListConverter(seq).asJava().toArray())
+                                                            .map(tableSchema::normalize)
+                                                            .keyBy(tokenizer::getDecoratedKey)
+                                                            .repartitionAndSortWithinPartitions(broadcastContext.getValue().job().getTokenPartitioner());
```
yifan-c: The indentation was so for the sake of readability. 

The reformatting has no other change but just increasing indentations. Such reformatting is not preferred as it creates noise for review.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -339,140 +336,190 @@ public String getLowestCassandraVersion()
...
+        // Removes blocked instances. If this is included, remove blockedInstances from CL checks
+        merged.addAll(instancesList1.stream().filter(i -> !blockedIPs.contains(i)).collect(Collectors.toSet()));
+        merged.addAll(instancesList2.stream().filter(i -> !blockedIPs.contains(i)).collect(Collectors.toSet()));
+
+        return merged;
     }
 
-    private static RingInstance getCasInstanceMethodsImpl(RingEntry ringEntry)
+    // Pending replicas are currently calculated by extracting the non-read-replicas from the write-replica-set
+    // This will be replaced by the instance state metadata when it is supported by the token-ranges API
```
yifan-c: Replica metadata is exposed from the `TokenRangeReplicasResponse` now. Is the implementation outdated?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -339,140 +336,190 @@ public String getLowestCassandraVersion()
...
-                catch (TimeoutException exception)
-                {
-                    Thread.currentThread().interrupt();
-                    throw new RuntimeException("Failed to retrieve gossip information", exception);
-                }
+                // For each writeReplica, get metadata and update map to include range
+                dcReplicaEntry.getValue().forEach(ipAddress -> {
+                    // Get metadata for this IP; Create RingInstance
+                    // TODO: Temporary change to extract IP from 'ip:port' string. THis will go oway once
+                    // corresponding change in sidecar is merged.
```
yifan-c: Is the todo resolved now?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -339,140 +336,190 @@ public String getLowestCassandraVersion()
...
+                    ReplicaMetadata replica = replicaMetadata.stream()
+                                                             .filter(r ->
+                                                                     r.address().equals(ipAddress.split(":")[0]))
```
yifan-c: what if it is an IPv6 address?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -339,140 +336,190 @@ public String getLowestCassandraVersion()
...
+                                                             .findFirst().get();
```
yifan-c: can you handle the potential NoSuchElementException from `get()` gracefully?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -339,140 +336,190 @@ public String getLowestCassandraVersion()
...
-    private static RingInstance getCasInstanceMethodsImpl(RingEntry ringEntry)
+    // Pending replicas are currently calculated by extracting the non-read-replicas from the write-replica-set
+    // This will be replaced by the instance state metadata when it is supported by the token-ranges API
+    private Map<String, Set<String>> getPendingReplicas(TokenRangeReplicasResponse response, Map<String, Set<String>> writeReplicasByDC)
     {
-        return new RingInstance(ringEntry);
+        Set<String> readReplicas = readReplicasFromTokenRangeResponse(response);
+        return writeReplicasByDC.entrySet()
+                                .stream()
+                                .filter(entry -> entry.getValue().stream().noneMatch(readReplicas::contains))
```
yifan-c: If my understanding is correct, the method is to find the pending nodes. If so, the filter to only include that list of instances (instances in a DC) that none is contained in `readReplicas` always returns `false`. The predicate feels wrong to me.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CommitCoordinator.java
```diff
@@ -128,61 +128,51 @@ private Stream<ListenableFuture<CommitResult>> commit(Map<RingInstance, Listenin
                                                           RingInstance instance,
                                                           Map<String, Range<BigInteger>> uploadRanges)
     {
-        if (cluster.instanceIsAvailable(instance))
```
yifan-c: What if the instance is down? The new code will try to commit the sstable on those down instances. Should it skip contacting sidecar and mark it as failure directly?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -110,20 +133,42 @@ public StreamResult write(Iterator<Tuple2<DecoratedKey, Object[]>> sourceIterato
         Map<String, Object> valueMap = new HashMap<>();
         try
         {
+            Set<Range<BigInteger>> newRanges = initialTokenRangeMapping.getRangeMap().asMapOfRanges().entrySet()
+                                                                       .stream()
+                                                                       .map(Map.Entry::getKey)
+                                                                       .collect(Collectors.toSet());
```
yifan-c: nit: this is simpler to get a copy of the key set

```suggestion
            Set<Range<BigInteger>> newRanges = new HashSet<>(initialTokenRangeMapping.getRangeMap().asMapOfRanges().keySet());
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -136,35 +181,139 @@ public StreamResult write(Iterator<Tuple2<DecoratedKey, Object[]>> sourceIterato
...
+            Instant localNow = Instant.now();
+            Instant remoteNow = Instant.ofEpochMilli(timeSkewResponse.currentTime);
+            Duration range = Duration.ofMinutes(timeSkewResponse.allowableSkewInMinutes);
+            if (localNow.isBefore(remoteNow.minus(range)) || localNow.isAfter(remoteNow.plus(range)))
+            {
+                final String message = String.format("Time skew between Spark and Cassandra is too large. "
+                                                     + "Allowable skew is %d minutes. Spark executor time is %s, Cassandra instance time is %s",
+                                                     timeSkewResponse.allowableSkewInMinutes, localNow, remoteNow);
+                throw new UnsupportedOperationException(message);
+            }
```
yifan-c: looks like it only adds the check for whether `replicas` is empty. The following has less indentation. 

```java
    private void validateAcceptableTimeSkewOrThrow(List<RingInstance> replicas)
    {
        if (replicas.isEmpty())
        {
            return;
        }
        // original logic
    }
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -175,30 +324,31 @@ public void writeRow(Map<String, Object> valueMap,
         }
     }
 
-    void checkBatchSize(StreamSession streamSession, int partitionId, JobInfo job) throws IOException
+    /**
+     * Stream to replicas; if batchSize is reached, "finalize" SST to "schedule" streamSession
+     */
+    private void checkBatchSize(final StreamSession streamSession, final int partitionId, final JobInfo job) throws IOException
```
yifan-c: drop unnecessary `final` in the method parameters.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/StreamSession.java
```diff
@@ -145,32 +157,27 @@ private List<CommitResult> commit(StreamResult streamResult) throws ExecutionExc
     @VisibleForTesting
     List<RingInstance> getReplicas()
     {
-        Map<Range<BigInteger>, List<RingInstance>> overlappingRanges = ring.getSubRanges(tokenRange).asMapOfRanges();
+        List<RingInstance> exclusions = failureHandler.getFailedInstances();
+        final Map<Range<BigInteger>, List<RingInstance>> overlappingRanges = tokenRangeMapping.getSubRanges(tokenRange).asMapOfRanges();
```
yifan-c: drop the `final` for local scoped variable.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/StreamSession.java
```diff
@@ -145,32 +157,27 @@ private List<CommitResult> commit(StreamResult streamResult) throws ExecutionExc
...
 
-        Preconditions.checkState(overlappingRanges.keySet().size() == 1,
-                                 String.format("Partition range %s is mapping more than one range %s",
-                                               tokenRange, overlappingRanges));
+        LOGGER.debug("[{}]: Stream session token range: {} overlaps with ring ranges: {}", sessionID, tokenRange, overlappingRanges);
+        Preconditions.checkState(!tokenRange.isEmpty(),
```
yifan-c: maybe move this check before getting the `overlappingRanges`, so that `tokenRange` is validated before using. 
Or.. do you mean to check `overlappingRanges` that is not empty?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ReplicaAwareFailureHandler.java
```diff
@@ -71,64 +72,138 @@ public void addFailure(Range<BigInteger> tokenRange, Instance casInstance, Strin
...
             Multimap<Instance, String> newErrorMap = ArrayListMultimap.create(entry.getValue());
-
             newErrorMap.put(casInstance, errMessage);
             mappingsToAdd.put(entry.getKey(), newErrorMap);
         }
         failedRangesMap.putAll(mappingsToAdd);
     }
 
-    public boolean hasFailed(ConsistencyLevel consistencyLevel, String localDC)
+    public List<Instance> getFailedInstances()
```
yifan-c: How about returning a set instead of list?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
```diff
@@ -94,6 +95,12 @@ public Object[] normalize(Object[] row)
     }
 
     public Object[] getKeyColumns(Object[] allColumns)
+    {
+        return getKeyColumns(allColumns, keyFieldPositions);
+    }
+
+    @NotNull
+    public static Object[] getKeyColumns(Object[] allColumns, List<Integer> keyFieldPositions)
```
yifan-c: What is the benefit of allow specifying the key field positions? Looks like the method is only used for testing via the `Tokenizer` constructor. If so, can you add the `@VisibleForTesting` annotation?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TokenPartitioner.java
```diff
@@ -138,15 +143,15 @@ private void validate()
...
+                                                           .sorted(Comparator.comparing(Range::lowerEndpoint))
+                                                           .collect(Collectors.toList());
         Range<BigInteger> previous = null;
         for (Range<BigInteger> current : sortedRanges)
         {
             if (previous != null)
             {
                 Preconditions.checkState(!current.isConnected(previous) || current.intersection(previous).isEmpty(),
-                        String.format("Two ranges in partition map are overlapping %s %s", previous, current));
+                                         String.format("Two ranges in partition map are overlapping %s %s", previous, current));
```
yifan-c: since this line is changed.. You can use the other checkState api to lazily evaluate the error message only when the check fails. 

```suggestion
                                         "Two ranges in partition map are overlapping %s %s", previous, current);
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TokenPartitioner.java
```diff
@@ -215,19 +221,19 @@ private void readObject(ObjectInputStream in) throws ClassNotFoundException, IOE
...
-    public int calculateSplits(CassandraRing<RingInstance> ring,
+    // Note that the actual number of partitions that result from this should always be at least the number of token ranges * the number of splits,
+    // but can be slightly more.
+    public int calculateSplits(TokenRangeMapping<RingInstance> tokenRangeMapping,
                                Integer numberSplits,
                                int defaultParallelism,
                                Integer cores)
     {
-        if (numberSplits >= 0)
+        if (numberSplits != -1)
```
yifan-c: This change is worrisome. There is no validation on the value of `numberSplits`. End user can provide a wrong value, say `-10`, the original code can handle it, but the changed code break.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TokenPartitioner.java
```diff
@@ -215,19 +221,19 @@ private void readObject(ObjectInputStream in) throws ClassNotFoundException, IOE
...
         {
             return numberSplits;
         }
-        int tasksToRun = Math.max(cores, defaultParallelism);
-        Map<Range<BigInteger>, List<RingInstance>> rangeListMap = ring.getRangeMap().asMapOfRanges();
+        final int tasksToRun = Math.max(cores, defaultParallelism);
+        final Map<Range<BigInteger>, List<RingInstance>> rangeListMap = tokenRangeMapping.getRangeMap().asMapOfRanges();
```
yifan-c: drop the `final`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
```diff
@@ -19,19 +19,42 @@
...
+     *   This is done to ensure that writes go to a quorum of healthy nodes while accounting for potential failure in pending nodes becoming healthy.
+     * - ONE and TWO consistency guarantees are maintained by ensuring that the failures leave us with at-least the corresponding healthy
+     *   (and non-pending) nodes.
+     *
+     *   For both the above cases, blocked instances are also considered as failures while performing consistency checks.
+     *   Write replicas are adjusted to exclude replacement nodes for consistency checks, if we have replacement nodes that are not among the failed instances.
+     *   This is to ensure that we are writing to sufficient non-replacement nodes as replacements can potentially fail leaving us with fewer nodes.
+     * </pre>
+     *
+     * TODO javadocs
```
yifan-c: Is it addressed?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
```diff
@@ -59,42 +87,34 @@ public boolean isLocal()
...
-                    if (failedInsts.stream()
-                                   .filter(instance -> instance.getDataCenter().matches(datacenter))
-                                   .count() > (rf - (rf / 2 + 1)))
-                    {
-                        return false;
-                    }
-                }
-
-                return true;
+                return (failedInstanceIps.size() + blockedInstances.size()) <= (writeReplicas.size() - (writeReplicas.size() / 2 + 1));
```
yifan-c: The implementation looks wrong to me. For EACH_QUORUM, the success condition is to meet quorum in each DC. The replication factor could vary per DC.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
```diff
@@ -59,42 +87,34 @@ public boolean isLocal()
...
             }
         },
         QUORUM
         {
+            // Keyspaces exist with RF 1 or 2
```
yifan-c: This comment does not make sense to be included in QUORUM.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ReplicaAwareFailureHandler.java
```diff
@@ -71,64 +72,138 @@ public void addFailure(Range<BigInteger> tokenRange, Instance casInstance, Strin
...
-    {
-        Multimap<Instance, String> failedInstances = ArrayListMultimap.create();
-        getFailedEntries(consistencyLevel, localDC).stream()
-                .flatMap(failedMultiEntry -> failedMultiEntry.getValue().entries().stream())
-                .forEach(failedEntry -> failedInstances.put(failedEntry.getKey(), failedEntry.getValue()));
+        List<AbstractMap.SimpleEntry<Range<BigInteger>, Multimap<Instance, String>>> failedEntries =
+        new ArrayList<>();
+
+        for (Map.Entry<Range<BigInteger>, Multimap<Instance, String>> failedRangeEntry
+        : failedRangesMap.asMapOfRanges().entrySet())
```
yifan-c: it is unusual to break it into 2 lines..

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
yifan-c: Looks like you moved the per DC check in `ReplicaAwareFailureHandler#validateConsistency`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/common/model/CassandraInstance.java
```diff
@@ -26,4 +26,6 @@ public interface CassandraInstance
     String getNodeName();
 
     String getDataCenter();
+
+    String getIpAddress();
```
yifan-c: add javadoc for the new method. 
It is fine to ignore the existing methods that are w/o javadoc.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/RecordWriterTest.java
```diff
@@ -69,64 +78,217 @@ public class RecordWriterTest
...
+        assertThat(uploads.values().stream().mapToInt(List::size).sum(), is(REPLICA_COUNT * FILES_PER_SSTABLE * UPLOADED_TABLES));
+        List<UploadRequest> requests = uploads.values().stream().flatMap(List::stream).collect(Collectors.toList());
+        for (UploadRequest ur : requests)
+        {
+            assertNotNull(ur.fileHash);
+        }
+    }
+
+    @Test
+    public void testWriteWithSubRanges2()
```
yifan-c: can the test name be more descriptive? If comment helps, please add comments too.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/TokenPartitionerTest.java
```diff
@@ -58,27 +58,34 @@ public void testOneSplit()
...
-        assertEquals(5, getPartitionForToken(100150));
-        assertEquals(5, getPartitionForToken(150000));
-        assertEquals(6, getPartitionForToken(150001));
-        assertEquals(6, getPartitionForToken(200000));
-        assertEquals(7, getPartitionForToken(200001));
-        assertEquals(7, getPartitionForToken(new BigInteger("4611686018427388003")));
-        assertEquals(7, getPartitionForToken(new BigInteger("4611686018427487903")));
-        assertEquals(8, getPartitionForToken(new BigInteger("4611686018427487904")));
-        assertEquals(9, getPartitionForToken(new BigInteger("9223372036854775807")));  // Single token range
+        final TokenRangeMapping<RingInstance> tokenRangeMapping = TokenRangeMappingUtils.buildTokenRangeMapping(0, ImmutableMap.of("DC1", 3), 3);
```
yifan-c: drop the final

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/CassandraSidecarTestContext.java
```diff
@@ -209,31 +222,49 @@ private InstancesConfig buildInstancesConfig(CassandraVersionProvider versionPro
...
+            IInstanceConfig config = instance.config();
+            String hostName = JMXUtil.getJmxHost(config);
+            int nativeTransportPort = tryGetIntConfig(config, "native_transport_port", 9042);
+            InetSocketAddress address = InetSocketAddress.createUnresolved(hostName,
+                                                                           nativeTransportPort);
+            TemporaryCqlSessionProvider sessionProvider = new TemporaryCqlSessionProvider(address, new NettyOptions(), addresses);
             this.sessionProviders.add(sessionProvider);
-            JmxClient jmxClient = new JmxClient(hostName, config.jmxPort());
+            // The in-jvm dtest framework sometimes returns a cluster before all the jmx infrastructure is initialized.
+            // In these cases, we want to wait longer than the default retry/delay settings to connect.
```
yifan-c: 👍  thanks for the comment

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/IntegrationTestModule.java
```diff
@@ -90,16 +88,25 @@ public InstanceMetadata instanceFromId(int id) throws NoSuchElementException
...
         }
     }
 
     @Provides
     @Singleton
     public SidecarConfiguration sidecarConfiguration()
     {
-        return new SidecarConfigurationImpl(new ServiceConfigurationImpl("127.0.0.1"));
+        ServiceConfiguration conf = ServiceConfigurationImpl.builder()
+                                                            .host("0.0.0.0") // binds to all interfaces, potential security issue if left running for long
```
yifan-c: Is it required to bind to `0.0.0.0`?

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/TemporaryCqlSessionProvider.java
```diff
@@ -0,0 +1,147 @@
...
+    {
+        super(target, options);
+        inet = target;
+        nettyOptions = options;
+        reconnectionPolicy = new ExponentialReconnectionPolicy(100, 1000);
+        this.addresses.addAll(addresses);
+    }
+
+    @Override
+    public synchronized @Nullable Session localCql()
```
yifan-c: nit
```suggestion
    @Override @Nullable
    public synchronized Session localCql()
```

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/ResiliencyTestBase.java
```diff
@@ -0,0 +1,384 @@
...
+                                                           : ss.getRangeToEndpointWithPortMap(TEST_KEYSPACE);
+
+        // filter ranges that belong to the instance
+        return ranges.entrySet()
+                            .stream()
+                            .filter(e -> e.getValue().contains(instance.broadcastAddress().getAddress().getHostAddress()
+                                                               + ":" + instance.broadcastAddress().getPort()))
+                            .map(e -> unwrapRanges(e.getKey()))
+                            .flatMap(Collection::stream)
+                            .collect(Collectors.toList());
```
yifan-c: not aligned

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/testing/CassandraIntegrationTest.java
```diff
@@ -59,6 +59,13 @@
      */
     int numDcs() default 1;
 
+    /**
+     * This is only applied in context of multi-DC tests. Returns true if the keyspace is replicated
+     * across multiple DCs. Defaults to {@code true}
+     * @return whether the multi-DC test uses a cross-DC keyspace
+     */
+    boolean useCrossDcKeyspace() default true;
```
yifan-c: It is more flexible to define keyspaces in the test directly with some helper method, instead of declaring an option in the annotation. For example, you can define keyspace with varying number of DCs and RFs per test. 
That being said. I am fine with the current code.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/expansion/JoiningBaseTest.java
```diff
@@ -0,0 +1,163 @@
...
+import org.apache.cassandra.distributed.api.IUpgradeableInstance;
+import org.apache.cassandra.distributed.api.TokenSupplier;
+import org.apache.cassandra.distributed.shared.ClusterUtils;
+import org.apache.cassandra.testing.CassandraIntegrationTest;
+import org.apache.cassandra.testing.ConfigurableCassandraTestContext;
+
+import static junit.framework.TestCase.assertNotNull;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class JoiningBaseTest extends ResiliencyTestBase
```
yifan-c: should it be abstract class?

yifan-c (commented):
Submitting my comments so far. I am now at the joining tests.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
arjunashok: Makes sense.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/RecordWriterTest.java
arjunashok: In general, I agree with the flakiness introduced by sleep. This was added because when the entire test suite was executed, we did see the uploads not finishing before we look up the no. files that we uploaded. We could potentially use a latch in the `MockBulkWriterContext` to make this more deterministic. Will explore some more.

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/IntegrationTestModule.java
arjunashok: Will defer this one to @JeetKunDoug.

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/testing/CassandraIntegrationTest.java
arjunashok: Agreed. We can revisit this when we are refactoring these helpers to make them less error prone (minimize the no. params being passed).

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/expansion/JoiningBaseTest.java
arjunashok: In these scenario specific base classes, we're only grouping common functionality instead of defining a contract for subclasses to implement, so this seemed appropriate.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/RecordWriterTest.java
yifan-c: Thanks for making the improvement!

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -339,140 +336,188 @@ public String getLowestCassandraVersion()
...
-                catch (TimeoutException exception)
-                {
-                    Thread.currentThread().interrupt();
-                    throw new RuntimeException("Failed to retrieve gossip information", exception);
-                }
+                // For each writeReplica, get metadata and update map to include range
+                dcReplicaEntry.getValue().forEach(ipAddress -> {
+                    // Get metadata for this IP; Create RingInstance
+                    ReplicaMetadata replica = replicaMetadata.get(ipAddress);
+                    instanceToRangeMap.put(new RingInstance(replica), range);
```
yifan-c: What if `replicaMetadata.get(ipAddress)` returns null? How about add a check for the invalid state?

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/IntegrationTestBase.java
```diff
@@ -231,4 +273,84 @@ protected void waitForKeyspaceAndTable(String keyspaceName, String tableName)
...
     }
+
+    /**
+     * A {@link DnsResolver} instance used for tests that provides fast DNS resolution, to avoid blocking
+     * DNS resolution at the JDK/OS-level.
+     *
+     * <p><b>NOTE:</b> The resolver assumes that the addresses are of the form 127.0.0.x, which is what is currently
+     * configured for integration tests.
+     */
+    static class FastDnsResolver implements DnsResolver
```
yifan-c: nit: `FastLocalhostResolver` to be more specific.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/ResiliencyTestBase.java
```diff
@@ -0,0 +1,501 @@
...
+            }
+        }
+    }
+
+    @BeforeAll
+    static void sparkSetup()
+    {
+
+    }
+
```
yifan-c: This method is meaningless. Can you remove it?

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/ResiliencyTestBase.java
```diff
@@ -0,0 +1,501 @@
...
+            rf = ImmutableMap.of("datacenter1", DEFAULT_RF, "datacenter2", DEFAULT_RF);
+        }
+        else
+        {
+            rf = ImmutableMap.of("datacenter1", DEFAULT_RF);
+        }
+
+        QualifiedTableName schema = initializeSchema(rf);
+        Thread.sleep(2000);
+        List<String> command = new ArrayList<>();
```
yifan-c: Can you add a comment to explain why running the spark job in a sub-process? I assume it is for resource isolation.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/ResiliencyTestBase.java
```diff
@@ -0,0 +1,501 @@
...
+            LOGGER.info("Running: {}", String.join("\n", command));
+            Process process = builder.start();
+            CountDownLatch finishLatch = startReadOutputThreads(process, String.join(" ", command));
+            finishLatch.await(); // TODO: Timeout? - can call `process.destroy` if it hangs
+            int exitCode = process.waitFor();
+            if (exitCode != 0)
+            {
+                String stdout = new String(outputBytes.get());
+                String stdErr = new String(errorOutput.get());
+                LOGGER.error("Spark STDOUT:\n*****{}\n*****", stdout);
```
yifan-c: Should it use `info` for stdout?

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/expansion/JoiningDoubleClusterTest.java
```diff
@@ -0,0 +1,194 @@
...
+        }
+    }
+
+    /**
+     * ByteBuddy helper for doubling cluster size failure scenario
+     */
+    @Shared
+    public static class BBHelperDoubleClusterSizeFailure
+    {
+        static CountDownLatch transientStateStart = new CountDownLatch(5);
```
yifan-c: nit: rename to `transitionalStateStart`
`transient` is a term can be confused with witness replica.

yifan-c (commented):
A few comments. I noticed some of my comments from the last round review were not yet addressed.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/ResiliencyTestBase.java
arjunashok: Addressed

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/IntegrationTestBase.java
arjunashok: Renamed to `LocalhostResolver`

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/ResiliencyTestBase.java
arjunashok: Well, this only happens when the spark job (process) fails, so error seems appropriate.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
arjunashok: makes sense.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
arjunashok: addressed

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
arjunashok: done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/StreamSession.java
arjunashok: Moved this up to the point where we extract the token-range for the task/partition. We do not need to do this check for overlapping ranges as is it implicitly validated if we do not have any resulting replicas.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TokenPartitioner.java
arjunashok: Yep, the previous check makes sense. Don't recall why this change was needed. Will revert.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ReplicaAwareFailureHandler.java
arjunashok: Yep, that should be a `Set`

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/RecordWriterTest.java
arjunashok: Addressed

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/TokenRangeMappingUtils.java
arjunashok: This is a new file though

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/ResiliencyTestBase.java
arjunashok: Addressed

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriteValidator.java
arjunashok: Updated to make the availability checks more cohesive.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
arjunashok: Agreed, this is out of scope from this PR. I believe this was applied by the code-style plugin. Will leave it as-is for now.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
arjunashok: Correct. Updated to utilize metadata.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
arjunashok: You're right. This will go away per previous comment

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
arjunashok: This is checked right before it is used.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
arjunashok: Testing only. Added

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CommitCoordinator.java
arjunashok: The expectation is for the instance to be available or transitioning at this point. If the instance does happen to be down, it should result in a commit failure, which is further evaluated against the CL requirements.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
arjunashok: Correct, the idea was to keep this class confined to CL validation

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
yifan-c: Thanks for adding the check!

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/ResiliencyTestBase.java
yifan-c: fair.

yifan-c (commented):
The patch looks good to me. Can you rebase onto trunk? Feel free to squash to a single commit before rebasing. I will give +1 once the rebased patch has a green CI run

arjunashok:
Reopen attempt

arjunashok:
Attempt to reopen PR

arjunashok:
Attempt to reopen PR

arjunashok:
Attempt to reopen PR

PR #19, opened 2023-11-14
CASSANDRA-19024 Fix bulk reading when using identifiers that need quotes

frankgh:
Cassandra in nature is case insensitive when creating identifiers (i.e. keyspace names, table names, column names, etc). We can define a case-sensitive identifier or we can use a reserved word as an identifier by quoting it during DDL creation.

In the analytics library, bulk reads fail when we encounter these identifiers. In this, commit, we fix the issue by properly propagating information about whether identifiers need to be quoted by exposing a new data frame option (`quote_identifiers`). When set to `true`, it will maybe quote the keyspace/table and it will properly be able to read data when these situations are encountered.

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/clients/SidecarInstanceImpl.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/clients/SidecarInstanceImpl.java
@@ -90,14 +90,14 @@ public class SidecarInstanceImpl implements Serializable, SidecarInstance
 
     private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundException
     {
-        LOGGER.warn("Falling back to JDK deserialization");
+        LOGGER.debug("Falling back to JDK deserialization");
         hostname = in.readUTF();
         port = in.readInt();
     }
 
     private void writeObject(ObjectOutputStream out) throws IOException, ClassNotFoundException
     {
-        LOGGER.warn("Falling back to JDK serialization");
+        LOGGER.debug("Falling back to JDK serialization");
         out.writeUTF(hostname);
         out.writeInt(port);
     }
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
@@ -56,21 +56,21 @@ public class BulkSparkConf implements Serializable
     private static final Logger LOGGER = LoggerFactory.getLogger(BulkSparkConf.class);
 
     public static final String JDK11_OPTIONS = " -Djdk.attach.allowAttachSelf=true"
-                                             + " --add-exports java.base/jdk.internal.misc=ALL-UNNAMED"
-                                             + " --add-exports java.base/jdk.internal.ref=ALL-UNNAMED"
-                                             + " --add-exports java.base/sun.nio.ch=ALL-UNNAMED"
-                                             + " --add-exports java.management.rmi/com.sun.jmx.remote.internal.rmi=ALL-UNNAMED"
-                                             + " --add-exports java.rmi/sun.rmi.registry=ALL-UNNAMED"
-                                             + " --add-exports java.rmi/sun.rmi.server=ALL-UNNAMED"
-                                             + " --add-exports java.sql/java.sql=ALL-UNNAMED"
-                                             + " --add-opens java.base/java.lang.module=ALL-UNNAMED"
-                                             + " --add-opens java.base/jdk.internal.loader=ALL-UNNAMED"
-                                             + " --add-opens java.base/jdk.internal.ref=ALL-UNNAMED"
-                                             + " --add-opens java.base/jdk.internal.reflect=ALL-UNNAMED"
-                                             + " --add-opens java.base/jdk.internal.math=ALL-UNNAMED"
-                                             + " --add-opens java.base/jdk.internal.module=ALL-UNNAMED"
-                                             + " --add-opens java.base/jdk.internal.util.jar=ALL-UNNAMED"
-                                             + " --add-opens jdk.management/com.sun.management.internal=ALL-UNNAMED";
+                                               + " --add-exports java.base/jdk.internal.misc=ALL-UNNAMED"
+                                               + " --add-exports java.base/jdk.internal.ref=ALL-UNNAMED"
+                                               + " --add-exports java.base/sun.nio.ch=ALL-UNNAMED"
+                                               + " --add-exports java.management.rmi/com.sun.jmx.remote.internal.rmi=ALL-UNNAMED"
+                                               + " --add-exports java.rmi/sun.rmi.registry=ALL-UNNAMED"
+                                               + " --add-exports java.rmi/sun.rmi.server=ALL-UNNAMED"
+                                               + " --add-exports java.sql/java.sql=ALL-UNNAMED"
+                                               + " --add-opens java.base/java.lang.module=ALL-UNNAMED"
+                                               + " --add-opens java.base/jdk.internal.loader=ALL-UNNAMED"
+                                               + " --add-opens java.base/jdk.internal.ref=ALL-UNNAMED"
+                                               + " --add-opens java.base/jdk.internal.reflect=ALL-UNNAMED"
+                                               + " --add-opens java.base/jdk.internal.math=ALL-UNNAMED"
+                                               + " --add-opens java.base/jdk.internal.module=ALL-UNNAMED"
+                                               + " --add-opens java.base/jdk.internal.util.jar=ALL-UNNAMED"
+                                               + " --add-opens jdk.management/com.sun.management.internal=ALL-UNNAMED";
 
     public static final int DEFAULT_NUM_SPLITS = -1;
     public static final int DEFAULT_HTTP_CONNECTION_TIMEOUT = 100_000;
@@ -128,6 +128,7 @@ public class BulkSparkConf implements Serializable
     public final int commitThreadsPerInstance;
     protected final int effectiveSidecarPort;
     protected final int userProvidedSidecarPort;
+    public boolean quoteIdentifiers;
     protected boolean useOpenSsl;
     protected int ringRetryCount;
 
@@ -166,6 +167,7 @@ public class BulkSparkConf implements Serializable
         this.ringRetryCount = getInt(RING_RETRY_COUNT, DEFAULT_RING_RETRY_COUNT);
         this.ttl = MapUtils.getOrDefault(options, WriterOptions.TTL.name(), null);
         this.timestamp = MapUtils.getOrDefault(options, WriterOptions.TIMESTAMP.name(), null);
+        this.quoteIdentifiers = MapUtils.getBoolean(options, WriterOptions.QUOTE_IDENTIFIERS.name(), false, "quote identifiers");
         validateEnvironment();
     }
 
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
@@ -56,21 +56,21 @@ public class BulkSparkConf implements Serializable
     private static final Logger LOGGER = LoggerFactory.getLogger(BulkSparkConf.class);
 
     public static final String JDK11_OPTIONS = " -Djdk.attach.allowAttachSelf=true"
-                                               + " --add-exports java.base/jdk.internal.misc=ALL-UNNAMED"
-                                               + " --add-exports java.base/jdk.internal.ref=ALL-UNNAMED"
-                                               + " --add-exports java.base/sun.nio.ch=ALL-UNNAMED"
-                                               + " --add-exports java.management.rmi/com.sun.jmx.remote.internal.rmi=ALL-UNNAMED"
-                                               + " --add-exports java.rmi/sun.rmi.registry=ALL-UNNAMED"
-                                               + " --add-exports java.rmi/sun.rmi.server=ALL-UNNAMED"
-                                               + " --add-exports java.sql/java.sql=ALL-UNNAMED"
-                                               + " --add-opens java.base/java.lang.module=ALL-UNNAMED"
-                                               + " --add-opens java.base/jdk.internal.loader=ALL-UNNAMED"
-                                               + " --add-opens java.base/jdk.internal.ref=ALL-UNNAMED"
-                                               + " --add-opens java.base/jdk.internal.reflect=ALL-UNNAMED"
-                                               + " --add-opens java.base/jdk.internal.math=ALL-UNNAMED"
-                                               + " --add-opens java.base/jdk.internal.module=ALL-UNNAMED"
-                                               + " --add-opens java.base/jdk.internal.util.jar=ALL-UNNAMED"
-                                               + " --add-opens jdk.management/com.sun.management.internal=ALL-UNNAMED";
+                                             + " --add-exports java.base/jdk.internal.misc=ALL-UNNAMED"
+                                             + " --add-exports java.base/jdk.internal.ref=ALL-UNNAMED"
+                                             + " --add-exports java.base/sun.nio.ch=ALL-UNNAMED"
+                                             + " --add-exports java.management.rmi/com.sun.jmx.remote.internal.rmi=ALL-UNNAMED"
+                                             + " --add-exports java.rmi/sun.rmi.registry=ALL-UNNAMED"
+                                             + " --add-exports java.rmi/sun.rmi.server=ALL-UNNAMED"
+                                             + " --add-exports java.sql/java.sql=ALL-UNNAMED"
+                                             + " --add-opens java.base/java.lang.module=ALL-UNNAMED"
+                                             + " --add-opens java.base/jdk.internal.loader=ALL-UNNAMED"
+                                             + " --add-opens java.base/jdk.internal.ref=ALL-UNNAMED"
+                                             + " --add-opens java.base/jdk.internal.reflect=ALL-UNNAMED"
+                                             + " --add-opens java.base/jdk.internal.math=ALL-UNNAMED"
+                                             + " --add-opens java.base/jdk.internal.module=ALL-UNNAMED"
+                                             + " --add-opens java.base/jdk.internal.util.jar=ALL-UNNAMED"
+                                             + " --add-opens jdk.management/com.sun.management.internal=ALL-UNNAMED";
 
     public static final int DEFAULT_NUM_SPLITS = -1;
     public static final int DEFAULT_HTTP_CONNECTION_TIMEOUT = 100_000;
@@ -128,7 +128,6 @@ public class BulkSparkConf implements Serializable
     public final int commitThreadsPerInstance;
     protected final int effectiveSidecarPort;
     protected final int userProvidedSidecarPort;
-    public boolean quoteIdentifiers;
     protected boolean useOpenSsl;
     protected int ringRetryCount;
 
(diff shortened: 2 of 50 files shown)


cassandra-four-zero/src/main/java/org/apache/cassandra/spark/reader/SchemaBuilder.java
```diff
@@ -519,7 +519,7 @@ private List<CqlField> buildFields(TableMetadata metadata, Map<String, CqlField.
             boolean isPartitionKey = col.isPartitionKey();
             boolean isClusteringColumn = col.isClusteringColumn();
             boolean isStatic = col.isStatic();
-            String name = col.name.toCQLString();
+            String name = col.name.toString();
```
frankgh, line 523: `toCQLString` gives us the quoted string, but we immediately drop the quotes when creating the field. So we should just return the unquoted string, which is what we use

cassandra-bridge/src/main/java/org/apache/cassandra/spark/data/CqlField.java
```diff
@@ -263,7 +263,7 @@ public CqlField(boolean isPartitionKey,
         this.isPartitionKey = isPartitionKey;
         this.isClusteringColumn = isClusteringColumn;
         this.isStaticColumn = isStaticColumn;
-        this.name = name.replaceAll("\"", "");
+        this.name = name;
```
frankgh, line 266: we no longer need to drop the quotes, since we no longer pass the quoted string from here: https://github.com/apache/cassandra-analytics/pull/19/files#diff-143d0bc336629bfe78bf9f0c64d83803701d38b22a4be736323277862b784b25R522

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/utils/CqlUtils.java
```diff
@@ -95,25 +95,6 @@ else if (schema.charAt(index) == '(')
         return schema.substring(0, index + 1);
     }
 
-    /**
-     * The schema might contain quotes, but on-disk the table path doesn't have quotes, so causes problems for create/list snapshot
```
frankgh, line 105: this is now handled by Sidecar properly

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
```diff
@@ -950,204 +973,7 @@ protected void await(CountDownLatch latch)
         }
     }
 
-    public static final class ClientConfig
```
frankgh, line 925: moved this to its own class

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
```diff
@@ -45,5 +45,10 @@ public enum WriterOptions implements WriterOption
     ROW_BUFFER_MODE,
     SSTABLE_DATA_SIZE_IN_MB,
     TTL,
-    TIMESTAMP
+    TIMESTAMP,
+    /**
+     * Option that specifies whether the identifiers (i.e. keyspace, table name, column names) should be quoted to
+     * support mixed case and reserved keyword names for these fields.
+     */
+    QUOTE_IDENTIFIERS,
```
yifan-c: why it is a writer option?

cassandra-analytics-core/src/main/java/org/apache/cassandra/clients/SidecarInstanceImpl.java
```diff
@@ -90,14 +90,14 @@ public String toString()
 
     private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundException
     {
-        LOGGER.warn("Falling back to JDK deserialization");
+        LOGGER.debug("Falling back to JDK deserialization");
```
yifan-c, line 93: +1 on changing the log level to debug. If warning on serialization method is desired, it should be logged separated and warn just once.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -56,21 +56,21 @@ public class BulkSparkConf implements Serializable
...
+                                               + " --add-exports java.rmi/sun.rmi.server=ALL-UNNAMED"
+                                               + " --add-exports java.sql/java.sql=ALL-UNNAMED"
+                                               + " --add-opens java.base/java.lang.module=ALL-UNNAMED"
+                                               + " --add-opens java.base/jdk.internal.loader=ALL-UNNAMED"
+                                               + " --add-opens java.base/jdk.internal.ref=ALL-UNNAMED"
+                                               + " --add-opens java.base/jdk.internal.reflect=ALL-UNNAMED"
+                                               + " --add-opens java.base/jdk.internal.math=ALL-UNNAMED"
+                                               + " --add-opens java.base/jdk.internal.module=ALL-UNNAMED"
+                                               + " --add-opens java.base/jdk.internal.util.jar=ALL-UNNAMED"
+                                               + " --add-opens jdk.management/com.sun.management.internal=ALL-UNNAMED";
```
yifan-c: In principle, we should not change the code only for the sake of correcting the style/format. The noise it creates voids the little benefit provided.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -128,6 +128,7 @@ public class BulkSparkConf implements Serializable
     public final int commitThreadsPerInstance;
     protected final int effectiveSidecarPort;
     protected final int userProvidedSidecarPort;
+    public boolean quoteIdentifiers;
```
yifan-c: Can it be `final`? It does not seem to be updated.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
```diff
@@ -239,12 +241,23 @@ public void initialize(@NotNull ClientConfig options)
...
         Preconditions.checkArgument(table != null, "Table must be non-null for Cassandra Bulk Reader");
+        ShutdownHookManager.addShutdownHook(org.apache.spark.util.ShutdownHookManager.TEMP_DIR_SHUTDOWN_PRIORITY(),
+                                            ScalaFunctions.wrapLambda(() -> shutdownHook(options)));
+
+        NodeSettings nodeSettings = sidecar.nodeSettings().get();
+        String cassandraVersion = getEffectiveCassandraVersionForRead(clusterConfig, nodeSettings);
+        Partitioner partitioner = Partitioner.from(nodeSettings.partitioner());
+        bridge = CassandraBridgeFactory.get(cassandraVersion);
+        // optionally quote identifiers if the option has been set, we need an instance for the bridge
+        maybeQuoteKeyspaceAndTable();
```
yifan-c, line 252: `maybeQuoteKeyspaceAndTable` was already called in the constructor. I guess it is preferred to call it here. But it does not harm with running the method twice.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
```diff
@@ -343,7 +346,7 @@ private CompletionStage<Map<String, AvailabilityHint>> createSnapshot(RingRespon
                                 snapshotName, keyspace, table, datacenter, ringEntry.fqdn());
```
yifan-c: The log message should use `maybeQuotedKeyspace` and `maybeQuotedTable` instead.
The parameters in the log message above (`LOGGER.warn("Skip snapshot creating when node is joining or down "`) should be updated too. 
And the log messages in `listInstance` method.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
```diff
@@ -343,7 +346,7 @@ private CompletionStage<Map<String, AvailabilityHint>> createSnapshot(RingRespon
...
                     SidecarInstance sidecarInstance = new SidecarInstanceImpl(ringEntry.fqdn(), sidecarClientConfig.effectivePort());
                     createSnapshotFuture = sidecar
-                                           .createSnapshot(sidecarInstance, keyspace, table, snapshotName)
+                                           .createSnapshot(sidecarInstance, maybeQuotedKeyspace, maybeQuotedTable, snapshotName)
```
yifan-c, line 345: What if `maybeQuotedKeyspace` and/or `maybeQuotedTable` are null in this method. You added the check in the shutdown hook, should it be checked here?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
```diff
@@ -888,14 +911,20 @@ protected void dialHome(@NotNull ClientConfig options)
 
     protected void clearSnapshot(Set<? extends SidecarInstance> clusterConfig, @NotNull ClientConfig options)
     {
+        if (maybeQuotedKeyspace == null || maybeQuotedTable == null)
+        {
+            LOGGER.info("Bridge was never initialized. Skipping clearing snapshots. This implies snapshots were never created");
```
yifan-c: Should it be a warn?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataSourceHelper.java
```diff
@@ -90,7 +94,7 @@ public static DataLayer getDataLayer(
         }
     }
 
-    protected static Cache<Map<String, String>, CassandraDataLayer> getCassandraDataLayerCache()
+    public static Cache<Map<String, String>, CassandraDataLayer> getCassandraDataLayerCache()
```
yifan-c: Why are the methods (including others in the file) are changed from `protected` to `public`?

cassandra-analytics-core/src/main/spark3/org/apache/cassandra/spark/sparksql/CassandraScanBuilder.java
```diff
@@ -135,17 +136,16 @@ private List<PartitionKeyFilter> buildPartitionKeyFilters()
...
+        if (partitionKeyValues.isEmpty())
         {
-            return new ArrayList<>();
+            return Collections.emptyList();
         }
+
+        List<List<String>> orderedValues = partitionKeyColumnNames.stream().map(partitionKeyValues::get).collect(Collectors.toList());
+        return FilterUtils.cartesianProduct(orderedValues).stream()
+                          .map(this::buildFilter)
+                          .collect(Collectors.toList());
```
yifan-c: It looks like only the if and else blocks are switched. I would discourage such refactoring. 
We can keep this change since it is done, but let's refrain from doing so in the future patches.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/EndToEndTests.java
```diff
@@ -154,15 +154,15 @@ public void testBasicSingleClusteringKey(CassandraBridge bridge)
...
                 Tester.builder(TestSchema.builder()
                                          .withPartitionKey("a", bridge.bigint())
                                          .withClusteringKey("b", clusteringKeyType)
                                          .withColumn("c", bridge.bigint())
                                          .withSortOrder(sortOrder))
                       .withExpectedRowCountPerSSTable(Tester.DEFAULT_NUM_ROWS)
-                      .run()
-            );
+                      .run();
+            });
```
yifan-c: unnecessary refactoring and many others in this file.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/EndToEndTests.java
```diff
@@ -1851,19 +1848,19 @@ public void testUdtsWithNulls(CassandraBridge bridge)
                   for (long pk = 0; pk < Tester.DEFAULT_NUM_ROWS; pk++)
                   {
                       Map<String, Object> value = ImmutableMap.of(
-                            pk < midPoint ? "a" : "b", RandomUtils.randomValue(bridge.text()).toString(),
-                            "c", RandomUtils.randomValue(bridge.text()).toString());
+                      pk < midPoint ? "a" : "b", RandomUtils.randomValue(bridge.text()).toString(),
+                      "c", RandomUtils.randomValue(bridge.text()).toString());
```
yifan-c: It is hard to read. It is not obvious that those 2 lines are the parameters for the `of` method without indentations. 
Please restore the formatting.

cassandra-bridge/src/main/java/org/apache/cassandra/spark/data/DataLayer.java
```diff
@@ -171,8 +171,8 @@ public CassandraVersion version()
     public abstract boolean isInPartition(int partitionId, BigInteger token, ByteBuffer key);
 
     public List<PartitionKeyFilter> partitionKeyFiltersInRange(
-            int partitionId,
-            List<PartitionKeyFilter> partitionKeyFilters) throws NoMatchFoundException
+    int partitionId,
+    List<PartitionKeyFilter> partitionKeyFilters) throws NoMatchFoundException
```
yifan-c, line 169: It is still hard to read after the change. Please restore the format. The change is unrelated and unnecessary.

cassandra-bridge/src/main/java/org/apache/cassandra/spark/data/CqlField.java
yifan-c: > we no longer need to drop the quotes, since we no longer pass the quoted string from here: https://github.com/apache/cassandra-analytics/pull/19/files#diff-143d0bc336629bfe78bf9f0c64d83803701d38b22a4be736323277862b784b25R522

This comment is confusing to me. The new test cases in `EndToEndTests` pass quoted names, and the tests fail if quotation marks are removed. 
For UDT, the names are always unquoted, and there is `fieldsString` to quote back if needed. But it is not true for plain `CqlField`. There is some inconsistency.

yifan-c (commented):
Overall looks good to me. There are too much noises from unrelated reformatting. Please do not have them in future patch. If there is a 100% required refactoring, it is clearer to have it in a separate patch.

yifan-c:
> Cassandra in nature is case insensitive when creating identifiers (i.e. keyspace names, table names, column names, etc). We can define a case-sensitive identifier or we can use a reserved word as an identifier by quoting it during DDL creation.

I won't say that Cassandra is case-insensitive in nature. The statement is not helpful to understand the commit, which is essentially adding a new option to toggle quote the identifiers whenever desired. That said, I won't be insisting on removing the line. But I would just write it explicitly that "Cassandra retreats all identifiers as they are in lower case unless quoted".

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
frankgh: this should go in CASSANDRA-19031 instead. I will clean it up here

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
frankgh: yeah, the constructor that calls this method is only called in the executors. For the driver, it will use this method to initialize the keyspace and table

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/EndToEndTests.java
frankgh: I'll revert the unnecessary changes

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataSourceHelper.java
frankgh: the change is not needed, I will revert

frankgh:
> But I would just write it explicitly that "Cassandra retreats all identifiers as they are in lower case unless quoted

Yeah, this makes more sense, I will update the commit message

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
frankgh: changes related to bulk writes, moving to another PR.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
frankgh: if the bridge is initialized correctly, the fields will never be null when we reach this point.

frankgh: however, the cleanup of snapshots happens on a different thread, if an exception occurs before the bridge has been initialized (i.e an error occurs when getting the cassandra version `String cassandraVersion = getEffectiveCassandraVersionForRead(clusterConfig, nodeSettings);` ) then the shutdown hook will be triggered and it will try to cleanup snapshots, even though the snapshot was never created in the first place

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/EndToEndTests.java
yifan-c: I am fine with keeping them, if it is trouble to revert all of them.

cassandra-bridge/src/main/java/org/apache/cassandra/spark/data/CqlField.java
frankgh: Handling of UDTs is different from handling of keyspace / table. For UDTs, the bridge was already taking care of preserving the quotes. However, for user facing options (the keyspace and table specifically), we need special handling based on whether the user explicitly requests to quote identifiers

cassandra-bridge/src/main/java/org/apache/cassandra/spark/data/DataLayer.java
frankgh: reverted

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/EndToEndTests.java
frankgh: reverted

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/EndToEndTests.java
frankgh: reverted them already

cassandra-analytics-core/src/main/spark3/org/apache/cassandra/spark/sparksql/CassandraScanBuilder.java
frankgh: reverted

cassandra-bridge/src/main/java/org/apache/cassandra/spark/data/CqlField.java
frankgh: ok, took a deeper look. I think the confusion is mainly on the EndToEnd testing code. So what I've done is address the test setup in testing code. I have made it so that we always quote the fields, when quotes are needed for testing purposes. 

I have also added a test case that validates the column filter (`PartialRowBuilder` class). That way we guarantee that we are able to retrieve columns that are requested in the schema where columns are case sensitive in Cassandra

frankgh:
Closed via https://github.com/apache/cassandra-analytics/commit/457b36bcb3c8a865cca83ca6c402246798113ab4

PR #20, opened 2023-11-16
CASSANDRA-19031: Fix bulk writing when using identifiers that need quotes

frankgh:
Cassandra treats all identifiers as lower case unless explicitly quoted by the users,
(i.e. keyspace names, table names, column names, etc). We can define a case-sensitive
identifier or we can use a reserved word as an identifier by quoting it during DDL
creation.

In the analytics library, bulk writing fails when we encounter these identifiers. In
this commit, we fix the issue by property propagating the information about whether
identifiers need to be quoted by exposing a new dataframe option (`quote_identifiers`).
When set to `true`, it will _maybe_ quote the keyspace/table/column names and it will
properly be able to write data when using mixed-case or reserved words in the
identifiers.

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
@@ -34,6 +34,8 @@ import com.google.common.annotations.VisibleForTesting;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.cassandra.bridge.CassandraBridge;
+import org.apache.cassandra.bridge.CassandraBridgeFactory;
 import org.apache.cassandra.bridge.CassandraVersionFeatures;
 import org.apache.cassandra.clients.Sidecar;
 import org.apache.cassandra.sidecar.client.SidecarInstance;
@@ -52,6 +54,8 @@ import org.apache.cassandra.spark.data.partitioner.Partitioner;
 import org.apache.cassandra.spark.utils.CqlUtils;
 import org.apache.cassandra.spark.utils.FutureUtils;
 import org.jetbrains.annotations.NotNull;
+
+import static org.apache.cassandra.bridge.CassandraBridgeFactory.maybeQuotedIdentifier;
 
 public class CassandraClusterInfo implements ClusterInfo, Closeable
 {
@@ -233,7 +237,9 @@ public class CassandraClusterInfo implements ClusterInfo, Closeable
 
     protected String getCurrentKeyspaceSchema() throws Exception
     {
-        SchemaResponse schemaResponse = getCassandraContext().getSidecarClient().schema(conf.keyspace).get();
+        SchemaResponse schemaResponse = getCassandraContext().getSidecarClient()
+                                                             .schema(maybeQuotedIdentifier(bridge(), conf.quoteIdentifiers, conf.keyspace))
+                                                             .get();
         return schemaResponse.schema();
     }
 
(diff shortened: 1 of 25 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -535,6 +548,15 @@ protected boolean isReplacement(RingEntry ringEntry)
         return false;
     }
 
+    protected CassandraBridge bridge()
+    {
+        if (bridge == null)
+        {
+            bridge = CassandraBridgeFactory.get(getLowestCassandraVersion());
```
frankgh: underlying ConcurrentHashMap in factory guarantees single bridge initialization here, so no need for synchronization

yifan-c: In that case, why not just call `CassandraBridgeFactory.get(getLowestCassandraVersion())`? Given the value effectively cached already.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/TableSchemaTest.java
```diff
@@ -231,4 +234,10 @@ private TableSchemaTestCommon.MockTableSchemaBuilder getValidSchemaBuilder()
                 .withWriteMode(WriteMode.INSERT)
                 .withDataFrameSchema(validDataFrameSchema);
     }
+
+    // Removes the unique table name to make validation consistent
+    private static String replaceTableName(String statement)
```
yifan-c: nit: how about rename to `trimTestTableId`?

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/QuoteIdentifiersWriteTest.java
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * <p>These tests exercise a full integration test, which includes testing Sidecar behavior when dealing with quoted
+ * identifiers.
+ */
+@ExtendWith(VertxExtension.class)
+class QuoteIdentifiersWriteTest extends SparkIntegrationTestBase
+{
+    static final int ROW_COUNT = 10_000;
+
+    @CassandraIntegrationTest(nodesPerDc = 3, gossip = true)
```
yifan-c: Does the test quire 3 nodes? If not, let's use 1 node to consume less resources.

yifan-c (commented):
lgtm overall. A few nits. Feel free to ignore.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
frankgh: yeah, having a local cache saves look ups from the map. But lookups are fast anyway, so I think we can remove the local field used as a cache

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/QuoteIdentifiersWriteTest.java
frankgh: good point, we don't need 3 instances. I will lower it to 1

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/TableSchemaTest.java
frankgh: table Id might cause confusion with Cassandra's table ID. I will rename it to `trimUniqueTableName`

frankgh:
> lgtm overall. A few nits. Feel free to ignore.

Thanks for the review. I've addressed your comments.

frankgh:
Squashed and lowered concurrency for JDK8 tests

frankgh:
Closed via https://github.com/apache/cassandra-analytics/commit/c7c3bbca2c7cb415b39689e924fa2357c239f043

PR #22, opened 2023-11-30
[CASSANDRA-19377] Startup Validation Failures when Checking Sidecar Connectivity (Analytics)

5:
Increased the overall timeout for Sidecar health checks from current 30 seconds to 5 minutes and made it configurable.

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/validation/CassandraValidation.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/validation/CassandraValidation.java
@@ -29,7 +29,7 @@ import org.apache.cassandra.sidecar.common.data.HealthResponse;
  */
 public class CassandraValidation implements StartupValidation
 {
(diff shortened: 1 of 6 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/validation/CassandraValidation.java
```diff
@@ -29,7 +29,7 @@
  */
 public class CassandraValidation implements StartupValidation
 {
-    private static final int TIMEOUT_SECONDS = 30;
+    private static final int TIMEOUT_SECONDS = 300;
```
frankgh: should these be configurable?

5: Made timeouts for Sidecar/Cassandra startup validations configurable.

frankgh (approved):
+1 Looks good to me

PR #25, opened 2023-12-12
[Updated] Cassandra 18852: Make bulk writer resilient to cluster resize events

arjunashok:
This is a copy of the PR: https://github.com/apache/cassandra-analytics/pull/17, which includes the review comments.

### Changes:
Replaces dependency on Sidecar ring endpoint to use the token-ranges endpoint
Queries sidecar token-ranges endpoint to get the instance metadata, token-ranges and write replica-sets which include the pending ranges for transient nodes.
Data is written to RF replicas calculated including the transient nodes.
RingMonitor and other checks to fail-fast on detecting resizing cluster are removed
Ring changes within the scope of a task are checked to allow the task to retry.

### Testing
Added unit tests and in-jvm-dtest based integration tests for cluster expansion, shrink, node-movement and replacement scenarios
Validates partitioned data on each recipient instance of the write, including transitioning nodes.


frankgh:
closed via https://github.com/apache/cassandra-analytics/commit/0aaf5659028dd874c8d666c636f11eae63c429e6

PR #27, opened 2023-12-20
CASSANDRA-19212: Fix forced Spark job timeout in ResiliencyTestBase

JeetKunDoug:


cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/ResiliencyTestBase.java
```diff
@@ -375,10 +375,22 @@ protected void bulkWriteData(ConsistencyLevel writeCL, QualifiedName schema)
             if (!completed)
             {
                 process.destroyForcibly();
-                finishLatch.await(10, TimeUnit.MINUTES);
+                // The process doesn't necessarily finish right after destroyForcibly, but will exit.
+                process.waitFor(1, TimeUnit.MINUTES);
```
frankgh, line 380: maybe add a comment on why we only let the spawned spark process run for 1 minute. And the `errorMessage` in line 380 will need to be updated as well

JeetKunDoug: The comment on 380 is still valid - after 10 minutes, we attempt to kill the process. I did add "and was killed" along with a bit more commenting on the additional 1 minute of wait.

JeetKunDoug:
Closing the PR for now as we aren't running Spark jobs in a separate process any more. May revisit if we decide to pull that feature back in to the Analytics library.

PR #29, opened 2024-01-05
CASSANDRA-19251 Speed up integration tests

frankgh:
This commit introduces an opinionated way to run integration tests where a test class reuses the same in-jvm dtest cluster, and it offers certain ordering that help running tests faster.

The test setup does the following:
- Find the Cassandra version to run
- Provision a cluster for the test
- Initialize schemas required for tests
- Start the Sidecar service

The above approach guarantess that Sidecar is ready once the setup method completes, which means we no longer need to spend time waiting for schema propagation. This optimization also helps in reducing test time.

The drawback of this approach is that if we need the cluster to be in some state for testing, for example a node needs to be in joining state while executing the bulk test then, that cluster can only be used for tests in that state. Which means that testing different states of the cluster requires a new test class.

Patch by Francisco Guerrero for CASSANDRA-19251.


cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    @AfterAll
+    protected void tearDown() throws InterruptedException
+    {
+        beforeSidecarStop();
+        stopSidecar();
+        beforeClusterShutdown();
+        closeCluster();
+        beforeTearDownEnd();
```
yifan-c: nit: how about `afterClusterShutdown`?

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
```diff
@@ -0,0 +1,417 @@
...
+
+    protected void createTestKeyspace(QualifiedName name, Map<String, Integer> rf)
+    {
+        createTestKeyspace(name.maybeQuotedKeyspace(), rf);
+    }
+
+    protected void createTestKeyspace(String keyspace, Map<String, Integer> rf)
+    {
+        cluster.schemaChangeIgnoringStoppedInstances("CREATE KEYSPACE IF NOT EXISTS " + keyspace
+                                                     + " WITH REPLICATION = { 'class' : 'NetworkTopologyStrategy', " +
```
yifan-c, line 206: I guess it is same to assume that we always want to test with `NetworkTopologyStrategy`

yifan-c (approved):
LGTM. One nit on renaming and feel free to ignore.

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
frankgh: yeah, I didn't change this as all tests are currently using `NetworkTopologyStrategy`, but this is open to modification in the future as needed

frankgh:
Closed via https://github.com/apache/cassandra-analytics/commit/550bdfa1c6082537e2cfb93449128a61dbe3a1fb

PR #31, opened 2024-01-16
CASSANDRA-19273 Allow setting TTL for snapshots created through reader

sarankk:
We want to allow analytics users to set TTL option for snapshots created through bulk reader. This is to aid in clearing snapshots, especially in case of job failures, when clearSnapshot call is not executed.

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
@@ -290,8 +290,10 @@ public class CassandraDataLayer extends PartitionedDataLayer implements StartupV
 
     protected void shutdownHook(ClientConfig options)
     {
-        // Preserves previous behavior, but we may just want to check for the clearSnapshot option in the future
-        if (options.clearSnapshot())
+        ClientConfig.ClearSnapshotStrategy clearSnapshotStrategy = options.clearSnapshotStrategy();
+        boolean clearOnCompletion = ClientConfig.ClearSnapshotStrategyType.onCompletion.equals(clearSnapshotStrategy.type());
+        boolean clearOnCompletionOrTTL = ClientConfig.ClearSnapshotStrategyType.onCompletionOrTTL.equals(clearSnapshotStrategy.type());
+        if (clearOnCompletion || clearOnCompletionOrTTL)
         {
             if (options.createSnapshot())
             {
@@ -304,6 +306,10 @@ public class CassandraDataLayer extends PartitionedDataLayer implements StartupV
                             + "snapshotName={} keyspace={} table={} dc={}",
                             snapshotName, keyspace, table, datacenter);
             }
+        }
+        if (ClientConfig.ClearSnapshotStrategyType.TTL.equals(clearSnapshotStrategy.type()))
+        {
+            LOGGER.warn("Skipping clearing snapshot because snapshot TTL was set to {}", clearSnapshotStrategy.ttl());
         }
 
         try
@@ -341,15 +347,9 @@ public class CassandraDataLayer extends PartitionedDataLayer implements StartupV
                     LOGGER.info("Creating snapshot on instance snapshotName={} keyspace={} table={} datacenter={} fqdn={}",
                                 snapshotName, maybeQuotedKeyspace, maybeQuotedTable, datacenter, ringEntry.fqdn());
                     SidecarInstance sidecarInstance = new SidecarInstanceImpl(ringEntry.fqdn(), sidecarClientConfig.effectivePort());
-                    String resolvedSnapshotTtl = options.clearSnapshot() ? options.effectiveSnapshotTtl() : null;
-                    if (!options.clearSnapshot() && options.userProvidedSnapshotTtl() != null)
-                    {
-                        LOGGER.warn("Snapshot TTL option was provided along with Clear Snapshot option, bulk reader" +
-                                    "can honor only one of the 2. Clear Snapshot takes precedence over Snapshot TTL, " +
-                                    "hence snapshot {} will not be removed after the specified snapshot TTL option", snapshotName);
-                    }
+                    ClientConfig.ClearSnapshotStrategy clearSnapshotStrategy = options.clearSnapshotStrategy();
                     createSnapshotFuture = sidecar.createSnapshot(sidecarInstance, maybeQuotedKeyspace,
-                                                                  maybeQuotedTable, snapshotName, resolvedSnapshotTtl)
+                                                                  maybeQuotedTable, snapshotName, clearSnapshotStrategy.ttl())
                                                   .handle((resp, throwable) -> {
                                                       if (throwable == null)
                                                       {
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
@@ -350,9 +350,9 @@ public class CassandraDataLayer extends PartitionedDataLayer implements StartupV
                     String resolvedSnapshotTtl = clearSnapshot ? effectiveSnapshotTtl : null;
                     if (!clearSnapshot && userProvidedSnapshotTtl != null)
                     {
-                        LOGGER.error("Snapshot TTL option was provided along with Clear Snapshot option, bulk reader" +
-                                     "can honor only one of the 2. Clear Snapshot takes precedence over Snapshot TTL, " +
-                                     "hence snapshot {} will not be removed after the specified snapshot TTL option", snapshotName);
+                        LOGGER.warn("Snapshot TTL option was provided along with Clear Snapshot option, bulk reader" +
+                                    "can honor only one of the 2. Clear Snapshot takes precedence over Snapshot TTL, " +
+                                    "hence snapshot {} will not be removed after the specified snapshot TTL option", snapshotName);
                     }
                     createSnapshotFuture = sidecar.createSnapshot(sidecarInstance, maybeQuotedKeyspace,
                                                                   maybeQuotedTable, snapshotName, resolvedSnapshotTtl)
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
@@ -112,7 +112,8 @@ public class CassandraDataLayer extends PartitionedDataLayer implements StartupV
                 .build();
 
     protected String snapshotName;
-    protected transient String snapshotTtl;
+    protected transient String userProvidedSnapshotTtl;
+    protected transient String effectiveSnapshotTtl;
     protected transient boolean clearSnapshot;
     protected boolean quoteIdentifiers;
     protected String keyspace;
@@ -145,7 +146,8 @@ public class CassandraDataLayer extends PartitionedDataLayer implements StartupV
     {
         super(options.consistencyLevel(), options.datacenter());
         this.snapshotName = options.snapshotName();
-        this.snapshotTtl = options.snapshotTtl();
+        this.userProvidedSnapshotTtl = options.userProvidedSnapshotTtl();
+        this.effectiveSnapshotTtl = options.effectiveSnapshotTtl();
         this.clearSnapshot = options.clearSnapshot();
         this.keyspace = options.keyspace();
         this.table = options.table();
@@ -345,12 +347,15 @@ public class CassandraDataLayer extends PartitionedDataLayer implements StartupV
                     LOGGER.info("Creating snapshot on instance snapshotName={} keyspace={} table={} datacenter={} fqdn={}",
                                 snapshotName, maybeQuotedKeyspace, maybeQuotedTable, datacenter, ringEntry.fqdn());
                     SidecarInstance sidecarInstance = new SidecarInstanceImpl(ringEntry.fqdn(), sidecarClientConfig.effectivePort());
-                    CompletableFuture<Void> createSnapshotCall
-                    = clearSnapshot
-                      ? sidecar.createSnapshot(sidecarInstance, maybeQuotedKeyspace, maybeQuotedTable, snapshotName,
-                                               snapshotTtl)
-                      : sidecar.createSnapshot(sidecarInstance, maybeQuotedKeyspace, maybeQuotedTable, snapshotName);
-                    createSnapshotFuture = createSnapshotCall
+                    String resolvedSnapshotTtl = clearSnapshot ? effectiveSnapshotTtl : null;
+                    if (!clearSnapshot && userProvidedSnapshotTtl != null)
+                    {
+                        LOGGER.error("Snapshot TTL option was provided along with Clear Snapshot option, bulk reader" +
+                                     "can honor only one of the 2. Clear Snapshot takes precedence over Snapshot TTL, " +
+                                     "hence snapshot {} will not be cleared with TTL", snapshotName);
+                    }
+                    createSnapshotFuture = sidecar.createSnapshot(sidecarInstance, maybeQuotedKeyspace,
+                                                                  maybeQuotedTable, snapshotName, resolvedSnapshotTtl)
                                            .handle((resp, throwable) -> {
                                                if (throwable == null)
                                                {
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
@@ -112,9 +112,6 @@ public class CassandraDataLayer extends PartitionedDataLayer implements StartupV
                 .build();
 
     protected String snapshotName;
-    protected transient String userProvidedSnapshotTtl;
-    protected transient String effectiveSnapshotTtl;
-    protected transient boolean clearSnapshot;
     protected boolean quoteIdentifiers;
     protected String keyspace;
     protected String table;
@@ -146,9 +143,6 @@ public class CassandraDataLayer extends PartitionedDataLayer implements StartupV
     {
         super(options.consistencyLevel(), options.datacenter());
         this.snapshotName = options.snapshotName();
-        this.userProvidedSnapshotTtl = options.userProvidedSnapshotTtl();
-        this.effectiveSnapshotTtl = options.effectiveSnapshotTtl();
-        this.clearSnapshot = options.clearSnapshot();
         this.keyspace = options.keyspace();
         this.table = options.table();
         this.quoteIdentifiers = options.quoteIdentifiers();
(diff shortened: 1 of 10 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
```diff
@@ -341,8 +347,15 @@ private CompletionStage<Map<String, AvailabilityHint>> createSnapshot(RingRespon
                     LOGGER.info("Creating snapshot on instance snapshotName={} keyspace={} table={} datacenter={} fqdn={}",
                                 snapshotName, maybeQuotedKeyspace, maybeQuotedTable, datacenter, ringEntry.fqdn());
                     SidecarInstance sidecarInstance = new SidecarInstanceImpl(ringEntry.fqdn(), sidecarClientConfig.effectivePort());
-                    createSnapshotFuture = sidecar
-                                           .createSnapshot(sidecarInstance, maybeQuotedKeyspace, maybeQuotedTable, snapshotName)
+                    String resolvedSnapshotTtl = clearSnapshot ? effectiveSnapshotTtl : null;
+                    if (!clearSnapshot && userProvidedSnapshotTtl != null)
+                    {
+                        LOGGER.error("Snapshot TTL option was provided along with Clear Snapshot option, bulk reader" +
```
frankgh: warn instead?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
```diff
@@ -341,8 +347,15 @@ private CompletionStage<Map<String, AvailabilityHint>> createSnapshot(RingRespon
...
+                                     "can honor only one of the 2. Clear Snapshot takes precedence over Snapshot TTL, " +
+                                     "hence snapshot {} will not be cleared with TTL", snapshotName);
```
frankgh: nit
```suggestion
                                     "hence snapshot {} will not be removed after the specified snapshot TTL option", snapshotName);
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
```diff
@@ -43,6 +43,14 @@ public final class ClientConfig
...
     public static final String CREATE_SNAPSHOT_KEY = "createSnapshot";
     public static final String CLEAR_SNAPSHOT_KEY = "clearSnapshot";
+    /**
+     * snapshotTTL a time to live option for the snapshot (available since Cassandra 4.1+).
+     * TTL value specified must contain unit along. For e.g. 2d represents a TTL for 2 days;
+     * 1h represents a TTL of 1 hour, etc. Valid units are {@code d}, {@code h}, {@code s},
+     * {@code ms}, {@code us}, {@code µs}, {@code ns}, and {@code m}.
+     */
+    public static final String USER_PROVIDED_SNAPSHOT_TTL = "snapshot_ttl";
+    public static final String DEFAULT_SNAPSHOT_TTL = "2d";
```
frankgh: this seems like a reasonable default. I would expect most jobs would complete in well under 1 day

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
```diff
@@ -91,6 +101,8 @@ private ClientConfig(Map<String, String> options)
         this.datacenter = options.get(MapUtils.lowerCaseKey(DC_KEY));
         this.createSnapshot = MapUtils.getBoolean(options, CREATE_SNAPSHOT_KEY, true);
         this.clearSnapshot = MapUtils.getBoolean(options, CLEAR_SNAPSHOT_KEY, createSnapshot);
+        this.userProvidedSnapshotTtl = MapUtils.getOrDefault(options, USER_PROVIDED_SNAPSHOT_TTL, null);
+        this.effectiveSnapshotTtl = this.userProvidedSnapshotTtl == null ? DEFAULT_SNAPSHOT_TTL : USER_PROVIDED_SNAPSHOT_TTL;
```
frankgh: this looks incorrect, did you mean?
```suggestion
        this.effectiveSnapshotTtl = this.userProvidedSnapshotTtl == null ? DEFAULT_SNAPSHOT_TTL : this.userProvidedSnapshotTtl;
```

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/data/CassandraDataLayerTests.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import com.google.common.collect.ImmutableMap;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class CassandraDataLayerTests
```
frankgh: nit
```suggestion
class CassandraDataLayerTests
```

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/data/CassandraDataLayerTests.java
```diff
@@ -0,0 +1,47 @@
...
+{
+    public static final Map<String, String> REQUIRED_CLIENT_CONFIG_OPTIONS = ImmutableMap.of(
+    "keyspace", "big-data",
+    "table", "customers",
+    "sidecar_instances", "localhost");
+
+    @Test
+    public void testDefaultSnapshotTTL()
```
frankgh: NIT:
```suggestion
   void testDefaultSnapshotTTL()
```

frankgh (commented):
Looks good, I've left a few comments in the PR

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
sarankk: Yes thanks for noticing that

frankgh (approved):
+1 Looks good, thanks for addressing the comments.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
```diff
@@ -146,6 +158,16 @@ public boolean clearSnapshot()
...
 
+    public String userProvidedSnapshotTtl()
+    {
+        return userProvidedSnapshotTtl;
+    }
+
+    public String effectiveSnapshotTtl()
+    {
+        return effectiveSnapshotTtl;
+    }
```
yifan-c, line 222: It looks like the `effectiveSnapshotTtl` is the resolved value on reading from the map. It is confusing to expose both the raw and the resolved values. 
Having `effectiveSnapshotTtl()` is suffice.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
```diff
@@ -341,24 +341,31 @@ private CompletionStage<Map<String, AvailabilityHint>> createSnapshot(RingRespon
...
-                                               if (isExhausted(throwable))
-                                               {
-                                                   LOGGER.warn("Failed to create snapshot on instance", throwable);
-                                                   return PartitionedDataLayer.AvailabilityHint.DOWN;
-                                               }
-
-                                               LOGGER.error("Unexpected error creating snapshot on instance", throwable);
-                                               return PartitionedDataLayer.AvailabilityHint.UNKNOWN;
-                                           });
+                    String resolvedSnapshotTtl = options.clearSnapshot() ? options.effectiveSnapshotTtl() : null;
```
yifan-c: The condition is reversed. The resolved snapshot ttl is set, when `!options.clearSnapshot()`.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
```diff
@@ -341,24 +341,31 @@ private CompletionStage<Map<String, AvailabilityHint>> createSnapshot(RingRespon
...
+                    if (!options.clearSnapshot() && options.userProvidedSnapshotTtl() != null)
+                    {
+                        LOGGER.warn("Snapshot TTL option was provided along with Clear Snapshot option, bulk reader" +
+                                    "can honor only one of the 2. Clear Snapshot takes precedence over Snapshot TTL, " +
+                                    "hence snapshot {} will not be removed after the specified snapshot TTL option", snapshotName);
+                    }
```
yifan-c: The condition is reversed too. Beside that, I think the warning is best emitted from `ClientConfig`, and only use the resolved value at the call-sites.

yifan-c (changes requested):

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
frankgh: I think the condition here is correct. When clear snapshot is set, it allows the snapshot TTL to be set. When clear snapshot is set to false, we don't want the TTL to clear the snapshot at a later point. I think that's how I understand it

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
frankgh: The user provided option is used for logging:
```
 if (!options.clearSnapshot() && options.userProvidedSnapshotTtl() != null)
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
yifan-c: my understanding is when clearSnapshot is true, the snapshot is cleared on job completion; otherwise it sets a TTL and let Cassandra server to delete. 
I'd wait for Saranya to clarify.

sarankk:
As per @yifan-c suggestions, to avoid confusion caused by allowing both `clearSnapshot` option and `snapshotTTL` option, we are going ahead with following setting to determine whether snapshots should be cleared. Introducing a new option `clearSnapshotStrategy`. Possible values it can take are,

`clearSnapshotStrategy: onCompletion // Only clear snapshot on the completion of the job. The onCompletion hook is evaluated at the best effort. It is possible the clear snapshot operation is not issued in some circumstances. `
`clearSnapshotStrategy: noOp // Do not clear snapshot at all.`
`clearSnapshotStrategy: onCompletionOrTTL [2d] // Clear the snapshot on completion or after the specified duration (2 days as the default), whichever comes first. Note that there is a risk that the snapshot is deleted before job completion if it takes more than the specified duration. In most cases, the job should complete before the default 2 days`
`clearSnapshotStrategy: TTL [2d] // Only clear snapshot after 2 days`
`clearSnapshot: true === clearSnapshotStrategy: onCompletionOrTTL`
`clearSnapshot: false === clearSnapshotStrategy: noOp`

`clearSnapshot` option will be deprecated. For backward compatibility, we will still understand the option and resolve it to appropriate `clearSnapshotStrategy` option.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/SharedClusterSparkIntegrationTestBase.java
```diff
@@ -84,7 +84,7 @@ protected DataFrameReader bulkReaderDataFrame(QualifiedName tableName)
                   // Shutdown hooks are called after the job ends, and in the case of integration tests
                   // the sidecar is already shut down before this. Since the cluster will be torn
                   // down anyway, the integration job skips clearing snapshots.
-                  .option("clearSnapshot", "false")
+                  .option("clearSnapshotStrategy", "OnCompletion")
```
yifan-c: before the patch, the test does not clear snapshot. Can you comment why it is changed?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
```diff
@@ -138,7 +138,7 @@ private ClearSnapshotStrategy parseClearSnapshotStrategy(boolean hasDeprecatedOp
                                                              boolean clearSnapshot,
                                                              String clearSnapshotStrategyOption)
     {
-        if (hasDeprecatedOption)
+        if (hasDeprecatedOption && clearSnapshotStrategyOption == null)
```
yifan-c: If the deprecated option is set, the warning should always be logged, regardless whether the new option is present or not. 

```java
if (hasDeprecatedOption)
{
  log warning
  if (clearSnapshotStrategyOption == null)
  {
    return the default based on the value
  }
}    
```

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/SharedClusterSparkIntegrationTestBase.java
sarankk: since `onCompletion` behaviour was similar to test setup (clear after test method completes), kept that. But for consistency, will change it to `noop`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
sarankk: Makes sense, will update it

yifan-c (approved):

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
```diff
@@ -109,6 +134,50 @@ private ClientConfig(Map<String, String> options)
         this.quoteIdentifiers = MapUtils.getBoolean(options, QUOTE_IDENTIFIERS, false);
     }
 
+    private ClearSnapshotStrategy parseClearSnapshotStrategy(boolean hasDeprecatedOption,
```
frankgh: maybe make it protected for extensibility?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
```diff
@@ -109,6 +134,50 @@ private ClientConfig(Map<String, String> options)
...
+                String msg = "Incorrect value set for clearSnapshotStrategy, expected format is " +
+                             "{strategy [snapshotTTLvalue]}. TTL value specified must contain unit along. " +
+                             "For e.g. 2d represents a TTL for 2 days";
+                throw new IllegalArgumentException(msg);
+            }
+        }
+        else
+        {
+            LOGGER.error("Invalid value for ClearSnapshotStrategy: '{}'", clearSnapshotStrategyOption);
+            throw new IllegalArgumentException("Invalid value: " + clearSnapshotStrategyOption);
```
frankgh, line 177: In ClearSnapshotStrategy.create the fallback behavior is to return the default strategy. Should we consider that fallback strategy here as well?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
```diff
@@ -24,25 +24,45 @@
...
+     * clear snapshot strategies allowed check {@link ClearSnapshotStrategy}
+     */
+    public static final String CLEAR_SNAPSHOT_STRATEGY_KEY = "clearSnapshotStrategy";
+    /**
+     * TTL value is time to live option for the snapshot (available since Cassandra 4.1+). TTL value specified must
+     * contain unit along. For e.g. 2d represents a TTL for 2 days; 1h represents a TTL of 1 hour, etc.
+     * Valid units are {@code d}, {@code h}, {@code s}, {@code ms}, {@code us}, {@code µs}, {@code ns}, and {@code m}.
+     */
+    public static final String DEFAULT_SNAPSHOT_TTL_VALUE = "2d";
+    public static final String SNAPSHOT_TTL_PATTERN = "\\d+(d|h|m|s|ms)|(\\d+|\\d+\\.\\d+|\\.\\d+)[eE][+-](\\d+|\\d+\\.\\d+|\\.\\d+)(us|µs|ns)";
```
frankgh: TTL support is limited to days , hours, minutes and seconds, and it doesn't support e notation. This is the pattern used in Cassandra

```
private static final Pattern UNITS_PATTERN = Pattern.compile(("^(\\d+)(d|h|s|ms|us|µs|ns|m)$"));
```

we'll need to limit it to `d|h|m|s`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
```diff
@@ -237,4 +311,125 @@ private List<SchemaFeature> initRequestedFeatures(Map<String, String> options)
...
+            else
+            {
+                ClearSnapshotStrategy defaultStrategy = defaultStrategy();
+                LOGGER.warn("Unknown ClearSnapshotStrategy {} is passed. Fall back to default strategy {}.",
+                            name, defaultStrategy);
+                return defaultStrategy;
+            }
+        }
+
+        static ClearSnapshotStrategy defaultStrategy()
```
frankgh: can we make it public?

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/data/ClientConfigTests.java
```diff
@@ -0,0 +1,59 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class ClientConfigTests
+{
+    @Test
+    void testPositiveSnapshotTTLPatterns()
+    {
+        assertTrue("2h".matches(SNAPSHOT_TTL_PATTERN));
+        assertTrue("200s".matches(SNAPSHOT_TTL_PATTERN));
+        assertTrue("20000ms".matches(SNAPSHOT_TTL_PATTERN));
```
frankgh: ms is not supported by the backend

frankgh (changes requested):
Can we also add tests to cover parsing the clear snapshot strategy option?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
sarankk: Thanks updated pattern accordingly

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
sarankk: Should we fail fast there too?. If users are setting `clearSnapshotStrategy` then they might want a particular behaviour, if that behaviour can't be met, better to alert earlier.

cassandra-bridge/src/main/java/org/apache/cassandra/spark/utils/MapUtils.java
```diff
@@ -225,4 +225,15 @@ public static String getOrDefault(Map<String, String> options, String key, Strin
...
         return options.getOrDefault(lowerCaseKey(key), defaultValue);
     }
+
+    /**
+     * Method to check if key is present in {@code options} map.
+     * @param options   the map
+     * @param key       the key to the map
+     * @return boolean
+     */
+    public static boolean contains(Map<String, String> options, String key)
```
frankgh: should this method be called `containsKey` instead?

sarankk: makes sense, will make it consistent with `Map`

frankgh (approved):

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
yifan-c: Fail fast sounds good.

frankgh:
Closed via https://github.com/apache/cassandra-analytics/commit/8c20b452dd0728a6fad6d276a7be9fa1b9274495

PR #32, opened 2024-01-16
[CASSANDRA-19272] Adds integration testing for blocklist handling

arjunashok:
### Changes
- Adds spark config with default empty list for `blocked_instances`
- Adds tests for blocked instances

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
@@ -50,5 +50,10 @@ public enum WriterOptions implements WriterOption
      * support mixed case and reserved keyword names for these fields.
      */
     QUOTE_IDENTIFIERS,
+    /**
+     * Option to specify a comma separated list of Cassandra instances that are blocked from receiving any
+     * writes from the analytics library. Note that this contributes to the number of unavailable instances for
+     * the write consistency level validations.
+     */
     BLOCKED_CASSANDRA_INSTANCES,
 }
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
@@ -50,5 +50,5 @@ public enum WriterOptions implements WriterOption
      * support mixed case and reserved keyword names for these fields.
      */
     QUOTE_IDENTIFIERS,
-    BLOCKED_INSTANCES,
+    BLOCKED_CASSANDRA_INSTANCES,
 }
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
@@ -50,4 +50,5 @@ public enum WriterOptions implements WriterOption
      * support mixed case and reserved keyword names for these fields.
      */
     QUOTE_IDENTIFIERS,
+    BLOCKED_INSTANCES,
 }
--- a/cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BlockedInstancesTest.java
+++ b/cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BlockedInstancesTest.java
@@ -1,0 +1,167 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.analytics;
+
+import java.io.IOException;
+import java.lang.reflect.Method;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+
+import org.apache.commons.lang3.StringUtils;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
+
+import org.apache.cassandra.distributed.UpgradeableCluster;
+import org.apache.cassandra.distributed.api.Feature;
+import org.apache.cassandra.distributed.api.IInstance;
+import org.apache.cassandra.sidecar.testing.QualifiedName;
+import org.apache.cassandra.spark.bulkwriter.WriterOptions;
+import org.apache.cassandra.testing.TestVersion;
+import org.apache.spark.sql.Dataset;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.SparkSession;
+
+import static com.datastax.driver.core.ConsistencyLevel.ALL;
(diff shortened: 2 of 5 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
```diff
@@ -50,4 +50,5 @@ public enum WriterOptions implements WriterOption
      * support mixed case and reserved keyword names for these fields.
      */
     QUOTE_IDENTIFIERS,
+    BLOCKED_INSTANCES,
```
frankgh: how about calling them `BLOCKED_SIDECAR_INSTANCES` to be explicit about which instances are blocked. Also, I think we should add javadocs to all these properties (in a different jira); but in the context of this PR, we should add javadocs to the property that is being added.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BlockedInstancesTest.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Override
+    protected void initializeSchemaForTest()
+    {
+        createTestKeyspace(TEST_KEYSPACE, DC1_RF3);
+
+        try
+        {
+            Class<?> testClass = Class.forName("org.apache.cassandra.analytics.BlockedInstancesTest");
```
frankgh: how about doing `this.getClass()` instead in case the name of the file changes or package moves?

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BlockedInstancesTest.java
```diff
@@ -0,0 +1,167 @@
...
+            Method[] methods = testClass.getDeclaredMethods();
+
+            for (Method method : methods)
+            {
+                if (method.isAnnotationPresent(Test.class))
+                {
+                    createTestTable(new QualifiedName(TEST_KEYSPACE, method.getName().toLowerCase()), CREATE_TEST_TABLE_STATEMENT);
```
frankgh: there are some limitations on valid keyspace/table names in Cassandra, including length, so we need to be a bit careful there. but this looks like a clever approach to setup schemas

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BlockedInstancesTest.java
```diff
@@ -0,0 +1,167 @@
...
+        df = DataGenerationUtils.generateCourseData(spark, ROW_COUNT);
+        // Generate the expected data for the new instances
+        expectedInstanceData = generateExpectedInstanceData(cluster, Collections.emptyList(), ROW_COUNT);
+    }
+
+    @Override
+    protected UpgradeableCluster provisionCluster(TestVersion testVersion) throws IOException
+    {
+        ClusterBuilderConfiguration configuration = testClusterConfiguration();
+        return clusterBuilder(configuration, testVersion);
```
frankgh: minor NIT: inline? 
```suggestion
        return clusterBuilder(testClusterConfiguration(), testVersion);
```

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BlockedInstancesTest.java
```diff
@@ -0,0 +1,167 @@
...
+    Dataset<Row> df;
+    Map<IInstance, Set<String>> expectedInstanceData;
+
+    @Test
+    void testSingleBlockedNodeSucceeds(TestInfo testInfo)
+    {
+        TestConsistencyLevel cl = TestConsistencyLevel.of(QUORUM, QUORUM);
+        QualifiedName table = new QualifiedName(TEST_KEYSPACE, testInfo.getTestMethod().get().getName().toLowerCase());
+        bulkWriterDataFrameWriter(df, table).option(WriterOptions.BULK_WRITER_CL.name(), cl.writeCL.name())
+                                            .option("blocked_instances", "127.0.0.2")
```
frankgh: should we use writer options here as well for consistency with the previous option?
```suggestion
                                            .option(WriterOptions.BLOCKED_INSTANCES.name(), "127.0.0.2")
```

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BlockedInstancesTest.java
```diff
@@ -0,0 +1,167 @@
...
+                                            .save();
+        expectedInstanceData.entrySet()
+                            .stream()
+                            .filter(e -> e.getKey().broadcastAddress().getAddress().getHostAddress().equals("127.0.0.2"))
+                            .forEach(e -> e.setValue(Collections.emptySet()));
+        validateNodeSpecificData(table, expectedInstanceData);
+        validateData(table, cl.readCL, ROW_COUNT);
+
```
frankgh: nit empty line

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BlockedInstancesTest.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Test
+    void testSingleBlockedNodeWithWriteFailure(TestInfo testInfo)
+    {
+        TestConsistencyLevel cl = TestConsistencyLevel.of(ONE, ALL);
+        QualifiedName table = new QualifiedName(TEST_KEYSPACE, testInfo.getTestMethod().get().getName().toLowerCase());
+        Throwable thrown = catchThrowable(() ->
+                                          bulkWriterDataFrameWriter(df, table).option(WriterOptions.BULK_WRITER_CL.name(), cl.writeCL.name())
+                                                                              .option("blocked_instances", "127.0.0.2")
```
frankgh: use the writer option instead for consistency?

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BlockedInstancesTest.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    @Test
+    void testMultipleBlockedNodesWithWriteFailure(TestInfo testInfo)
+    {
+        TestConsistencyLevel cl = TestConsistencyLevel.of(QUORUM, QUORUM);
+        QualifiedName table = new QualifiedName(TEST_KEYSPACE, testInfo.getTestMethod().get().getName().toLowerCase());
+        Throwable thrown = catchThrowable(() ->
+                                          bulkWriterDataFrameWriter(df, table).option(WriterOptions.BULK_WRITER_CL.name(), cl.writeCL.name())
+                                                                              .option("blocked_instances", "127.0.0.2,127.0.0.3")
```
frankgh: use writer option instead for consistency?

frankgh (commented):
Looks good, I've left a few comments in the PR

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
frankgh: on second thought these are actually Cassandra instances, correct? I think we should maybe explicitly call them `BLOCKED_CASSANDRA_INSTANCES` maybe? I don't think Sidecar itself is blocked, but rather specific Cassandra instances, right?

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BlockedInstancesTest.java
arjunashok: good point. Added a class level javadoc describing what we're doing for now.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
arjunashok: Sure, doesn't hurt to make it more explicit.

frankgh (approved):
+1 Looks good to me

frankgh:
Closed via https://github.com/apache/cassandra-analytics/commit/fa6df8e2c09ad3d27bfe8c0ce016c839094630f6

PR #36, opened 2024-01-27
CASSANDRA-19334 Upgrade to Cassandra 4.0.12 and remove BufferMode and…

yifan-c:
… BatchSize options

In cassandra-all:4.0.12, improvements were made for the CQLSSTableWriter. The sorted writer now can produce size-capped SSTables. It replaces the need for the unsorted sstable writer, which has to buffer and sort data on flushing. The dataset to write in the spark application is already sorted. By avoiding using the unsorted writer, it prevents wasting CPU time on sorting the sorted data. Since the sorted sstable writer does not need to buffer data, its size estimation is more accurate than the unsorted one, meaning the produced sstables files are closer to the expectation.

By removing the unsorted sstable writer, it no longer requires the RowBufferMode option. By supporting size-capping in sorted writer, it no longer requires the BatchSize option.


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -147,9 +143,7 @@ public BulkSparkConf(SparkConf conf, Map<String, String> options)
         this.consistencyLevel = ConsistencyLevel.CL.valueOf(MapUtils.getOrDefault(options, WriterOptions.BULK_WRITER_CL.name(), "EACH_QUORUM"));
         this.localDC = MapUtils.getOrDefault(options, WriterOptions.LOCAL_DC.name(), null);
         this.numberSplits = MapUtils.getInt(options, WriterOptions.NUMBER_SPLITS.name(), DEFAULT_NUM_SPLITS, "number of splits");
-        this.rowBufferMode = MapUtils.getEnumOption(options, WriterOptions.ROW_BUFFER_MODE.name(), DEFAULT_ROW_BUFFER_MODE, "row buffering mode");
-        this.sstableDataSizeInMB = MapUtils.getInt(options, WriterOptions.SSTABLE_DATA_SIZE_IN_MB.name(), 160, "sstable data size in MB");
-        this.sstableBatchSize = MapUtils.getInt(options, WriterOptions.BATCH_SIZE.name(), 1_000_000, "sstable batch size");
+        this.sstableDataSizeInMB = MapUtils.getInt(options, WriterOptions.SSTABLE_DATA_SIZE_IN_MB.name(), DEFAULT_SSTABLE_DATA_SIZE_IN_MB, "sstable data size in mebibytes");
```
frankgh: if data size is mebibytes, we should probably rename the variables to MIB.

```
WriterOptions.SSTABLE_DATA_SIZE_IN_MB -> WriterOptions.SSTABLE_DATA_SIZE_IN_MIB
DEFAULT_SSTABLE_DATA_SIZE_IN_MB -> MIB
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/JobInfo.java
```diff
@@ -33,13 +32,11 @@ public interface JobInfo extends Serializable
 
     String getLocalDC();
 
-    @NotNull
-    RowBufferMode getRowBufferMode();
-
+    /**
+     * @return the max sstable data file size in mebibytes
+     */
     int getSstableDataSizeInMB();
```
frankgh:
```suggestion
    int getSstableDataSizeInMiB();
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -368,16 +344,23 @@ private static Map<String, Object> getBindValuesForColumns(Map<String, Object> m
         return map;
     }
 
+    /**
+     * Close the {@link RecordWriter#sstableWriter} is present. Schedule a stream session with the produced sstables.
```
frankgh:
```suggestion
     * Close the {@link RecordWriter#sstableWriter} if present. Schedule a stream session with the produced sstables.
```

cassandra-four-zero/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
```diff
@@ -164,13 +166,31 @@ public static synchronized void setup()
...
             DatabaseDescriptor.setEndpointSnitch(new SimpleSnitch());
             Keyspace.setInitialized();
 
             setup = true;
         }
     }
 
+    public static void setupCommitLogConfigs(Path path)
+    {
+        String commitLogPath = path + "/commitlog";
```
frankgh, line 179: can commit logs be disabled for offline tools instead?

scripts/build-dtest-jars.sh
```diff
@@ -38,8 +38,8 @@ else
   #   "cassandra-4.0:cassandra-4.0"
   # Due to MacOS being stuck on Bash < 4, we don't use associative arrays here.
   CANDIDATE_BRANCHES=(
-    "cassandra-4.0:1f79c8492528f01bcc5f88951a1cc9e0d7265c54"
-    "cassandra-4.1:725655dda2776fef35567496a6e331102eb7610d"
+    "cassandra-4.0:af752fcd535ccdac69b9fed88047b2dd7625801e"
```
frankgh: it feels like we should use the tag here instead: `cassandra-4.0.12`
```suggestion
    "cassandra-4.0:cassandra-4.0.12"
```

frankgh (commented):
Nice improvement! I've left some minor comments

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
yifan-c: I am with you. The reason that I only changed the description is because the option has been named as "MB" but used as "MiB" since it is introduced. 
How about I add a new option, mark this one `WriterOptions.SSTABLE_DATA_SIZE_IN_MB` as deprecated (with the new description as the correction), and both are treated the same, as they are both in MiB effectively.

cassandra-four-zero/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
yifan-c: It should not be necessary. But the current library requires the config to be present. The code path that checks/loads the config is when getting the local host uuid. Seemingly related with newly added CMS. Setting the config is not a big deal, so it is not burning issue unless I missed something. 
1. `org.apache.cassandra.io.sstable.SSTableSimpleWriter#getOrCreateWriter`
2. `org.apache.cassandra.io.sstable.AbstractSSTableSimpleWriter#createWriter`
3. `org.apache.cassandra.io.sstable.SimpleSSTableMultiWriter#create`
4. `org.apache.cassandra.io.sstable.metadata.MetadataCollector#MetadataCollector(org.apache.cassandra.db.ClusteringComparator)`
5. `org.apache.cassandra.service.StorageService#getLocalHostUUID`

scripts/build-dtest-jars.sh
yifan-c: Good catch. Forgot that tags are supported.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
frankgh: +1 that makes sense

frankgh (approved):
+1 looks good

PR #39, opened 2024-02-13
CASSANDRA-19285 Fix flaky Host replacement tests and shrink tests

yifan-c:
The flakiness is caused by inspecting a class whose classloader is already closed. The fix is to include the those classes in the sharedClassLoader, so that the classLoader is not closed during the test.


frankgh (approved):
+1 Thanks for addressing this issue!

PR #40, opened 2024-02-20
CASSANDRA-19411: Bulk reader fails to produce a row when regular column values are null

frankgh:
Bulk Reader won't emit a row when the regular column values are all `null`. For example, a schema `PK` = `a`, `b` ; `CK` = `c`, `d` ; and columns = `e`, `f`.

|  a  |  b  |  c  |  d  |  e   |  f   |
| --- | --- | --- | --- | ---- | ---- |
| pk1 | pk2 | ck1 | ck2 | null | null |

When queried from Analytics bulk reader, it won't produce a row.

This issue also occurs when the projected regular column values are all `null`, where other non-projected columns might have some values.

Patch by Francisco Guerrero; Reviewed by TBD for CASSANDRA-19411

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 1.0.0
+ * Bulk reader fails to produce a row when regular column values are null (CASSANDRA-19411)
  * Use XXHash32 for digest calculation of SSTables (CASSANDRA-19369)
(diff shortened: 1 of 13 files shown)


frankgh:
Closed via https://github.com/apache/cassandra-analytics/commit/46c35d0ef2efb66512133a7913df9936b0a80dc8

PR #41, opened 2024-02-21
CASSANDRA-19418  - Changes to report additional bulk analytics job stats for instrumentation

arjunashok:
## Changes

- Update to relay the results of the spark tasks to the executor to instrument the total no. of rows written.
- Abstracts the current `dialHome` implementation within `CassandraBulkWriterContext` to be pluggable via a `JobStats` interface, that can allow recording of job parameters from the spark driver.
- To allow reporting of stats such as no. rows, bytes written and resize detection, a `WriterResult` type is returned from each executor, and are reported by the driver after aggregation.
- The above abstraction also separates the recording of the metadata as the job progresses and separately publish them at the end of the job with success/failure status and failure reason where applicable.

## Testing
- Validated successful build run w/ unit/integration tests.

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriterContext.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriterContext.java
@@ -21,7 +21,7 @@ package org.apache.cassandra.spark.bulkwriter;
 
 import java.io.Serializable;
 
-import org.apache.cassandra.spark.common.stats.JobStats;
+import org.apache.cassandra.spark.common.stats.JobStatsPublisher;
 
 public interface BulkWriterContext extends Serializable
 {
@@ -29,7 +29,7 @@ public interface BulkWriterContext extends Serializable
 
     JobInfo job();
 
-    JobStats jobStats();
+    JobStatsPublisher jobStats();
 
     SchemaInfo schema();
 
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
@@ -116,7 +116,7 @@ public class CassandraBulkSourceRelation extends BaseRelation implements Inserta
         try
         {
             List<WriteResult> writeResults = sortedRDD
-                                 .mapPartitions(partitionsFlatMapFunc(broadcastContext, columnNames))
+                                 .mapPartitions(writeRowsInPartition(broadcastContext, columnNames))
                                  .collect();
 
             recordSuccessfulJobStats(writeResults);
@@ -131,7 +131,7 @@ public class CassandraBulkSourceRelation extends BaseRelation implements Inserta
         {
             try
             {
-                writerContext.publishJobStats();
+                writerContext.jobStats().publishJobStats();
                 writerContext.shutdown();
                 sqlContext().sparkContext().clearJobGroup();
             }
@@ -158,7 +158,7 @@ public class CassandraBulkSourceRelation extends BaseRelation implements Inserta
                     rowCount,
                     totalBytesWritten,
                     hasClusterTopologyChanged);
-        writerContext.recordJobStats(new HashMap<>()
+        writerContext.jobStats().recordJobStats(new HashMap<>()
         {
             {
                 put("rowsWritten", Long.toString(rowCount));
@@ -172,7 +172,7 @@ public class CassandraBulkSourceRelation extends BaseRelation implements Inserta
 
     private void recordFailureStats(String reason)
     {
-        writerContext.recordJobStats(new HashMap<>()
+        writerContext.jobStats().recordJobStats(new HashMap<>()
         {
             {
                 put("jobStatus", "Failed");
@@ -195,7 +195,7 @@ public class CassandraBulkSourceRelation extends BaseRelation implements Inserta
      * @return FlatMapFunction
      */
     private static FlatMapFunction<Iterator<Tuple2<DecoratedKey, Object[]>>, WriteResult>
-    partitionsFlatMapFunc(Broadcast<BulkWriterContext> ctx, String[] columnNames)
+    writeRowsInPartition(Broadcast<BulkWriterContext> ctx, String[] columnNames)
     {
         return iterator -> Collections.singleton(new RecordWriter(ctx.getValue(), columnNames).write(iterator)).iterator();
     }
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
@@ -19,10 +19,13 @@
 
 package org.apache.cassandra.spark.bulkwriter;
 
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -112,13 +115,11 @@ public class CassandraBulkSourceRelation extends BaseRelation implements Inserta
     {
         try
         {
-            List<StreamResult> results = sortedRDD
-                                         .mapPartitions(partitionsFlatMapFunc(broadcastContext, columnNames))
-                                         .collect();
-            long rowCount = results.stream().mapToLong(res -> res.rowCount).sum();
-            long totalBytesWritten = results.stream().mapToLong(res -> res.bytesWritten).sum();
-            LOGGER.info("Bulk writer has written {} rows and {} bytes", rowCount, totalBytesWritten);
-            recordSuccessfulJobStats(rowCount, totalBytesWritten);
+            List<WriteResult> writeResults = sortedRDD
+                                 .mapPartitions(partitionsFlatMapFunc(broadcastContext, columnNames))
+                                 .collect();
+
+            recordSuccessfulJobStats(writeResults);
         }
         catch (Throwable throwable)
         {
@@ -142,14 +143,29 @@ public class CassandraBulkSourceRelation extends BaseRelation implements Inserta
         }
     }
 
-    private void recordSuccessfulJobStats(long rowCount, long totalBytesWritten)
+    private void recordSuccessfulJobStats(List<WriteResult> writeResults)
     {
+        List<StreamResult> streamResults = writeResults.stream()
+                                                       .map(WriteResult::streamResults)
+                                                       .flatMap(Collection::stream)
+                                                       .collect(Collectors.toList());
+
+        long rowCount = streamResults.stream().mapToLong(res -> res.rowCount).sum();
+        long totalBytesWritten = streamResults.stream().mapToLong(res -> res.bytesWritten).sum();
+        boolean hasClusterTopologyChanged = writeResults.stream()
+                                                        .map(WriteResult::isClusterResizeDetected)
+                                                        .anyMatch(b -> b);
+        LOGGER.info("Bulk writer has written {} rows and {} bytes with cluster-resize status: {}",
+                    rowCount,
+                    totalBytesWritten,
+                    hasClusterTopologyChanged);
         writerContext.recordJobStats(new HashMap<>()
         {
             {
                 put("rowsWritten", Long.toString(rowCount));
                 put("bytesWritten", Long.toString(totalBytesWritten));
                 put("jobStatus", "Succeeded");
+                put("clusterResizeDetected", String.valueOf(hasClusterTopologyChanged));
                 put("jobElapsedTimeMillis", Long.toString(getElapsedTimeMillis()));
             }
         });
@@ -179,10 +195,10 @@ public class CassandraBulkSourceRelation extends BaseRelation implements Inserta
      * @param ctx BulkWriterContext broadcast variable
      * @return FlatMapFunction
      */
-    private static FlatMapFunction<Iterator<Tuple2<DecoratedKey, Object[]>>, StreamResult>
+    private static FlatMapFunction<Iterator<Tuple2<DecoratedKey, Object[]>>, WriteResult>
     partitionsFlatMapFunc(Broadcast<BulkWriterContext> ctx, String[] columnNames)
     {
-        return iterator -> new RecordWriter(ctx.getValue(), columnNames).write(iterator).iterator();
+        return iterator -> Collections.singleton(new RecordWriter(ctx.getValue(), columnNames).write(iterator)).iterator();
     }
 
     /**
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
@@ -153,9 +153,8 @@ public class CassandraBulkSourceRelation extends BaseRelation implements Inserta
         long rowCount = streamResults.stream().mapToLong(res -> res.rowCount).sum();
         long totalBytesWritten = streamResults.stream().mapToLong(res -> res.bytesWritten).sum();
         boolean hasClusterTopologyChanged = writeResults.stream()
-                                                        .map(WriteResult::isClusterResizeDetected)
-                                                        .anyMatch(b -> b);
-        LOGGER.info("Bulk writer has written {} rows and {} bytes with cluster-resize status: {}",
+                                                        .anyMatch(WriteResult::isClusterResizeDetected);
+        LOGGER.info("Bulk writer job complete. rows={} bytes={} cluster_resize={}",
                     rowCount,
                     totalBytesWritten,
                     hasClusterTopologyChanged);
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
@@ -116,8 +116,9 @@ public class CassandraBulkSourceRelation extends BaseRelation implements Inserta
                                          .mapPartitions(partitionsFlatMapFunc(broadcastContext, columnNames))
                                          .collect();
             long rowCount = results.stream().mapToLong(res -> res.rowCount).sum();
-            LOGGER.info("Bulk writer has written {} rows", rowCount);
-            recordSuccessfulJobStats(rowCount);
+            long totalBytesWritten = results.stream().mapToLong(res -> res.bytesWritten).sum();
+            LOGGER.info("Bulk writer has written {} rows and {} bytes", rowCount, totalBytesWritten);
+            recordSuccessfulJobStats(rowCount, totalBytesWritten);
         }
         catch (Throwable throwable)
         {
@@ -141,12 +142,13 @@ public class CassandraBulkSourceRelation extends BaseRelation implements Inserta
         }
     }
 
-    private void recordSuccessfulJobStats(long rowCount)
+    private void recordSuccessfulJobStats(long rowCount, long totalBytesWritten)
     {
         writerContext.recordJobStats(new HashMap<>()
         {
             {
                 put("rowsWritten", Long.toString(rowCount));
+                put("bytesWritten", Long.toString(totalBytesWritten));
                 put("jobStatus", "Succeeded");
                 put("jobElapsedTimeMillis", Long.toString(getElapsedTimeMillis()));
             }
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
@@ -119,11 +119,11 @@ public class CassandraBulkSourceRelation extends BaseRelation implements Inserta
                                  .mapPartitions(writeRowsInPartition(broadcastContext, columnNames))
                                  .collect();
 
-            recordSuccessfulJobStats(writeResults);
+            publishSuccessfulJobStats(writeResults);
         }
         catch (Throwable throwable)
         {
-            recordFailureStats(throwable.getMessage());
+            publishFailureJobStats(throwable.getMessage());
             LOGGER.error("Bulk Write Failed", throwable);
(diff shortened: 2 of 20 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriterContext.java
```diff
@@ -21,7 +21,9 @@
 
 import java.io.Serializable;
 
-public interface BulkWriterContext extends Serializable
+import org.apache.cassandra.spark.common.Reportable;
+
+public interface BulkWriterContext extends Serializable, Reportable
```
frankgh: I don't think this belongs in the `BulkWriterContext`. The context is something we share with executors. However the job stats are tracked from the driver only. I would avoid adding anything else to the BulkWriterContext

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/common/Reportable.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cassandra.spark.common;
```
frankgh: should this be in a stats or metrics package instead?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/common/Reportable.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.Map;
+
+/**
+ * Interface to provide functionality to report Spark Job Statistics and/or properties
+ * that can optionally be instrumented. The default implementation merely logs these
+ * stats at the end of the job.
+ */
+public interface Reportable
```
frankgh: How about naming this class `BulkWriterJobStats` ?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RingInstance.java
```diff
@@ -49,6 +49,7 @@ public RingInstance(ReplicaMetadata replica)
                          .datacenter(replica.datacenter())
                          .state(replica.state())
                          .status(replica.status())
+                         .token("")
```
frankgh: why this change?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RingInstance.java
```diff
@@ -125,40 +126,28 @@ private void writeObject(ObjectOutputStream out) throws IOException
         out.writeUTF(ringEntry.address());
         out.writeInt(ringEntry.port());
         out.writeUTF(ringEntry.datacenter());
-        out.writeUTF(ringEntry.load());
```
frankgh, line 128: why do we remove these?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -107,17 +112,25 @@ private void persist(@NotNull JavaPairRDD<DecoratedKey, Object[]> sortedRDD, Str
...
         {
+            recordFailureStats(throwable.getMessage());
             LOGGER.error("Bulk Write Failed", throwable);
             throw new RuntimeException("Bulk Write to Cassandra has failed", throwable);
         }
         finally
         {
             try
             {
+                writerContext.publishJobStats();
```
frankgh: I don't think this will do what you think it will do. This object _should_ be treated as a read-only object from the point of view of the executors. Data won't propagate back to the driver once the executors complete

frankgh (changes requested):
I think there are some issues with the PR in the sense that the data "reported" from executors will never make it back to the driver. Maybe revisit that aspect and see if you are trying to do will work

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/common/Reportable.java
arjunashok: Makes sense

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/common/Reportable.java
arjunashok: This is not meant to be specific to the writer, so can be renamed to `JobStats` maybe?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RingInstance.java
arjunashok: Since We are now returning the `StreamResult` back from the tasks, the existing implementation will result in NPEs while serializing the contained `RingInstance`, due to many of these fields not being defined when we create the `RingInstance` from `ReplicaMetadata`.  The change removes the fields not being used from RingInstance from the serialization context.

Likewise, the change is also explicitly setting the `token` field to a default for the same reason, since `token` is part of the equals/hashcode validations.

Stack trace:

```
Caused by: org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 1.0 failed 4 times, most recent failure: Lost task 0.3 in stage 1.0 (TID 40) (172.20.10.7 executor driver): com.esotericsoftware.kryo.KryoException: Error during Java serialization.
Serialization trace:
instance (org.apache.cassandra.spark.bulkwriter.CommitResult)
commitResults (org.apache.cassandra.spark.bulkwriter.StreamResult)
	at org.apache.cassandra.spark.bulkwriter.util.SbwJavaSerializer.write(SbwJavaSerializer.java:58)
	at com.esotericsoftware.kryo.Kryo.writeObject(Kryo.java:575)
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RingInstance.java
arjunashok: Answered in the response below

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriterContext.java
arjunashok: So, the functionality provided by the new interface is replacing the existing `dialHome` method in the `CassandraBulkWriterContext` .  

The thinking is that this is tied to the "context" that is shared across executors, as we "record" initial stats and job status stats at the executor level and the "inflight" stats at the task level.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
arjunashok: So, the change is not propagating data back to the driver, but publishes stats from the executors. I am assuming here that the context is made available to the executors so what you are saying does not need to happen. Let me know if that makes sense.

I have been able to validate this using the in-jvm-dtests.

arjunashok: You're right. Fixed to only report/publish stats from the driver.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriterContext.java
arjunashok: We could potentially have this implemented by the concrete `CassandraBulkWriterContext` to keep it consistent with how `dialHome` is currently implemented.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/common/JobStats.java
```diff
@@ -0,0 +1,41 @@
...
+package org.apache.cassandra.spark.common;
+
+import java.util.Map;
+
+/**
+ * Interface to provide functionality to report Spark Job Statistics and/or properties
+ * that can optionally be instrumented. The default implementation merely logs these
+ * stats at the end of the job.
+ */
+public interface JobStats
```
frankgh: Since `JobStats` is not part of the context, we need to make sure we define this class as serializable, since the `BulkWriterContext` is a `Broadcast` variable

```suggestion
public interface JobStats extends Serializable
```

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/MockBulkWriterContext.java
```diff
@@ -77,6 +78,23 @@ public class MockBulkWriterContext implements BulkWriterContext, ClusterInfo, Jo
     private int sstableDataSizeInMB = 128;
     private int sstableWriteBatchSize = 2;
 
+    private final Map<String, String> jobStats = new HashMap<>();
+
+    public Map<String, String> jobStats()
+    {
+        return jobStats;
+    }
```
frankgh: NIT: missing line break
```suggestion
    }

```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriteResult.java
```diff
@@ -0,0 +1,54 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cassandra.spark.bulkwriter;
+
+import java.util.List;
+
+import scala.Serializable;
```
frankgh: shouldn't this be a `java.io.Serializable` instead?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriteResult.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class WriteResult implements Serializable
```
frankgh, line 29: let's add javadocs here

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriteResult.java
```diff
@@ -0,0 +1,54 @@
...
+    public final List<StreamResult> streamResults;
+    public final boolean clusterResizeDetected;
+
+    public WriteResult(List<StreamResult> streamResults, boolean clusterResizeDetected)
+    {
+        this.streamResults = streamResults;
+        this.clusterResizeDetected = clusterResizeDetected;
+    }
+
+    public List<StreamResult> streamResults()
```
frankgh, line 46: javadocs

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SSTableWriter.java
```diff
@@ -59,6 +59,9 @@ public class SSTableWriter
     private final Map<Path, Digest> fileDigestMap = new HashMap<>();
     private final DigestAlgorithm digestAlgorithm;
 
+    private long rowCount = 0;
+
```
frankgh, line 63: NIT: remove extra line break
```suggestion
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SSTableWriter.java
```diff
@@ -100,6 +103,15 @@ public void addRow(BigInteger token, Map<String, Object> boundValues) throws IOE
         }
         maxToken = token;
         cqlSSTableWriter.addRow(boundValues);
+        rowCount += 1;
+    }
+
+    /**
+     * @return the total number of rows added
```
frankgh:
```suggestion
     * @return the total number of rows written
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -75,6 +75,8 @@ public class RecordWriter implements Serializable
     private SSTableWriter sstableWriter = null;
     private int outputSequence = 0; // sub-folder for possible subrange splits
 
+    private static final String STATS_KEY_RESIZE_DETECTED = "clusterResizeDetected";
```
frankgh: do we need this? it seem unused
```suggestion
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
```diff
@@ -56,10 +57,12 @@ public class CassandraBulkWriterContext implements BulkWriterContext, KryoSerial
     private final CassandraClusterInfo clusterInfo;
     private final SchemaInfo schemaInfo;
 
-    private CassandraBulkWriterContext(@NotNull BulkSparkConf conf,
-                                       @NotNull CassandraClusterInfo clusterInfo,
-                                       @NotNull StructType dfSchema,
-                                       SparkContext sparkContext)
+    private final Map<String, String> jobStats = new HashMap<>();
```
frankgh: should this be transient? I don't think we want to serialize this value to the executors?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RingInstance.java
frankgh: shouldn't we handle nulls during serialization/deserialization instead?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -129,6 +143,64 @@ private void persist(@NotNull JavaPairRDD<DecoratedKey, Object[]> sortedRDD, Str
...
+        List<StreamResult> streamResults = writeResults.stream()
+                                                       .map(WriteResult::streamResults)
+                                                       .flatMap(Collection::stream)
+                                                       .collect(Collectors.toList());
+
+        long rowCount = streamResults.stream().mapToLong(res -> res.rowCount).sum();
+        long totalBytesWritten = streamResults.stream().mapToLong(res -> res.bytesWritten).sum();
+        boolean hasClusterTopologyChanged = writeResults.stream()
+                                                        .map(WriteResult::isClusterResizeDetected)
+                                                        .anyMatch(b -> b);
```
frankgh: NIT:
```suggestion
                                                        .anyMatch(WriteResult::isClusterResizeDetected);
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -129,6 +143,64 @@ private void persist(@NotNull JavaPairRDD<DecoratedKey, Object[]> sortedRDD, Str
...
+        LOGGER.info("Bulk writer has written {} rows and {} bytes with cluster-resize status: {}",
```
frankgh: NIT:
```suggestion
        LOGGER.info("Bulk writer job complete. rows={} bytes={} cluster_resize={}",
```

frankgh (commented):
Thanks for addressing the comments. I have a few follow up comments

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -129,6 +143,64 @@ private void persist(@NotNull JavaPairRDD<DecoratedKey, Object[]> sortedRDD, Str
...
+    }
+
+    /**
+     * Get a ref copy of BulkWriterContext broadcast variable and compose a function to transform a partition into StreamResult
+     *
+     * @param ctx BulkWriterContext broadcast variable
+     * @return FlatMapFunction
+     */
+    private static FlatMapFunction<Iterator<Tuple2<DecoratedKey, Object[]>>, WriteResult>
+    partitionsFlatMapFunc(Broadcast<BulkWriterContext> ctx, String[] columnNames)
```
JeetKunDoug: Why rename this function from `writeRowsInPartition` to `partitionsFlatMapFunc` - the first is a descriptive name for what the function does, the second is a descriptive name for what interface the returned value implements... one is "business domain" and the other is "implementation detail", and I think the former has much more value than the latter.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriterContext.java
```diff
@@ -21,7 +21,9 @@
 
 import java.io.Serializable;
 
-public interface BulkWriterContext extends Serializable
+import org.apache.cassandra.spark.common.JobStats;
+
+public interface BulkWriterContext extends Serializable, JobStats
```
JeetKunDoug: Rather than extending JobStats, please add a `jobStats` method to this interface and return an instance of something that implements `JobStats` like the other methods here do. We did this originally because the number of things this interface included was huge, so we decided to break them up into smaller implementation classes.

An example `JobStatsImpl` class might look something like this:
```java
package org.apache.cassandra.spark.bulkwriter;

import java.util.HashMap;
import java.util.Map;

import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

import org.apache.cassandra.spark.common.JobStats;

class JobStatsImpl implements JobStats
{
    private static final Logger LOGGER = LoggerFactory.getLogger(JobStatsImpl.class);;
    Map<String, String> stats = new HashMap<>();

    public JobStatsImpl(String sparkVersion, JobInfo jobInfo)
    {
        stats.put("sparkVersion", sparkVersion);
        stats.put("jobId", jobInfo.getId().toString());
        stats.put("keyspace", jobInfo.getId().toString());
        stats.put("table", jobInfo.getId().toString());
    }

    public void record(Map<String, String> statsToRecord)
    {
        stats.putAll(statsToRecord);
    }

    public void publish()
    {
        LOGGER.info("Job Stats: {}", stats);
    }
}
```

Then, CassandraBulkWriterContext just creates a new one of these in its constructor and returns it from the `jobStats` method. The actual field can/should be marked `transient` as Francisco has pointed out, so it doesn't get transferred to executors (and, therefore, it _doesn't_ need to be Serializable).

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RingInstance.java
JeetKunDoug: Either that, or if the fields really aren't used, can we simply remove them completely? But, to Francisco's point, we really should handle this cleanly otherwise.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/common/JobStats.java
JeetKunDoug: See my other comments about how to implement this in BulkWriterContext - may not need to be Serializable if we move it to an instance variable like we've done with other parts of the BulkWriterContext interface.

JeetKunDoug (changes requested):
Generally agree with Francisco's comments - please address them and let us know when you're ready for a second review round. I've added a few other requests here, but generally looks reasonable.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
arjunashok: Makes sense. Reverted

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RingInstance.java
arjunashok: Unfortunately, the fields are members of the `RingEntry` instance (part of the sidecar client library) so they cannot be removed here.  

I can add checks for the ones not in `ReplicaMetadata` for completeness, but the intention here was to not write the fields that we know are not used anyway.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriteResult.java
arjunashok: Correct. fixed.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
arjunashok: Addressed

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriterContext.java
arjunashok: Makes sense. Incorporated this change

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/common/stats/JobStatsImpl.java
```diff
@@ -0,0 +1,45 @@
...
+public class JobStatsImpl implements JobStats
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(JobStatsImpl.class);
+    private final transient Map<String, String> jobStats = new HashMap<>();
+    public void recordJobStats(Map<String, String> stats)
+    {
+        jobStats.putAll(stats);
+    }
+
+    public void publishJobStats()
```
frankgh: can we add override annotations?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RingInstance.java
```diff
@@ -130,6 +130,22 @@ private void writeObject(ObjectOutputStream out) throws IOException
         out.writeUTF(ringEntry.fqdn());
         out.writeUTF(ringEntry.status());
         out.writeUTF(ringEntry.state());
+        if (ringEntry.rack() != null)
```
frankgh: I think this is not sufficient, I pinged you offline for a suggestion

arjunashok: Addressed

frankgh (approved):
+1 Thanks for addressing the comments

PR #43, opened 2024-02-23
CASSANDRA-19424 Check for expired certificate during start up validation

sarankk:


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/validation/KeyStoreValidation.java
```diff
@@ -81,6 +85,15 @@ public void validate()
                 throw new RuntimeException("KeyStore is empty");
             }
 
+            for (Enumeration<String> aliases = keyStore.aliases(); aliases.hasMoreElements();)
+            {
+                Certificate cert = keyStore.getCertificate(aliases.nextElement());
+                if (cert instanceof X509Certificate && !(cert instanceof PemX509Certificate))
```
frankgh: I don't think we need to worry about `PemX509Certificate`s.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/validation/KeyStoreValidationTests.java
```diff
@@ -97,4 +98,15 @@ public void testValidKeyStore()
...
+
+    @Test
+    public void testExpiredKeyStore()
+    {
+        SecretsProvider secrets = TestSecretsProvider.forKeyStore("PKCS12", "keystore-expired.p12", "qwerty");
+        KeyStoreValidation validation = new KeyStoreValidation(secrets);
+
+        Throwable throwable = validation.perform();
+        assertInstanceOf(RuntimeException.class, throwable);
+        assertThat(throwable.getMessage()).startsWith("Certificate expired, valid NotAfter: ");
```
frankgh: the error message is odd, Should we just say:

`Certificate expired. NotAfter:` instead?

frankgh (commented):
Looks good, I have a couple of minor comments

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/validation/KeyStoreValidationTests.java
sarankk: Updated to `Certificate expired. ` instead

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/validation/KeyStoreValidation.java
```diff
@@ -81,6 +84,15 @@ public void validate()
...
             }
 
+            for (Enumeration<String> aliases = keyStore.aliases(); aliases.hasMoreElements();)
+            {
+                Certificate cert = keyStore.getCertificate(aliases.nextElement());
+                if (cert instanceof X509Certificate)
+                {
+                    ((X509Certificate) cert).checkValidity();
+                }
+            }
```
frankgh: Can we collapse the for lops in lines 87 and 96:
```suggestion
            for (Enumeration<String> aliases = keyStore.aliases(); aliases.hasMoreElements();)
            {
                String alias = aliases.nextElement();
                Certificate cert = keyStore.getCertificate(alias);
                if (cert instanceof X509Certificate)
                {
                    ((X509Certificate) cert).checkValidity();
                }

                Key key = keyStore.getKey(alias, password);
                if (key != null && key instanceof PrivateKey)
                {
                    return;  // KeyStore contains a private key
                }
            }
```

frankgh (commented):
One additional minor comment

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/validation/KeyStoreValidation.java
frankgh: ah, I just realized that the second loop looks for a PK< and it if finds it, it exits the validation as successful. Please disregard this comment.

frankgh (approved):
+1

yifan-c (approved):

PR #44, opened 2024-02-29
CASSANDRA-19452 Use constant reference time during bulk read process

yifan-c:
patch by Yifan Cai; reviewed by XX for CASSANDRA-19452

### DIFF
--- a/cassandra-bridge/src/main/java/org/apache/cassandra/spark/data/DataLayer.java
+++ b/cassandra-bridge/src/main/java/org/apache/cassandra/spark/data/DataLayer.java
@@ -164,6 +164,11 @@ public abstract class DataLayer implements Serializable
 
     public abstract boolean isInPartition(int partitionId, BigInteger token, ByteBuffer key);
 
+    /**
+     * @return a TimeProvider
+     */
+    public abstract TimeProvider timeProvider();
+
     public List<PartitionKeyFilter> partitionKeyFiltersInRange(
     int partitionId,
     List<PartitionKeyFilter> partitionKeyFilters) throws NoMatchFoundException
@@ -280,14 +285,6 @@ public abstract class DataLayer implements Serializable
         SparkRangeFilter rangeFilter = sparkRangeFilter(partitionId);
         return bridge().getPartitionSizeIterator(cqlTable(), partitioner(), sstables(partitionId, rangeFilter, Collections.emptyList()),
                                                  rangeFilter, timeProvider(), stats(), executorService());
-    }
-
-    /**
-     * @return a TimeProvider that returns the time now in seconds. User can override with their own provider
-     */
-    public TimeProvider timeProvider()
-    {
-        return bridge().timeProvider();
     }
 
     /**
--- a/cassandra-bridge/src/main/java/org/apache/cassandra/spark/utils/ReaderTimeProvider.java
+++ b/cassandra-bridge/src/main/java/org/apache/cassandra/spark/utils/ReaderTimeProvider.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cassandra.spark.utils;
+
+/**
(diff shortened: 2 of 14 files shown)


cassandra-bridge/src/main/java/org/apache/cassandra/spark/utils/test/TestSchema.java
```diff
@@ -398,9 +409,11 @@ private String buildCreateStatement(List<CqlField> columns,
 
         createStmtBuilder.append("))");
 
+        createStmtBuilder.append(" WITH comment= 'test table'"); // take 'WITH', so the rest can append 'AND' safely
```
jberragan:
```suggestion
        createStmtBuilder.append(" WITH comment = 'test table'"); // take 'WITH', so the rest can append 'AND' safely
```

cassandra-bridge/src/main/java/org/apache/cassandra/spark/utils/TimeProvider.java
```diff
@@ -19,16 +19,37 @@
...
 /**
  * Provides current time
   */
-@FunctionalInterface
 public interface TimeProvider
 {
-    TimeProvider INSTANCE = () -> (int) Math.floorDiv(System.currentTimeMillis(), 1000L);
+    TimeProvider DEFAULT = new TimeProvider()
+    {
+        private final int referenceEpoch = nowInSeconds();
```
jberragan: I worry this will be a gotcha for those unfamiliar with `referenceEpoch` and why it is required. Many users will not realize they need to override TimeProvider.

If they have long running Spark clusters, this `referenceEpoch` will be defined once and never updated, potentially resulting in values appearing that should be been TTL'ed

yifan-c: My thinking is that SBR reads a snapshot. Ideally, whatever is unexpired at the moment of taking snapshot, should be read back, making it "snapshot-isolated-ish". Using a fixed during the job can archive it, especially the job creates a snapshot. 
What is the "long running spark cluster"? Is it long running because the _one_ snapshot has a large amount of data?

cassandra-bridge/src/main/java/org/apache/cassandra/spark/utils/ReaderTimeProvider.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package org.apache.cassandra.spark.utils;
+
+/**
+ * A {@link TimeProvider} used by reader. It decides the reference epoch on instantiation and distribute to executors
+ */
+public class ReaderTimeProvider implements TimeProvider
+{
+    private static final long serialVersionUID = 880314923071431541L;
```
frankgh: probably not needed since we only serialize the int

cassandra-bridge/src/main/java/org/apache/cassandra/spark/utils/ReaderTimeProvider.java
```diff
@@ -0,0 +1,50 @@
...
+
+    private final int referenceEpoch;
```
frankgh: Explicitly define this variable
```suggestion
    private final int referenceEpochInSeconds;
```

cassandra-bridge/src/main/java/org/apache/cassandra/spark/utils/TimeProvider.java
frankgh: maybe we can preserve existing behavior by returning the current seconds every time.

```
        @Override
         public int referenceEpochInSeconds()
         {
             return nowInSeconds();
         }
```

One clarification, SBR when running with the CassandraDataLayer, reads a snapshot. The same is not true for when you are using the `LocalDataLayer` for example.

frankgh (commented):
Looks good in general, I've left a few comments

cassandra-bridge/src/main/java/org/apache/cassandra/spark/utils/TimeProvider.java
yifan-c: The reason of using a single time point is applicable to SBR in general.
The expectation of bulk read is no different than a single point read. It is surprising if data visibility changes (caused by the advancing time) during the read. 

Bulk read is essentially a transformation from a collection of sstable to a collection of mutations. The time reference has to be the same in order to get the same collection of mutations deterministically. The existing behavior is a bug, especially when TTL is considered. It has little to do with test setup or prod use case.

cassandra-bridge/src/main/java/org/apache/cassandra/spark/utils/ReaderTimeProvider.java
yifan-c: yeah. Forgot to remove the field. The class was serializable.

jberragan (approved):
+1, I just think we need to leave a comment explaining why `referenceEpoch` is required and why user should override the `TimeProvider.DEFAULT` value.

frankgh (approved):
+1 Looks good to me. Thanks for the improvement

cassandra-bridge/src/main/java/org/apache/cassandra/spark/data/DataLayer.java
```diff
@@ -164,6 +164,11 @@ public CassandraVersion version()
 
     public abstract boolean isInPartition(int partitionId, BigInteger token, ByteBuffer key);
 
+    /**
+     * @return a TimeProvider
+     */
+    public abstract TimeProvider timeProvider();
```
yifan-c, line 170: Do not return DEFAULT time prover. The concrete implementation should return the correct value.

PR #45, opened 2024-03-07
CASSANDRA-19340 - Support writing UDTs

JeetKunDoug:
This commit brings support for UDTs to the Bulk Writer. It converts Spark nested StructType(s) into UDT values.

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
@@ -54,6 +54,8 @@ public class CassandraBulkWriterContext implements BulkWriterContext, KryoSerial
     @NotNull
     private final BulkSparkConf conf;
     private final JobInfo jobInfo;
+    private final String lowestCassandraVersion;
+    private transient CassandraBridge bridge;
     private transient DataTransferApi dataTransferApi;
     private final CassandraClusterInfo clusterInfo;
     private final SchemaInfo schemaInfo;
@@ -68,9 +70,8 @@ public class CassandraBulkWriterContext implements BulkWriterContext, KryoSerial
         this.conf = conf;
         this.clusterInfo = clusterInfo;
         this.jobStatsPublisher = new LogStatsPublisher();
-        String lowestCassandraVersion = clusterInfo.getLowestCassandraVersion();
-        CassandraBridge bridge = CassandraBridgeFactory.get(lowestCassandraVersion);
-
+        lowestCassandraVersion = clusterInfo.getLowestCassandraVersion();
+        this.bridge = CassandraBridgeFactory.get(lowestCassandraVersion);
         TokenRangeMapping<RingInstance> tokenRangeMapping = clusterInfo.getTokenRangeMapping(true);
         jobInfo = new CassandraJobInfo(conf,
                                        new TokenPartitioner(tokenRangeMapping,
@@ -92,11 +93,23 @@ public class CassandraBulkWriterContext implements BulkWriterContext, KryoSerial
         Set<String> udts = CqlUtils.extractUdts(keyspaceSchema, keyspace);
         ReplicationFactor replicationFactor = CqlUtils.extractReplicationFactor(keyspaceSchema, keyspace);
         int indexCount = CqlUtils.extractIndexCount(keyspaceSchema, keyspace, table);
-        CqlTable cqlTable = bridge.buildSchema(createTableSchema, keyspace, replicationFactor, partitioner, udts, null, indexCount);
+        CqlTable cqlTable = bridge().buildSchema(createTableSchema, keyspace, replicationFactor, partitioner, udts, null, indexCount);
 
         TableInfoProvider tableInfoProvider = new CqlTableInfoProvider(createTableSchema, cqlTable);
         TableSchema tableSchema = initializeTableSchema(conf, dfSchema, tableInfoProvider, lowestCassandraVersion);
-        schemaInfo = new CassandraSchemaInfo(tableSchema);
+        schemaInfo = new CassandraSchemaInfo(tableSchema, udts, cqlTable);
+    }
+
+    @Override
+    public CassandraBridge bridge()
+    {
+        CassandraBridge currentBridge = this.bridge;
+        if (currentBridge != null)
+        {
+            return currentBridge;
+        }
+        this.bridge = CassandraBridgeFactory.get(lowestCassandraVersion);
+        return bridge;
     }
 
     public static BulkWriterContext fromOptions(@NotNull SparkContext sparkContext,
@@ -204,9 +217,8 @@ public class CassandraBulkWriterContext implements BulkWriterContext, KryoSerial
     {
         if (dataTransferApi == null)
         {
-            CassandraBridge bridge = CassandraBridgeFactory.get(clusterInfo.getLowestCassandraVersion());
             dataTransferApi = new SidecarDataTransferApi(clusterInfo.getCassandraContext(),
-                                                         bridge,
+                                                         bridge(),
                                                          jobInfo,
                                                          conf);
         }
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraSchemaInfo.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraSchemaInfo.java
@@ -19,19 +19,31 @@
 
 package org.apache.cassandra.spark.bulkwriter;
 
+import java.util.Set;
+
+import org.apache.cassandra.spark.data.CqlTable;
+
 public class CassandraSchemaInfo implements SchemaInfo
 {
     private static final long serialVersionUID = -2327383232935001862L;
     private final TableSchema tableSchema;
+    private final Set<String> userDefinedTypeStatements;
 
-    public CassandraSchemaInfo(TableSchema tableSchema)
+    public CassandraSchemaInfo(TableSchema tableSchema, Set<String> userDefinedTypeStatements, CqlTable cqlTable)
     {
         this.tableSchema = tableSchema;
+        this.userDefinedTypeStatements = userDefinedTypeStatements;
     }
 
     @Override
     public TableSchema getTableSchema()
     {
         return tableSchema;
+    }
+
+    @Override
+    public Set<String> getUserDefinedTypeStatements()
+    {
+        return userDefinedTypeStatements;
     }
 }
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CqlTableInfoProvider.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CqlTableInfoProvider.java
@@ -98,9 +98,20 @@ public class CqlTableInfoProvider implements TableInfoProvider
     @Override
     public List<ColumnType<?>> getPartitionKeyTypes()
     {
-        return cqlTable.partitionKeys().stream()
-                       .map(cqlField -> DATA_TYPES.get(cqlField.type().cqlName().toLowerCase()))
-                       .collect(Collectors.toList());
+        List<ColumnType<?>> types = cqlTable.partitionKeys().stream()
+                                              .map(cqlField -> {
+                                                  String typeName = cqlField.type().cqlName().toLowerCase();
+                                                  ColumnType<?> type = DATA_TYPES.get(typeName);
+                                                  if (type == null)
+                                                  {
+                                                      throw new RuntimeException(
+                                                      "Could not find ColumnType for type name" + typeName);
+                                                  }
+                                                  return type;
+                                              })
+                                              .collect(Collectors.toList());
+        return types;
+
     }
 
     @Override
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
@@ -37,6 +37,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
@@ -44,6 +45,7 @@ import java.util.stream.Stream;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Range;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -51,6 +53,10 @@ import org.slf4j.LoggerFactory;
 import o.a.c.sidecar.client.shaded.common.data.TimeSkewResponse;
 import org.apache.cassandra.spark.bulkwriter.token.ReplicaAwareFailureHandler;
 import org.apache.cassandra.spark.bulkwriter.token.TokenRangeMapping;
+import org.apache.cassandra.spark.data.BridgeUdtValue;
+import org.apache.cassandra.spark.data.CqlField;
+import org.apache.cassandra.spark.data.CqlTable;
+import org.apache.cassandra.spark.data.ReplicationFactor;
 import org.apache.cassandra.spark.utils.DigestAlgorithm;
 import org.apache.spark.InterruptibleIterator;
 import org.apache.spark.TaskContext;
@@ -61,6 +67,8 @@ import static org.apache.cassandra.spark.utils.ScalaConversionUtils.asScalaItera
 @SuppressWarnings({ "ConstantConditions" })
 public class RecordWriter implements Serializable
 {
+    public static final ReplicationFactor IGNORED_REPLICATION_FACTOR = new ReplicationFactor(ReplicationFactor.ReplicationStrategy.SimpleStrategy,
+                                                                                             ImmutableMap.of("replication_factor", 1));
     private static final Logger LOGGER = LoggerFactory.getLogger(RecordWriter.class);
     private static final long serialVersionUID = 3746578054834640428L;
     private final BulkWriterContext writerContext;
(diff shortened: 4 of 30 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
```diff
@@ -64,7 +65,6 @@ private CassandraBulkWriterContext(@NotNull BulkSparkConf conf,
         this.conf = conf;
         this.clusterInfo = clusterInfo;
         String lowestCassandraVersion = clusterInfo.getLowestCassandraVersion();
-        CassandraBridge bridge = CassandraBridgeFactory.get(lowestCassandraVersion);
```
frankgh, line 72: Nothing wrong still initializing the bridge here. Should we preserve it?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
```diff
@@ -87,11 +87,21 @@ private CassandraBulkWriterContext(@NotNull BulkSparkConf conf,
...
-        schemaInfo = new CassandraSchemaInfo(tableSchema);
+        schemaInfo = new CassandraSchemaInfo(tableSchema, udts, cqlTable);
+    }
+
+    @Override
+    public synchronized CassandraBridge bridge()
+    {
+        if (this.bridge == null)
+        {
+            this.bridge = CassandraBridgeFactory.get(this.clusterInfo.getLowestCassandraVersion());
```
frankgh: it feels like the driver should perform this operation once (it's a potentially expensive operation) and we should serialize the resulting version value to the executors. Executors should just use the version value that the driver computed.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CqlTableInfoProvider.java
```diff
@@ -98,9 +98,20 @@ public CqlField.CqlType getColumnType(String columnName)
     @Override
     public List<ColumnType<?>> getPartitionKeyTypes()
     {
-        return cqlTable.partitionKeys().stream()
-                       .map(cqlField -> DATA_TYPES.get(cqlField.type().cqlName().toLowerCase()))
-                       .collect(Collectors.toList());
+        List<ColumnType<?>> types = cqlTable.partitionKeys().stream()
+                                              .map(cqlField -> {
+                                                  String typeName = cqlField.type().cqlName().toLowerCase();
```
frankgh, line 103: why do we lower case the cqlName? Can we add tests for quoted UDTs as well?

frankgh (commented):
Not a full review yet, but I will push my review so far so it doesn't get lost

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
JeetKunDoug: Mostly a matter of getting the lowest version as late as possible - imagine a mid-upgrade scenario where a node comes back online with a newer version (and it was the last one to be upgraded)... In the grand scheme of things, this is pretty unlikely to happen, and the previous version's libraries will still work, but it seemed worthwhile to allow it to be lazy.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
JeetKunDoug: Same comment as above - it's possible the "lowest version" has changed from the beginning of the job. Again, maybe not worth the extra load but I don't think it's particularly heavy-weight.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CqlTableInfoProvider.java
JeetKunDoug: Note that this map (`DATA_TYPES`) currently doesn't support UDTs at all - it's something we need to implement separately to support UDTs for partition/clustering key columns (as `Tokenizer` doesn't support them at this time). So, for now, we throw if we can't find the type in that map, but we'll need to update the logic once the Tokenizer supports them.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
frankgh: On a smaller cluster maybe not, but I think it can add up on a big Cassandra cluster when running a large spark job. I think there's a tradeoff here. In the case where that one instance coming back with a lower version, the instance would probably miss the write (as it would be unable to import the SStables) and the data would need to be repaired. So the question is whether we want so save some REST endpoint requests on all the job runs vs save on repair cost on a very unlikely scenario that would very rarely happen.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
yifan-c: So, as you said, the effect is the same for both eager or lazy instantiation. But you can avoid synchronization if instantiate eagerly.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
yifan-c: It is safe (or safer) to use the low version among the entire job runtime, which is essential the low version when the job starts. 
Cassandra can take the sstable written with low version. 
There could be a risk that a node has to downgrade and the written sstables are in the new version. The scenario is rare, but you get my point.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraSchemaInfo.java
```diff
@@ -19,19 +19,33 @@
...
+import java.util.Set;
+
+import org.apache.cassandra.spark.data.CqlTable;
+
 public class CassandraSchemaInfo implements SchemaInfo
 {
     private static final long serialVersionUID = -2327383232935001862L;
     private final TableSchema tableSchema;
+    private final Set<String> userDefinedTypeStatements;
+    private final CqlTable cqlTable;
```
yifan-c: variable not exposed and not used anywhere

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -95,6 +105,7 @@ public RecordWriter(BulkWriterContext writerContext, String[] columnNames)
         this.writeValidator = new BulkWriteValidator(writerContext, failureHandler);
         this.digestAlgorithm = this.writerContext.job().digestAlgorithmSupplier().get();
 
+
```
yifan-c: nit: remove the new empty line

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -108,6 +119,26 @@ private String getStreamId(TaskContext taskContext)
         return String.format("%d-%s", taskContext.partitionId(), UUID.randomUUID());
     }
 
+    private CqlTable cqlTable()
```
yifan-c, line 120: The only callsite of `cqlTable()` is already `synchronized` on record writer. I think you can remove the double-checked locking entirely, if keeping the synchronization at `getUdt()`.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SqlToCqlTypeConverter.java
```diff
@@ -763,6 +766,12 @@ else if (object instanceof Map)
             throw new RuntimeException("Unsupported conversion for MAP from " + object.getClass().getTypeName());
         }
 
+        @Override
+        public String toString()
+        {
+            return "Map";
```
yifan-c: How about the richer string? 

```suggestion
            return "Map<" + keyConverter.toString() + ", " + valConverter.toString() + '>';
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SqlToCqlTypeConverter.java
```diff
@@ -788,11 +797,54 @@ else if (object instanceof scala.Tuple2)
             }
             return map;
         }
+    }
+
+    public static class UdtConverter extends NullableConverter<Object>
+    {
+        private final String name;
+        private final HashMap<Object, Converter<?>> converters;
```
yifan-c: Key is `String`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SqlToCqlTypeConverter.java
```diff
@@ -788,11 +797,54 @@ else if (object instanceof scala.Tuple2)
...
+        private CqlField.CqlUdt udt;
```
yifan-c: Unused.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SqlToCqlTypeConverter.java
```diff
@@ -788,11 +797,54 @@ else if (object instanceof scala.Tuple2)
             }
             return map;
         }
+    }
+
+    public static class UdtConverter extends NullableConverter<Object>
```
yifan-c: Why it converts to `Object` instead of `BridgeUdtValue`?

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteUdtTest.java
```diff
@@ -0,0 +1,208 @@
...
+        assertThat(actualEntries.size()).isEqualTo(sourceData.size());
+
+        // remove from actual entries to make sure that the data read is the same as the data written
+        Set<String> sourceEntries = sourceData.stream().map(BulkWriteUdtTest::getFormatedSourceEntry)
+                           .collect(Collectors.toSet());
+        assertThat(actualEntries).as("All entries are expected to be read from database")
+                                 .containsExactlyInAnyOrderElementsOf(sourceEntries);
+    }
+
+    private static String getFormatedSourceEntry(Row row)
```
yifan-c: typo: `getFormattedSourceEntry` (double t)

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/TableSchemaNormalizeTest.java
```diff
@@ -296,6 +303,24 @@ public void testNestedNormalization()
...
 
+    @Test
+    public void testUdtNormalization()
+    {
+        StructType structType = new StructType()
+                                .add(new StructField("f1", DataTypes.IntegerType, false, Metadata.empty()))
+                                .add(new StructField("f2", DataTypes.StringType, false, Metadata.empty()));
+
+        GenericRowWithSchema source = new GenericRowWithSchema(new Object[]{1, "course"}, structType);
+        // NOTE: UDT Types cary their type name around, so the use of `udt_field` consistently here is a bit
```
yifan-c: `carry`?

cassandra-four-zero/src/main/java/org/apache/cassandra/bridge/SSTableWriterImplementation.java
```diff
@@ -81,17 +85,23 @@ static CQLSSTableWriter.Builder configureBuilder(String inDirectory,
...
+                                           .using(insertStatement)
+                                           // The data frame to write is always sorted,
+                                           // see org.apache.cassandra.spark.bulkwriter.CassandraBulkSourceRelation.insert
+                                           .sorted()
+                                           .withMaxSSTableSizeInMiB(bufferSizeMB);
+        for (String udt : udts)
+        {
+            builder.withType(udt);
+        }
+        return builder;
```
yifan-c: nit
```suggestion
        CQLSSTableWriter.Builder builder = CQLSSTableWriter.builder();
        for (String udt : udts)
        {
            builder.withType(udt);
        }
        
        return builder
               .inDirectory(inDirectory)
               .forTable(createStatement)
               .withPartitioner(cassPartitioner)
               .using(insertStatement)
               // The data frame to write is always sorted,
               // see org.apache.cassandra.spark.bulkwriter.CassandraBulkSourceRelation.insert
               .sorted()
               .withMaxSSTableSizeInMiB(bufferSizeMB);
```

yifan-c (commented):
mostly nits for cosmetic changes

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SqlToCqlTypeConverter.java
JeetKunDoug: Ha - sure... although technically this PR didn't actually change that, it's just a formatting thing

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SqlToCqlTypeConverter.java
JeetKunDoug: Originally, I was trying to convert to the final values, but it ended up it wasn't possible due to some serialization issues, so it returns the `BridgeUdtValue` - I'll update it to be more specific.

cassandra-four-zero/src/main/java/org/apache/cassandra/bridge/SSTableWriterImplementation.java
JeetKunDoug: I'm not sure why this is any different/better than the existing code - just less indented?

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/MockBulkWriterContext.java
```diff
@@ -77,6 +79,7 @@ public class MockBulkWriterContext implements BulkWriterContext, ClusterInfo, Jo
     private ConsistencyLevel.CL consistencyLevel;
     private int sstableDataSizeInMB = 128;
     private int sstableWriteBatchSize = 2;
+    private CassandraBridge bridge = CassandraBridgeFactory.get(CassandraVersion.FOURZERO);
```
frankgh, line 82: should this be passed to the mock BW context? I wonder what tests for 50, 51 will look like if we use 40 here always?

cassandra-bridge/src/main/java/org/apache/cassandra/spark/data/BridgeUdtValue.java
```diff
@@ -0,0 +1,69 @@
...
+        BridgeUdtValue udtValue = (BridgeUdtValue) o;
+        return Objects.equals(name, udtValue.name) && Objects.equals(udtMap, udtValue.udtMap);
+    }
+
+    public int hashCode()
+    {
+        return Objects.hash(name, udtMap);
+    }
+
+    public String toString()
```
frankgh, line 62: NIT, override annotation

cassandra-bridge/src/main/java/org/apache/cassandra/spark/data/BridgeUdtValue.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+        if (o == null || getClass() != o.getClass())
+        {
+            return false;
+        }
+        BridgeUdtValue udtValue = (BridgeUdtValue) o;
+        return Objects.equals(name, udtValue.name) && Objects.equals(udtMap, udtValue.udtMap);
+    }
+
+    public int hashCode()
```
frankgh, line 57: NIT, override annotation

cassandra-bridge/src/main/java/org/apache/cassandra/spark/data/BridgeUdtValue.java
```diff
@@ -0,0 +1,69 @@
...
+    public final String name;
+    public final Map<String, Object> udtMap;
+
+    public BridgeUdtValue(String name, Map<String, Object> valueMap)
+    {
+        this.name = name;
+        this.udtMap = valueMap;
+    }
+
+    public boolean equals(Object o)
```
frankgh, line 43: NIT, override annotation

JeetKunDoug:
Rebased, addressed comments, green CircleCI: https://app.circleci.com/pipelines/github/JeetKunDoug/cassandra-analytics/141/workflows/9d8c886c-81c9-4d43-9e78-43a52fab14f9

JeetKunDoug:
Rebased/squashed/fixed CHANGES.txt conflict (no other changes/conflicts)

yifan-c (approved):

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
```diff
@@ -92,11 +92,17 @@ protected CassandraBulkWriterContext(@NotNull BulkSparkConf conf,
...
         TableInfoProvider tableInfoProvider = new CqlTableInfoProvider(createTableSchema, cqlTable);
         TableSchema tableSchema = initializeTableSchema(conf, dfSchema, tableInfoProvider, lowestCassandraVersion);
-        schemaInfo = new CassandraSchemaInfo(tableSchema);
+        schemaInfo = new CassandraSchemaInfo(tableSchema, udts, cqlTable);
+    }
+
+    @Override
+    public CassandraBridge bridge()
+    {
+        return this.bridge;
```
frankgh: if this is serialized to executors, and it's declared as transient, we'll need to initialize it for the executors at some point. I suggest we add a field called `private final String lowestCassandraVersion;` and we initialize it in the constructor (line 72). I think at the executor level we'll see NPEs for the bridge

frankgh (commented):
I've added one additional comment that I think needs to be addressed

frankgh (approved):
+1 Looks good. Thanks

PR #46, opened 2024-03-27
CASSANDRA-19500 Mask hash value produced by lz4's xxhash implementation

frankgh:
This PR bumps the Sidecar version to the current latest HEAD of Sidecar. Bumping the version surfaced an issue in the LZ4's implementation of the XXHash32 hash value. The hash value is not masked and this causes the negative sign to be forwarded producing the incorrect hash result. Other implemnentations correctly apply the mask to the produced value (i.e commons-codec).

We do not consider switching to `commons-codec` because Spark has a dependency of codec, and in older versions of Spark (2), the XXHash32 implementation is either non existent or the APIs have changed.

Patch by Francisco Guerrero; Reviewed by TBD for CASSANDRA-19500


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -43,20 +43,20 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import o.a.c.sidecar.client.shaded.common.NodeSettings;
```
frankgh, line 46: the new version of Sidecar relocates the common dependencies to o.a.c.sidecar.client.shaded . See https://issues.apache.org/jira/browse/CASSANDRASC-104

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/utils/XXHash32DigestAlgorithmTest.java
```diff
@@ -51,8 +51,8 @@ class XXHash32DigestAlgorithmTest
 
     @ParameterizedTest(name = "{index} fileName={0} expectedMd5={1}")
     @CsvSource({
-    "file1.txt,ffffffffd76a44a5",
-    "file2.txt,ffffffffef976cbe",
+    "file1.txt,d76a44a5",
```
frankgh, line 54: now it matches the expected xxh32sum

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/utils/XXHash32DigestAlgorithm.java
```diff
@@ -61,7 +61,10 @@ public Digest calculateFileDigest(Path path) throws IOException
             {
                 hasher.update(buffer, 0, len);
             }
-            return new XXHash32Digest(Long.toHexString(hasher.getValue()), SEED);
+            // lz4 library doesn't mask the hash value, so we need to mask it to
+            // prevent forwarding the negative sign bit when converting to a long value
+            long hash = hasher.getValue() & 0xffffffffL;
```
yifan-c: The comment is not true. 
There is no mask required for lz4's implementation since it returns `int`. Masking is required because our code casts `int` to `long`, and the sign bit cannot be carried over. 
Alternatively, you can just call this w/o masking.

```java
Integer.toHexString(hasher.getValue())
```

And the proof.
```
jshell> int i = -1;
i ==> -1

jshell> Integer.toHexString(i);
$2 ==> "ffffffff"

jshell> long l = i & 0xffffffffL;
l ==> 4294967295

jshell> Long.toHexString(l);
$4 ==> "ffffffff"
```

frankgh: that makes sense. Thanks for the clarification. I've pushed a commit to address this.

PR #53, opened 2024-04-16
CASSANDRA-19563: Support bulk write via S3

yifan-c:
This commit adds a configuration (writer) option to pick a transport other than the previously-implemented "direct upload to all sidecars" (now known as the "Direct" transport).  The second transport, now being implemented, is the "S3_COMPAT" transport, which allows the job to upload the generated SSTables to an S3-compatible storage system, and then inform the Cassandra Sidecar that those files are available for download & commit.

Additionally, a plug-in system was added to allow communications between custom transport hooks and the job, so the custom hook can provide updated credentials and out-of-band status updates on S3-related issues.

Co-Authored-By: Yifan Cai <ycai@apache.org>
Co-Authored-By: Doug Rohrer <drohrer@apple.com>
Co-Authored-By: Francisco Guerrero <frankgh@apache.org>
Co-Authored-By: Saranya Krishnakumar <saranya_k@apple.com>

Patch by Yifan Cai, Doug Rohrer, Francisco Guerrero, Saranya Krishnakumar; Reviewed by TBD for CASSANDRA-19563


gradle.properties
```diff
@@ -16,7 +16,7 @@
 # under the License.
 
 group=org.apache.cassandra.spark
-version=1.0.0
+version=1.0.0.36
```
frankgh: this shouldn't change
```suggestion
version=1.0.0
```

gradle.properties
```diff
@@ -31,5 +31,6 @@ spark=3
 # force version 4.5.1 of vertx to prevent issues initializing io.vertx.core.json.jackson.JacksonCodec,
 # which requires a newer version of jackson, which is not available in spark 2
 vertxVersion=4.5.1
+aswSdkVersion=2.20.43
```
frankgh: this is trailing very behind in the current dependency. Should we bump this to the latest or a more recent version?

scripts/build-sidecar.sh
```diff
@@ -47,22 +47,27 @@ else
   cd "${SIDECAR_BUILD_DIR}"
   echo "branch ${SIDECAR_BRANCH} sha ${SIDECAR_COMMIT}"
   # check out the correct cassandra version:
+  # if the SIDECAR_BRANCH directory does not exist; we initialize from the scratch
```
frankgh, line 50: great comments!

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/utils/XXHash32DigestAlgorithmTest.java
```diff
@@ -49,7 +49,7 @@ class XXHash32DigestAlgorithmTest
     // $ xxh32sum file2.txt # -> ef976cbe
     // $ xxh32sum file3.txt # -> 08321e1e
 
-    @ParameterizedTest(name = "{index} fileName={0} expectedMd5={1}")
+    @ParameterizedTest(name = "{index} fileName={0} expectedDigest={1}")
```
frankgh, line 52: 👍

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CancelJobEvent.java
```diff
@@ -0,0 +1,38 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cassandra.spark.bulkwriter;
+
+public class CancelJobEvent
```
frankgh, line 26: can we add some javadocs?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
```diff
@@ -19,16 +19,42 @@
...
+    /**
+     * Check consistency level with the collection of the succeeded instances
+     *
+     * @param succeededInstances the succeeded instances in the replica set
+     * @param replicationFactor replication factor to check with
+     * @param localDC the local data center name if required for the check
+     * @return true means the consistency level is _definitively_ satisfied.
+     *         Meanwhile, returning false means no conclusion can be drawn
+     */
+    boolean canBeSatisfied(Collection<? extends CassandraInstance> succeededInstances,
```
frankgh, line 49: it feels like this class should have unit tests. Maybe a follow up?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/blobupload/StorageClientConfig.java
```diff
@@ -0,0 +1,83 @@
...
+        this.nioHttpClientMaxConcurrency = nioHttpClientMaxConcurrency;
+    }
+
+    private URI toURI(String uriString, String hint)
+    {
+        if (uriString == null)
+        {
+            return null;
+        }
+        else
```
frankgh: NIT: unnecessary else block adds unnecessary indentation

scripts/build-sidecar.sh
```diff
@@ -24,7 +24,7 @@ else
   SCRIPT_DIR=$( dirname -- "$( readlink -f -- "$0"; )"; )
   SIDECAR_REPO="${SIDECAR_REPO:-https://github.com/apache/cassandra-sidecar.git}"
   SIDECAR_BRANCH="${SIDECAR_BRANCH:-trunk}"
-  SIDECAR_COMMIT="${SIDECAR_COMMIT:-20795db4d708b9287e0a2281695923bfb6fa9138}"
+  SIDECAR_COMMIT="${SIDECAR_COMMIT:-686d9e8699bd0a56857e24523d883120023b9841}"
```
frankgh: It looks like there's a new SHA, should we bump it ?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriterContext.java
```diff
@@ -26,6 +26,8 @@
 
 public interface BulkWriterContext extends Serializable
 {
+    BulkSparkConf conf();
```
frankgh: this feels out of place. The `JobInfo` seems to be the abstraction to expose the configured values for the context, but now we have both the abstraction and the direct values in the same class. Why have the abstraction in the first place if we are just going to expose the raw configuration here?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraTopologyMonitor.java
```diff
@@ -0,0 +1,107 @@
...
+                return;
+            }
+            retryCount = 0;
+        }
+        catch (Exception exception)
+        {
+            if (retryCount++ > MAX_CHECK_ATTEMPTS)
+            {
+                LOGGER.error("Could not retrieve current topology. All hosts exhausted. The retrieval has failed consecutive for {} times", retryCount);
+                onCancelJob.accept(new CancelJobEvent("Could not retrieve current cassandra topology. All hosts and retry have exhausted."));
```
frankgh: NIT
```suggestion
                onCancelJob.accept(new CancelJobEvent("Could not retrieve current cassandra topology. All hosts and retries have been exhausted."));
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -97,13 +111,26 @@ public String getPackageVersion(String lowestCassandraVersion)
...
+        // first row
+        if (rowCount == 0)
         {
             minToken = token;
+            maxToken = token;
+        }
+        else
+        {
+            // rows are sorted. Therefore, only update the maxToken
+            maxToken = token;
```
frankgh: NIT: maxToken is always updated so no need to have it in an if/else branch
```suggestion
        if (rowCount == 0)
        {
            minToken = token;
        }
        maxToken = token;
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
```diff
@@ -61,4 +61,16 @@ public enum WriterOptions implements WriterOption
      * If unspecified, it defaults to {@code XXHash32} digests. The legacy {@code MD5} digest is also supported.
      */
     DIGEST,
+    DATA_TRANSPORT,
```
frankgh, line 73: can we add javadocs for these new options?

frankgh (commented):
Looking good so far, I have a few comments.

scripts/build-sidecar.sh
yifan-c: will do

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
yifan-c: Adding test in this patch.

scripts/build-sidecar.sh
yifan-c: Updated to `fd6f7ac5f9f19dbbeeb9e7f80ca1fcbf60d5a4c6`

frankgh (approved):
+1 Thanks for addressing the comments . Looks good to me

PR #59, opened 2024-05-11
CASSANDRA-19634: Improve test coverage for downed instances

frankgh:
Patch by Francisco Guerrero; Reviewed by Yifan Cai for CASSANDRA-19634

### DIFF
--- a/cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
+++ b/cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
@@ -27,15 +27,15 @@ import java.net.UnknownHostException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import java.util.stream.Collectors;
-import java.util.stream.IntStream;
+import java.util.stream.StreamSupport;
 
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
@@ -142,19 +142,15 @@ import static org.assertj.core.api.Assertions.assertThat;
 @ExtendWith(VertxExtension.class)
 public abstract class SharedClusterIntegrationTestBase
 {
-    public static final String SIDECAR_S3_ENDPOINT_OVERRIDE_OPT = "S3_ENDPOINT_OVERRIDE";
-
     protected final Logger logger = LoggerFactory.getLogger(SharedClusterIntegrationTestBase.class);
     private static final int MAX_CLUSTER_PROVISION_RETRIES = 5;
 
     @TempDir
     static Path secretsPath;
 
-    protected Vertx vertx;
-    protected DnsResolver dnsResolver;
+    protected DnsResolver dnsResolver = new LocalhostResolver();
     protected IClusterExtension<? extends IInstance> cluster;
     protected Server server;
-    protected Injector injector;
     protected TestVersion testVersion;
     protected MtlsTestHelper mtlsTestHelper;
     private IsolatedDTestClassLoaderWrapper classLoaderWrapper;
@@ -321,9 +317,9 @@ public abstract class SharedClusterIntegrationTestBase
     /**
      * Override to provide additional options to configure sidecar
      */
-    protected Map<String, String> sidecarAdditionalOptions()
+    protected Function<SidecarConfigurationImpl.Builder, SidecarConfigurationImpl.Builder> configurationOverrides()
     {
-        return Collections.emptyMap();
+        return null;
     }
 
     /**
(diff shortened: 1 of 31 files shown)


cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
```diff
@@ -428,9 +436,11 @@ protected Object[][] queryAllData(QualifiedName table)
      * @param consistencyLevel the consistency level to use for querying the data
      * @return all the data queried from the table
      */
-    protected Object[][] queryAllData(QualifiedName table, ConsistencyLevel consistencyLevel)
+    protected Object[][] queryAllData(QualifiedName table, String consistencyLevel)
```
yifan-c: nit: prefer type over string

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
```diff
@@ -486,20 +496,23 @@ public static Cluster createDriverCluster(ICluster<? extends IInstance> dtest)
     static class IntegrationTestModule extends AbstractModule
     {
         private static final Logger LOGGER = LoggerFactory.getLogger(IntegrationTestModule.class);
-        private final ICluster<? extends IInstance> cluster;
+        private final Iterable<? extends IInstance> instances;
```
yifan-c, line 498: `ICluster` is already the `Iterable`. Why changing to the new type here? 
Beside, `ICluster` has convenient methods to return a stream of the instances.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteDownInstanceTest.java
```diff
@@ -0,0 +1,145 @@
...
+import static org.apache.cassandra.testing.TestUtils.TEST_KEYSPACE;
+
+/**
+ * Tests bulk writes in different scenarios when Cassandra instances are down
+ */
+class BulkWriteDownInstanceTest extends SharedClusterSparkIntegrationTestBase
+{
+    Set<IInstance> downInstances = new HashSet<>();
+
+    @ParameterizedTest(name = "{index} => instanceDownCount={0} {1}")
```
yifan-c: nit: so it outputs `instanceDownCount=0, readCL=ONE, writeCL=ALL`
```suggestion
    @ParameterizedTest(name = "{index} => instanceDownCount={0}, {1}")
```

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/replacement/HostReplacementMultiDCInsufficientReplicasTest.java
```diff
@@ -62,25 +61,10 @@ class HostReplacementMultiDCInsufficientReplicasTest extends HostReplacementTest
...
-        // Find the cause
-        while (cause != null && !StringUtils.contains(cause.getMessage(), "Failed to load"))
-        {
-            cause = cause.getCause();
-        }
-
-        assertThat(cause).isNotNull()
-                         .hasMessageFindingMatch("Failed to load (\\d+) ranges with EACH_QUORUM for " +
-                                                 "job ([a-zA-Z0-9-]+) in phase Environment Validation.");
+        sparkTestUtils.assertExpectedBulkWriteFailure(EACH_QUORUM.name(), dfWriter);
```
yifan-c, line 67: 👍

yifan-c (commented):
looks good overall

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
frankgh: This is so we can customize the number of instances handled by Sidecar like we do here: https://github.com/apache/cassandra-analytics/pull/59/files#diff-a23a3d0a797ed790cbf779edbb5f6892cb31acff2d2f9bdc9c10a3347411b6d9R103

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
frankgh: The main reason to use String here is convenience. At `org.apache.cassandra.sidecar.testing.SharedClusterIntegrationTestBase` we use type `org.apache.cassandra.distributed.api.ConsistencyLevel`, but at the call sites we have other types available. In this case, we have type `com.datastax.driver.core.ConsistencyLevel` from the `org.apache.cassandra.analytics.TestConsistencyLevel`. Alternatively we can try to consolidate into a single CL type that we use across all of the tests. WDYT?

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
frankgh: The goal is to allow a Sidecar service to handle a subset of instances from the cluster. For that we don't expose the cluster itself, but rather only give the Sidecar module access to the instances it has to handle.

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
yifan-c: yep. "consolidate into a single CL type" sounds good and makes it easier to maintain in the future.

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
yifan-c: Now I see

PR #60, opened 2024-06-17
CASSANDRA-19716: Invalid mapping when timestamp is used as a partitio…

frankgh:
…n key during bulk writes

Patch by Francisco Guerrero; Reviewed by Yifan Cai for CASSANDRA-19716


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/common/schema/TimestampType.java
```diff
@@ -0,0 +1,70 @@
...
+     * completion of the function, position will be moved by "length" bytes.
+     *
+     * @param buffer Buffer to parse column from
+     * @param length Serialized value size in buffer is as big as length
+     * @return value as {@link Date} type
+     */
+    @Override
+    public Date parseColumn(ByteBuffer buffer, int length)
+    {
+        assert length == Long.SIZE / Byte.SIZE;
```
yifan-c: nit: how about declare a constant for the division result instead? The value is used in `serialize` method too.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/SparkTestUtils.java
```diff
@@ -167,21 +173,27 @@ public SparkConf defaultSparkConf()
 
     public void validateWrites(List<Row> sourceData, Object[][] queriedData)
     {
+        validateWrites(sourceData, queriedData, null, Row::getInt);
```
yifan-c: can we pass the `DEFAULT_COLUMNS_MAPPER`, instead of `null`? and remove the null check in the method below?

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/PartitionKeyIntegrationTest.java
```diff
@@ -0,0 +1,172 @@
...
+/**
+ * Integration test that performs bulk reads and writes using different types for the partition key to test
+ * type mapping for the partition key
+ */
+class PartitionKeyIntegrationTest extends SharedClusterSparkIntegrationTestBase
+{
+    static final List<String> DATASET = Arrays.asList("a", "b", "c", "d", "e", "f", "g");
+    static final List<String> REDUCED_DATASET_FOR_BOOLEAN = Arrays.asList("a", "b");
+    static final String CREATE_TEST_TABLE_STATEMENT =
+    "CREATE TABLE IF NOT EXISTS %s (id %s, course text, marks int, PRIMARY KEY (id)) WITH read_repair='NONE';";
```
yifan-c, line 71: How about adding another test for the composite partition key case?

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/SparkTestUtils.java
frankgh: we pass `null` from the test class as well. I wanted to avoid passing it from the test class. But I guess we can disallow nulls here instead

yifan-c (approved):

PR #61, opened 2024-06-26
CASSANDRA-19727: Bulk writer fails validation stage when writing to a…

frankgh:
… cluster using RandomPartitioner

Patch by Francisco Guerrero; Reviewed by Yifan Cai for CASSANDRA-19727

### DIFF
--- a/cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/distributed/impl/CassandraCluster.java
+++ b/cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/distributed/impl/CassandraCluster.java
@@ -43,6 +43,7 @@ import org.apache.cassandra.distributed.api.IMessageSink;
 import org.apache.cassandra.distributed.api.TokenSupplier;
 import org.apache.cassandra.distributed.shared.ClusterUtils;
 import org.apache.cassandra.distributed.shared.Versions;
+import org.apache.cassandra.sidecar.adapters.base.Partitioner;
 import org.apache.cassandra.testing.ClusterBuilderConfiguration;
 import org.apache.cassandra.testing.IClusterExtension;
 import org.apache.cassandra.testing.TestTokenSupplier;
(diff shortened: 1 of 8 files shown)


cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/distributed/impl/CassandraCluster.java
```diff
@@ -43,6 +43,7 @@
 import org.apache.cassandra.distributed.api.TokenSupplier;
 import org.apache.cassandra.distributed.shared.ClusterUtils;
 import org.apache.cassandra.distributed.shared.Versions;
+import org.apache.cassandra.sidecar.adapters.base.Partitioner;
```
yifan-c, line 46: why does it depends on sidecar's partitioner? Can it implement with `org.apache.cassandra.spark.data.partitioner.Partitioner` instead?

It is unrelated with this patch. So I am fine w/o addressing this.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/SparkBulkAnalyticsSimpleTest.java
```diff
@@ -40,7 +40,7 @@
 import static org.apache.cassandra.testing.TestUtils.ROW_COUNT;
 import static org.apache.cassandra.testing.TestUtils.TEST_KEYSPACE;
 
-class SparkBulkWriterSimpleTest extends SharedClusterSparkIntegrationTestBase
+class SparkBulkAnalyticsSimpleTest extends SharedClusterSparkIntegrationTestBase
```
yifan-c: Since you are renaming the class.. would it be more fit to name it `CassandraAnalyticsSimpleTest`? 
"SparkBulkAnalytics" means "CassandraAnalytics" here.

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/distributed/impl/CassandraCluster.java
frankgh: ah yes, we don't have access to `org.apache.cassandra.spark.data.partitioner.Partitioner` here. I will not address it here, but it's something we need to address at some point

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/SparkBulkAnalyticsSimpleTest.java
frankgh: That was my first choice, so let me change it to `CassandraAnalyticsSimpleTest`

PR #62, opened 2024-07-03
[CASSANDRA-19748] Refactoring to introduce new cassandra-analytics-common module with m…

jberragan:
…inimal dependencies.

- Add new module cassandra-analytics-common with no dependencies on Spark or Cassandra and minimal standard dependencies (Guava, Jackson, Commons Lang Kryo etc)
- Move standalone classes to cassandra-analytics-common module.

Some additional refactoring and clean up:
- Rename SSTableInputStream -> BufferingInputStream
- Rename SSTableSource -> CassandraFileSource
- Introduce CassandraFile interface to be the implementing class for SSTable and CommitLog.
- Generalize IStats to work across different CassandraFile types
- Rename methods in StreamScanner to make the API clearer.
- Move ComplexTypeBuffer, ListBuffer, MapBuffer, SetBuffer, UdtBuffer to standalone classes
- Delete unused classes RangeTombstone, ReplicaSet and CollectionElement.


cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/partitioner/NotEnoughReplicasException.java
```diff
@@ -20,7 +20,7 @@
 package org.apache.cassandra.spark.data.partitioner;
 
 import java.math.BigInteger;
-import javax.validation.constraints.NotNull;
+import org.jetbrains.annotations.NotNull;
```
bbotella, line 23: Was this intended?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/reader/StreamScanner.java
```diff
@@ -60,7 +60,7 @@ public interface StreamScanner<Type> extends Closeable
      *
      * @return rid
```
bbotella: Update the comments?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/stats/IStats.java
```diff
@@ -0,0 +1,61 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cassandra.spark.stats;
+
+import org.apache.cassandra.spark.utils.streaming.CassandraFile;
+import org.apache.cassandra.spark.utils.streaming.CassandraFileSource;
+
+public interface IStats<FileType extends CassandraFile>
```
bbotella: It is pretty self explanatory, but I still think we can add some javadocs here.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/ComplexTypeBuffer.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.List;
+
+import org.apache.cassandra.db.marshal.AbstractType;
+import org.apache.cassandra.db.marshal.ListType;
+import org.apache.cassandra.db.marshal.MapType;
+import org.apache.cassandra.db.marshal.SetType;
+import org.apache.cassandra.db.marshal.UserType;
+import org.apache.cassandra.db.rows.Cell;
+
+public abstract class ComplexTypeBuffer
```
bbotella, line 37: Maybe we can take the opportunity to add some Javadocs now that the abstract class has its own file

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/partitioner/NotEnoughReplicasException.java
jberragan: Yes, we use `org.jetbrains.annotations.NotNull` elsewhere and `javax.validation.constraints.NotNull` is no longer depend on Jakata in the new common module.

bbotella:
LGTM (nb)

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/MapBuffer.java
```diff
@@ -17,28 +17,21 @@
...
+import org.apache.cassandra.db.rows.Cell;
 
-    public RangeTombstone(Bound open, Bound close)
+public class MapBuffer extends ComplexTypeBuffer
+{
+    MapBuffer(int cellCount)
     {
-        this.open = open;
-        this.close = close;
+        super(cellCount, cellCount * 2);
```
5, line 29: Can you please add a comment about why double cell count is used here?

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/ComplexTypeBuffer.java
```diff
@@ -0,0 +1,98 @@
...
+
+    void add(ByteBuffer buffer)
+    {
+        buffers.add(buffer);
+        length += buffer.remaining();
+    }
+
+    ByteBuffer build()
+    {
+        ByteBuffer result = ByteBuffer.allocate(4 + (buffers.size() * 4) + length);
```
5: Could you add a comment about why this size calculation makes sense?

Also, can that `4` be replaced with `Integer.BYTES` or something similar?

cassandra-bridge/src/main/java/org/apache/cassandra/spark/utils/FastThreadLocalUtf8Decoder.java
```diff
@@ -0,0 +1,62 @@
...
+        return ByteBufferUtils.string(buffer, FastThreadLocalUtf8Decoder.UTF8_DECODER::get);
+    }
+
+    public static String stringThrowRuntime(ByteBuffer buffer)
+    {
+        try
+        {
+            return string(buffer);
+        }
+        catch (final CharacterCodingException e)
```
5: Nit: I'd prefer we avoid single-symbol variable names for the sake of consistency and replace `e` with `exception`.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/TestUtils.java
```diff
@@ -167,7 +167,7 @@ static Dataset<Row> openLocalPartitionSizeSource(CassandraBridge bridge,
                                            .option("createStmt", createStmt)
                                            .option("dirs", dir.toAbsolutePath().toString())
                                            .option("version", version.toString())
-                                           .option("useSSTableInputStream", true) // use in the test system to test the SSTableInputStream
+                                           .option("useBufferingInputStream", true) // use in the test system to test the BufferingInputStream
```
5: Nit: I'd add and extra space and capitalization for the sake of consistency:
```java
.option("useBufferingInputStream", true)  // Use in the test system to test the BufferingInputStream
```

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/stats/IStats.java
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Generic Stats interface that works across all CassandraFile FileTypes.
+ *
+ * @param <FileType>
+ */
+public interface IStats<FileType extends CassandraFile>
+{
+    default void inputStreamEnd(CassandraFileSource<FileType> source, long runTimeNanos, long totalNanosBlocked)
+    {
+
```
5: Nit: No point in having extra empty lines within empty default methods here and below.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/ByteBufferUtils.java
```diff
@@ -30,20 +30,13 @@
 import java.nio.charset.CharsetDecoder;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
-
-import io.netty.util.concurrent.FastThreadLocal;
+import java.util.function.Supplier;
 
 public final class ByteBufferUtils
 {
+    public static final ThreadLocal<CharsetDecoder> UTF8_DECODER_PROVIDER = ThreadLocal.withInitial(StandardCharsets.UTF_8::newDecoder);
```
5: This line seems similar to what is going on in `FastThreadLocalUtf8Decoder`, can it be referenced and reused here?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/ByteBufferUtils.java
```diff
@@ -30,20 +30,13 @@
...
+    public static final int STATIC_MARKER = 0xFFFF;
```
5, line 39: Could you please add a comment about where this value comes from?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/ByteBufferUtils.java
```diff
@@ -93,25 +86,27 @@ public static byte[] getArray(ByteBuffer buffer)
...
-        {
-            return ByteBufferUtils.string(buffer);
-        }
-        catch (CharacterCodingException exception)
+        if (buffer.remaining() <= 0)
         {
-            throw new RuntimeException(exception);
+            return EMPTY_STRING;
         }
+        return decoderSupplier.get().decode(buffer.duplicate()).toString();
```
5: Nit: I personally would use a conditional operator here for improved readability:
```java
return buffer.remaining() > 0 ? decoderSupplier.get().decode(buffer.duplicate()).toString() : EMPTY_STRING;
```

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/ByteBufferUtils.java
```diff
@@ -222,4 +215,55 @@ public static void skipFully(InputStream is, long length) throws IOException
...
+
+    // Extract component position from buffer; return null if there are not enough components
+    public static ByteBuffer extractComponent(ByteBuffer buffer, int position)
+    {
+        buffer = buffer.duplicate();
+        readStatic(buffer);
+        int index = 0;
+        while (buffer.remaining() > 0)
+        {
+            ByteBuffer c = readBytesWithShortLength(buffer);
```
5: Nit: I would avoid single-symbol variable names and use something like `character` instead of `c`.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/ByteBufferUtils.java
```diff
@@ -222,4 +215,55 @@ public static void skipFully(InputStream is, long length) throws IOException
...
+            buffer.get();  // Skip end-of-component
+            ++index;
+        }
+        return null;
+    }
+
+    public static ByteBuffer[] split(ByteBuffer name, int numKeys)
+    {
+        // Assume all components, we'll trunk the array afterwards if need be, but most names will be complete
+        ByteBuffer[] l = new ByteBuffer[numKeys];
```
5: Nit: I would avoid single-symbol variable names like `l` (especially `l`, as it looks awfully like `1`).

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/streaming/CassandraFile.java
```diff
@@ -0,0 +1,24 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cassandra.spark.utils.streaming;
+
+public interface CassandraFile
```
5, line 25: Could you please add a short JavaDoc here explaining the purpose of this empty interface?

5 (approved):
Thank you for preparing these changes, James!

A few minor stylistic recommendations are below.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/MapBuffer.java
jberragan: Yes, it's because it requires 2 buffers, for the key & value. I'll add a comment.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/ByteBufferUtils.java
jberragan: We do not depend on netty in the `common` module, and we only depend transitively through Spark in the `core` module to use the `io.netty.util.concurrent.FastThreadLocal` class to provide the `CharsetDecoder`. 

This allows us to use `io.netty.util.concurrent.FastThreadLocal` when running in Spark, and fallback to `java.lang.ThreadLocal` if netty is not available.

cassandra-analytics-common/build.gradle
```diff
@@ -0,0 +1,54 @@
...
+            artifactId "${archivesBaseName}"
+            version System.getenv("CODE_VERSION") ?: "${version}"
+        }
+    }
+}
+
+dependencies {
+    implementation "org.slf4j:slf4j-api:${slf4jApiVersion}"
+    implementation "com.esotericsoftware:kryo-shaded:${kryoVersion}"
+    implementation "commons-lang:commons-lang:${commonsLangVersion}"
```
frankgh: we've avoided adding commons-lang dependencies because they tend to clash with spark provided commons-lang. I see it's only being used for hash builder and equality checking. This is easily done with native JDK APIs. Can you avoid adding this dependency?

cassandra-bridge/build.gradle
```diff
@@ -39,6 +39,7 @@ publishing {
 }
 
 dependencies {
+    implementation(project(':cassandra-analytics-common'))
```
frankgh: We should just declare this as an API implementation so we don't need to add both the bridge and common as dependencies in other subprojects (i.e four-zero)
```suggestion
    api(project(':cassandra-analytics-common'))
```

cassandra-four-zero-bridge/build.gradle
```diff
@@ -29,13 +29,15 @@ configurations {
 }
 
 dependencies {
+    compileOnly project(":cassandra-analytics-common")
```
frankgh: not needed if we declare it as api in the bridge. It will be brought as a transitive dep that can be used from the project

cassandra-four-zero-bridge/build.gradle
```diff
@@ -29,13 +29,15 @@ configurations {
...
     compileOnly(project(':cassandra-bridge'))
 
     compileOnly(group: "${sparkGroupId}", name: "spark-core_${scalaMajorVersion}", version: "${project.rootProject.sparkVersion}")
     compileOnly(group: "${sparkGroupId}", name: "spark-sql_${scalaMajorVersion}", version: "${project.rootProject.sparkVersion}")
 
     compileOnly(project(path: ':cassandra-four-zero', configuration: 'shadow'))
 
+    testImplementation project(":cassandra-analytics-common")
```
frankgh: also not needed if we make the change in bridge

gradle.properties
```diff
@@ -28,6 +28,10 @@ mockitoVersion=3.12.4
 jnaVersion=5.9.0
 scala=2.12
 spark=3
+kryoVersion=4.0.2
+commonsLangVersion=2.6
```
frankgh: can we remove this?

cassandra-analytics-common/build.gradle
```diff
@@ -0,0 +1,54 @@
...
+            groupId project.group
+            artifactId "${archivesBaseName}"
+            version System.getenv("CODE_VERSION") ?: "${version}"
+        }
+    }
+}
+
+dependencies {
+    implementation "org.slf4j:slf4j-api:${slf4jApiVersion}"
+    implementation "com.esotericsoftware:kryo-shaded:${kryoVersion}"
```
frankgh: hmm. this dependency used to be a transitive dependency through spark-core. Can we instead have it as a compileOnly dependency here instead?

cassandra-analytics-common/build.gradle
```diff
@@ -0,0 +1,54 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+plugins {
+    id('java-library')
+    id('maven-publish')
+}
+
+java {
```
frankgh, line 27: this gradle file is missing the test configuration section. You won't be able to run unit tests from some IDE environments for example without it. Also, test reporting will be missing for tests in this subproject. Please refer to other subprojects for reference

gradle.properties
```diff
@@ -28,6 +28,10 @@ mockitoVersion=3.12.4
...
+slf4jApiVersion=1.7.30
```
frankgh, line 32: nice! Can we update all places that use slf4j to use this variable ?

cassandra-analytics-common/build.gradle
```diff
@@ -0,0 +1,54 @@
...
+            version System.getenv("CODE_VERSION") ?: "${version}"
+        }
+    }
+}
+
+dependencies {
+    implementation "org.slf4j:slf4j-api:${slf4jApiVersion}"
+    implementation "com.esotericsoftware:kryo-shaded:${kryoVersion}"
+    implementation "commons-lang:commons-lang:${commonsLangVersion}"
+    implementation "com.google.guava:guava:${guavaVersion}"
```
frankgh: I think for guava we also need to bring it as compile only as it will resolve some version in the spark runtime environment

gradle.properties
```diff
@@ -28,6 +28,10 @@ mockitoVersion=3.12.4
...
+guavaVersion=16.0.1
```
frankgh, line 33: any reason we choose this version of guava? it seems old, and we have different versions in this project?

cassandra-analytics-common/build.gradle
frankgh: ah yes, it's coming transitively provided by the spark runtime environment. I'd move it to compile only to avoid problems

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/reader/EmptyStreamScanner.java
```diff
@@ -24,13 +24,13 @@ public class EmptyStreamScanner implements StreamScanner<Rid>
...
     @Override
-    public Rid rid()
+    public Rid data()
     {
         return null;
     }
 
     @Override
-    public boolean hasNext()
+    public boolean next()
```
frankgh, line 33: I think the method name `hasNext` makes more sense than `next` . Next would make sense if it returned the actual value. Any reason for changing it? Can we restore it?

cassandra-analytics-integration-tests/build.gradle
```diff
@@ -63,6 +63,7 @@ configurations {
 }
 
 dependencies {
+    testImplementation(project(':cassandra-analytics-common'))
```
frankgh: this dependency should come from core. No need to declare it here.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/reader/StreamScanner.java
```diff
@@ -49,30 +49,31 @@
  * <p>
  * Upon return from the next() call the current values of the scanner can be obtained by calling
  * the methods in Rid, getPartitionKey(), getColumnName(), getValue().
+ *
  * @param <Type> type of object returned by rid() method.
  */
 @SuppressWarnings("unused")
 public interface StreamScanner<Type> extends Closeable
```
aweisberg: Why doesn't this implement iterator so it can be used with all the iterator related things? Not really in scope for this refactor just curious.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/stats/IStats.java
```diff
@@ -0,0 +1,59 @@
...
+import org.apache.cassandra.spark.utils.streaming.CassandraFileSource;
+
+/**
+ * Generic Stats interface that works across all CassandraFile FileTypes.
+ *
+ * @param <FileType>
+ */
+public interface IStats<FileType extends CassandraFile>
+{
+    default void inputStreamEnd(CassandraFileSource<FileType> source, long runTimeNanos, long totalNanosBlocked)
```
aweisberg: If these are all going to be empty they could be single line with no line breaks between. Not really that important.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/ByteBufferUtils.java
```diff
@@ -222,4 +212,80 @@ public static void skipFully(InputStream is, long length) throws IOException
             throw new EOFException("EOF after " + skipped + " bytes out of " + length);
         }
     }
+
+    // Extract component position from buffer; return null if there are not enough components
+    public static ByteBuffer extractComponent(ByteBuffer buffer, int position)
```
aweisberg, line 223: These methods look like they are here to avoid depending on Spark. It would be helpful to comment where they originally came from.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/streaming/CassandraFile.java
aweisberg: Are there really no common methods? All the usages I can see are in `common` so not sure who would use this interface?

Seems like in practice the usage is always something concrete that extends `CassandraFile` but at that point the concrete class is doing all the heavy lifting.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/LocalDataLayer.java
```diff
@@ -174,7 +174,7 @@ public static LocalDataLayer from(Map<String, String> options)
                       .filter(StringUtils::isNotEmpty)
                       .collect(Collectors.toSet()),
                 SchemaFeatureSet.initializeFromOptions(options),
-                getBoolean(options, lowerCaseKey("useSSTableInputStream"), false),
+                getBoolean(options, lowerCaseKey("useBufferingInputStream"), false),
```
aweisberg: Is anything going to break because the name of this option changed? Should there be compatibility with the old name?

aweisberg (approved):
Some suggestions, but generally +1

I tried to build and test, but ran into unmappable characters in sidecar Javadoc.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/LocalDataLayer.java
jberragan: The `LocalDataLayer` is mainly for the tests to read/write SSTables, but I can maintain compatibility with the old name incase anyone is using it in the wild.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/streaming/CassandraFile.java
jberragan: This interface is somewhat in anticipation of Cdc and bringing in a `CommitLog` implementation, when that happens there will be some common  methods.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/reader/StreamScanner.java
jberragan: It is somewhat a legacy interface and it's mostly due to the way `AbstractStreamScanner` is implemented, it would be nice to move to a `java.util.Iterator` but nobody has grasped that nettle so far.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/ByteBufferUtils.java
frankgh: Any reason we need to make it public? I don't see it being used anywhere else.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/LocalDataLayer.java
```diff
@@ -174,7 +174,7 @@ public static LocalDataLayer from(Map<String, String> options)
                       .filter(StringUtils::isNotEmpty)
                       .collect(Collectors.toSet()),
                 SchemaFeatureSet.initializeFromOptions(options),
-                getBoolean(options, lowerCaseKey("useSSTableInputStream"), false),
+                getBoolean(options, lowerCaseKey("useBufferingInputStream"), getBoolean(options, lowerCaseKey("useSSTableInputStream"), false)),
```
frankgh, line 177: this is a behavior change. We should support the old property also for backwards compatibility

cassandra-bridge/src/main/java/org/apache/cassandra/spark/utils/FastThreadLocalUtf8Decoder.java
```diff
@@ -0,0 +1,62 @@
...
+        return ByteBufferUtils.string(buffer, FastThreadLocalUtf8Decoder.UTF8_DECODER::get);
+    }
+
+    public static String stringThrowRuntime(ByteBuffer buffer)
+    {
+        try
+        {
+            return string(buffer);
+        }
+        catch (final CharacterCodingException exception)
```
frankgh: NIT : avoid using final keyword here
```suggestion
        catch (CharacterCodingException exception)
```

frankgh (commented):
Overall looks good. But there are a few things that need to be addressed

cassandra-analytics-common/build.gradle
jberragan: ok

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/reader/EmptyStreamScanner.java
jberragan: `next()` actually causes the underlying scanner to move its position which is not normally the contract for `hasNext()` which is normally expected to be idempotent.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/LocalDataLayer.java
jberragan: The code here supports both the old and new prop.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/reader/EmptyStreamScanner.java
frankgh: ah ok, that makes sense now. I did not see a comment about that in the javadocs, can we add something to make it clear ?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/reader/StreamScanner.java
```diff
@@ -49,30 +49,31 @@
...
      *
-     * @return rid
+     * @return data of type {@code Type}
      */
-    Type rid();
+    Type data();
 
     /**
-     * Indicate if there are more data/rid avaiable
+     * Indicate if there are more data avaiable to read.
```
frankgh:
```suggestion
     * Indicate if there are more data available to read.
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/LocalDataLayer.java
frankgh: ah yes, I completely missed that. Thanks!

gradle.properties
jberragan: It only appears to be explicitly declared in the `sidecar-build` dependencies and it has its own variable `project.slf4jVersion`

gradle.properties
jberragan: `27.0.1-jre` appears to be the version used in `sidecar-build`, `16.0.1` is the version pulled in transitively from Spark. Moving to `27.0.1-jre` causes many of the tests to fail with `java.lang.NoSuchMethodError`

frankgh: I think the version is not that important now since we are bringing it as a compileONly dependency now

gradle.properties
frankgh: ah yeah, I think we are using it for tests too

jberragan:
Most recent CircleCI: https://app.circleci.com/pipelines/github/jberragan/cassandra-analytics/17/workflows/a63c72d8-79b7-4ab4-879a-221dd76fa3e7

frankgh (approved):
+1 thanks for addressing all the comments

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/reader/EmptyStreamScanner.java
```diff
@@ -24,13 +24,13 @@ public class EmptyStreamScanner implements StreamScanner<Rid>
     public static final EmptyStreamScanner INSTANCE = new EmptyStreamScanner();
 
     @Override
-    public Rid rid()
+    public Rid data()
```
yifan-c: Since you are renaming the method, can you rename the class to make it easier to understand? I'd propose `RowData`, instead of `Rid`

Is the method name `data()` clear enough to indicate that it returns a single element? In this code snippet `scanner.data()`, could it be read as all the data it has scanned? If so, it is clearer to call it `nextData`. 

I'd agree with @aweisberg that we should just move to the more common and well established interface, `Iterator`.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/partitioner/TokenPartitioner.java
```diff
@@ -51,7 +51,7 @@
 
 /**
  * Util class for partitioning Spark workers across the token ring
- * This class duplicates {@link org.apache.cassandra.spark.bulkwriter.TokenPartitioner} but is solely
+ * This class duplicates org.apache.cassandra.spark.bulkwriter.TokenPartitioner but is solely
```
yifan-c, line 54: The change is unrelated. Let's revert and make it clickable.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/reader/StreamScanner.java
```diff
@@ -49,30 +49,36 @@
  * <p>
  * Upon return from the next() call the current values of the scanner can be obtained by calling
  * the methods in Rid, getPartitionKey(), getColumnName(), getValue().
+ *
  * @param <Type> type of object returned by rid() method.
  */
 @SuppressWarnings("unused")
 public interface StreamScanner<Type> extends Closeable
```
yifan-c: Can you rename the type parameter from `Type` to `T` in this refactoring? I am strongly -1 on having the same name pattern as the class names. It is confusing when reading the code, mistaking the type parameter for the actual class types.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/stats/IStats.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.apache.cassandra.spark.utils.streaming.CassandraFile;
+import org.apache.cassandra.spark.utils.streaming.CassandraFileSource;
+
+/**
+ * Generic Stats interface that works across all CassandraFile FileTypes.
+ *
+ * @param <FileType>
+ */
+public interface IStats<FileType extends CassandraFile>
```
yifan-c: Can you rename the type parameter to just `F` or `FILE_TYPE` to be distinguishable with class names? (I would favor the former)

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/streaming/BufferingInputStream.java
```diff
@@ -30,29 +30,28 @@
...
  * buffering more bytes on-demand as the data is drained.
  * <p>
  * This class expects the consumer thread to be single-threaded, and the producer thread to be single-threaded OR serialized to ensure ordering of events.
  *
- * @param <T> SSTable
+ * @param <FileType> CassandraFile type e.g. SSTable, CommitLog
  */
 @SuppressWarnings({"WeakerAccess", "unused"})
-public class SSTableInputStream<T extends SSTable> extends InputStream implements StreamConsumer
+public class BufferingInputStream<FileType extends CassandraFile> extends InputStream implements StreamConsumer
```
yifan-c: Please either revert it back to `T` or use `F`. I think there are several other occurrences that the type parameter is named the same as class names. For the sake of readability, please update those and make the type parameter distinguishable.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/reader/StreamScanner.java
```diff
@@ -49,30 +49,36 @@
...
+     * so it should be called once and not called again until the {@code data} is consumed.
+     * </p>
      *
-     * @return true when the rid is available to be consumed;
+     * @return true when the data is available to be consumed;
      * otherwise, return false to indicate the scanner has exhausted
      * @throws IOException
      */
-    boolean hasNext() throws IOException;
+    boolean next() throws IOException;
```
yifan-c, line 78: The renaming does not make sense to me. 

As the updated doc states, it expects the consumer to always consume the next data. Meaning, it does not expect calling the `next()` method N times to skip N - 1 items. 

So, what we want is still just an `Iterator`. 

It is easier and less error prone by having the data advancing logic coded in the implementation, instead of the "implementation note". For example, the first call to `hasNext` advance to the next position, but the subsequent calls w/o calling `next()` does not advance and the state does not change.

yifan-c (commented):
comments/concerns on code style. I am -1 on some of the refactoring in the patch.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/partitioner/TokenPartitioner.java
jberragan: This class no longer depends on the core module and so the javadoc build fails.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/reader/StreamScanner.java
jberragan: Reworking `AbstreamStreamScanner` to be a `java.util.Iterator` is an involved change and too high risk to include in this PR. Contributions are welcome.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/stats/IStats.java
yifan-c: The current code style rule enforces brackets being placed in a new line. The rule has to be updated to allow inline brackets for empty blocks.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/reader/EmptyStreamScanner.java
jberragan: I suggest you take a deeper look at `AbstractStreamScanner` and the work involved in refactoring to an `java.util.Iterator`, `hasNext()` has the expectation of being idempotent and this is not currently the case for `AbstractStreamScanner`. I am +1 on moving to `java.util.Iterator` but it should be handled as an independent PR so it can be reviewed carefully and tested outside of these changes.

jberragan:
Latest CI: https://app.circleci.com/pipelines/github/jberragan/cassandra-analytics/25/workflows/e91b02d5-85eb-48bf-a5cb-bc71a942431f

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/reader/EmptyStreamScanner.java
yifan-c: Sounds good.

PR #63, opened 2024-07-13
CASSANDRA-19772: Deprecate option SIDECAR_INSTANCES and replace with SIDECAR_CONTACT_POINTS

yifan-c:
This patch introduces a new option SIDECAR_CONTACT_POINTS for both bulk writer and reader. The option name better describes the purpose, which is to specify the initial contact points to discover the cluster topology. The existing option SIDECAR_INSTANCES are used for the same purpose and it is now deprecated. In addition, it allows including the port value in the addresses when defining SIDECAR_CONTACT_POINTS

Patch by Yifan Cai; Reviewed by Francisco Guerrero for CASSANDRA-19772

### DIFF
--- a/cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/MapUtils.java
+++ b/cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/MapUtils.java
@@ -268,7 +268,7 @@ public final class MapUtils
             deprecatedOptionValue = resolver.apply(deprecated);
         }
 
-        if (options.containsKey(option))
+        if (options.containsKey(option) && deprecatedOptionValue != null)
         {
             LOGGER.info("The option: {} is defined. Favor the value over {}", option, deprecated);
             return resolver.apply(option);
--- a/cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/MapUtils.java
+++ b/cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/MapUtils.java
@@ -22,13 +22,19 @@ package org.apache.cassandra.spark.utils;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Function;
 import java.util.function.Supplier;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Utility methods for {@link Map}
  */
 public final class MapUtils
 {
+    private static final Logger LOGGER = LoggerFactory.getLogger(MapUtils.class);
+
     private MapUtils()
     {
         throw new IllegalStateException(getClass() + " is static utility class and shall not be instantiated");
@@ -238,5 +244,36 @@ public final class MapUtils
     public static boolean containsKey(Map<String, String> options, String key)
     {
         return options.containsKey(lowerCaseKey(key));
+    }
+
+    /**
+     * Resolve the value by consolidating the value associated with the current option and the deprecated option.
+     * The method assumes that both the new and the deprecated options expect the same value type.
+     * The value of the new option has the precedence over the deprecated one. Meaning when both options are present,
+     * the value associated with the new option is returned.
+     *
+     * @param options option keys and values map
+     * @param option new option
+     * @param deprecated deprecated option
+     * @param resolver function to resolve the value. If the input of the function is null, it asks for the default value
+     * @return resolved value
+     * @param <T> value type
+     */
+    public static <T> T resolveDeprecated(Map<String, String> options, String option, String deprecated, Function<String, T> resolver)
+    {
+        T deprecatedOptionValue = null;
+        if (options.containsKey(deprecated))
+        {
+            LOGGER.warn("The option: {} is deprecated. Please use {} instead. See option description for details.", deprecated, option);
+            deprecatedOptionValue = resolver.apply(deprecated);
+        }
+
+        if (options.containsKey(option))
+        {
+            LOGGER.info("The option: {} is defined. Favor the value over {}", option, deprecated);
+            return resolver.apply(option);
+        }
+
+        return deprecatedOptionValue == null ? resolver.apply(null) : deprecatedOptionValue;
     }
 }
--- a/cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/MapUtils.java
+++ b/cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/MapUtils.java
@@ -268,9 +268,12 @@ public final class MapUtils
             deprecatedOptionValue = resolver.apply(deprecated);
         }
 
-        if (options.containsKey(option) && deprecatedOptionValue != null)
+        if (options.containsKey(option))
         {
-            LOGGER.info("The option: {} is defined. Favor the value over {}", option, deprecated);
+            if (deprecatedOptionValue != null)
+            {
+                LOGGER.info("The option: {} is defined. Favor the value over {}", option, deprecated);
+            }
             return resolver.apply(option);
         }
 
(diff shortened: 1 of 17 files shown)


cassandra-analytics-core/src/test/spark3/org/apache/cassandra/spark/common/SidecarInstanceFactoryTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+        assertSidecarInstance(SidecarInstanceFactory.createFromString("localhost", 9999),
+                              "localhost", 9999);
+        assertSidecarInstance(SidecarInstanceFactory.createFromString("[2024:a::1]", 9999),
+                              "[2024:a::1]", 9999);
+        assertSidecarInstance(SidecarInstanceFactory.createFromString("localhost:8888", 9999),
+                              "localhost", 8888);
+        assertSidecarInstance(SidecarInstanceFactory.createFromString("127.0.0.1:8888", 9999),
+                              "127.0.0.1", 8888);
+        assertSidecarInstance(SidecarInstanceFactory.createFromString("[2024:a::1]:8888", 9999),
```
frankgh, line 46: I think we'll need to remove brackets from ipv6 addresses.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/MapUtils.java
```diff
@@ -239,4 +245,35 @@ public static boolean containsKey(Map<String, String> options, String key)
...
+        T deprecatedOptionValue = null;
+        if (options.containsKey(deprecated))
+        {
+            LOGGER.warn("The option: {} is deprecated. Please use {} instead. See option description for details.", deprecated, option);
+            deprecatedOptionValue = resolver.apply(deprecated);
+        }
+
+        if (options.containsKey(option))
+        {
+            LOGGER.info("The option: {} is defined. Favor the value over {}", option, deprecated);
```
frankgh: info level here? maybe condition it to `deprecatedOptionValue` not `null`?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -146,17 +148,16 @@ public class BulkSparkConf implements Serializable
     protected boolean useOpenSsl;
     protected int ringRetryCount;
     // create sidecarInstances from sidecarInstancesValue and effectiveSidecarPort
-    private final String sidecarInstancesValue;
+    private final String sidecarContactPointsCsv;
```
frankgh: Not a huge fan of this variable name, can we use `sidecarContactPointsValue` instead?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -240,38 +241,47 @@ protected Set<String> buildBlockedInstances(Map<String, String> options)
...
         return Arrays.stream(split)
-                     .map(hostname -> new SidecarInstanceImpl(hostname, effectiveSidecarPort))
+                     .filter(StringUtils::isNotEmpty)
+                     .map(hostname -> SidecarInstanceFactory.createFromString(hostname, effectiveSidecarPort))
                      .collect(Collectors.toSet());
     }
 
+    // extract port from input if present and use the extract port to create SidecarInstanceImpl
+    // otherwise, the input does not contain port info, and it uses the default port
+    static SidecarInstanceImpl createSidecarInstance(String input, int defaultPort)
```
frankgh: unused method?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/common/SidecarInstanceFactory.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Create SidecarInstance object by parsing the input string, which is IP address or hostname and optionally includes port
+     * @param input hostname string that can optionally includes the port. If port is present, the defaultPort param is ignored.
+     * @param defaultPort port value used when the input string contains no port
+     * @return SidecarInstanceImpl
+     */
+    public static SidecarInstanceImpl createFromString(String input, int defaultPort)
+    {
+        Preconditions.checkArgument(StringUtils.isNotEmpty(input), "Unable to create sidecar instance from empty input");
```
frankgh, line 47: I think HostAndPort from guava is an alternative option to use. It is included as a transitive dependency from spark-core. I think we need to make sure the API is compatible across all spark version we support

cassandra-analytics-core/src/test/spark3/org/apache/cassandra/spark/common/SidecarInstanceFactoryTest.java
yifan-c: Sidecar client constructor `URI` for requests. It asks for bracketed IPv6 address. 
> If a host is given then it is appended. If the host is a literal IPv6 address but is not enclosed in square brackets ('[' and ']') then the square brackets are added.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/MapUtils.java
yifan-c: good idea

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
yifan-c: forgot to merge after switch to `SidecarInstanceFactory`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
yifan-c: ok.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/common/SidecarInstanceFactory.java
yifan-c: Not sure if I follow. What is the benefit of creating a `HostAndPort` instance here? It cannot be used to construct `SidecarInstanceImpl`

yifan-c: Talked offline. We stick with the current implementation.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/MapUtils.java
```diff
@@ -239,4 +245,35 @@ public static boolean containsKey(Map<String, String> options, String key)
...
+        T deprecatedOptionValue = null;
+        if (options.containsKey(deprecated))
+        {
+            LOGGER.warn("The option: {} is deprecated. Please use {} instead. See option description for details.", deprecated, option);
+            deprecatedOptionValue = resolver.apply(deprecated);
+        }
+
+        if (options.containsKey(option) && deprecatedOptionValue != null)
+        {
+            LOGGER.info("The option: {} is defined. Favor the value over {}", option, deprecated);
```
frankgh: Wouldn't we be skipping this section altogether when the deprecated option is not provided?

```suggestion
        if (options.containsKey(option))
        {
            if (deprecatedOptionValue != null)
            {
                LOGGER.info("The option: {} is defined. Favor the value over {}", option, deprecated);
            }
```

frankgh (changes requested):

scripts/build-sidecar.sh
```diff
@@ -24,7 +24,7 @@ else
   SCRIPT_DIR=$( dirname -- "$( readlink -f -- "$0"; )"; )
   SIDECAR_REPO="${SIDECAR_REPO:-https://github.com/apache/cassandra-sidecar.git}"
   SIDECAR_BRANCH="${SIDECAR_BRANCH:-trunk}"
-  SIDECAR_COMMIT="${SIDECAR_COMMIT:-e95786a077e1137dcaae206854986987edc6a71e}"
+  SIDECAR_COMMIT="${SIDECAR_COMMIT:-55a9efee30555d3645680c6524043a6c9bc1194b}"
```
frankgh: rebase maybe?

frankgh (approved):
+1 Looks good to me

PR #67, opened 2024-07-22
CASSANDRA-19791: Remove other uses of Apache Commons lang for hashcode, equality and r…

jberragan:
…andom string generation

### DIFF
--- a/cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/RandomUtils.java
+++ b/cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/RandomUtils.java
@@ -24,15 +24,16 @@ import java.net.InetAddress;
 import java.nio.ByteBuffer;
 import java.util.Random;
 import java.util.UUID;
+import java.util.stream.IntStream;
 
 import com.google.common.net.InetAddresses;
 
-import org.apache.cassandra.spark.data.CqlField;
 import org.apache.cassandra.spark.data.partitioner.Partitioner;
 
 public final class RandomUtils
 {
-    private static final int MIN_COLLECTION_SIZE = 16;
+    private static final String ALPHANUMERIC = "ABCDEFGHIJKLMNOPQRSTUVWXYZ1234567890";
+    public static final int MIN_COLLECTION_SIZE = 16;
 
     public static final Random RANDOM = new Random();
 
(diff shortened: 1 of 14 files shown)


frankgh (approved):
+1 Looks good to me

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/RandomUtils.java
```diff
@@ -63,6 +64,24 @@ public static int randomPositiveInt(int bound)
         return RANDOM.nextInt(bound - 1) + 1;
     }
 
+    public static int nextInt(final int startInclusive, final int endExclusive)
```
yifan-c: please remove `final`

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/RandomUtils.java
```diff
@@ -63,6 +64,24 @@ public static int randomPositiveInt(int bound)
...
+        if (startInclusive < 0)
+        {
+            throw new IllegalArgumentException("Both range values must be non-negative.");
+        }
+        if (startInclusive == endExclusive)
+        {
+            return startInclusive;
+        }
+
+        return startInclusive + RANDOM.nextInt(endExclusive - startInclusive);
```
yifan-c, line 77: The bound value for random is exclusive. I think we are missing a value here. 

For example, when `endExclusive - startInclusive == 1`, the produce pseudo random values contains only a single value of 0, but it wants both 0 and 1 instead. 

The util is only used for testing, hence a minor issue. Let's get it correct still.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/RandomUtils.java
```diff
@@ -100,8 +119,18 @@ public static InetAddress randomInet()
...
+
+    public static String randomAlphanumeric(int maxLength)
     {
-        return type.randomValue(MIN_COLLECTION_SIZE);
+        StringBuilder sb = new StringBuilder(maxLength);
+        IntStream.rangeClosed(0, maxLength)
+                 .map(i -> RANDOM.nextInt(ALPHANUMERIC.length()))
+                 .mapToObj(ALPHANUMERIC::charAt)
+                 .forEach(sb::append);
+        return sb.toString();
```
yifan-c: The alternative in my mind is random from [0, 36), and map the lower 10 to numbers and upper 26 to chars. Therefore, you do not need to look up the ALPHANUMERIC string. 
It is just a nit, feel free to ignore.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/RandomUtils.java
jberragan: If it is end exclusive and start inclusive and the diff is 1, then shouldn't it only return 1 value? E.g. (4, 5] should always return 4?

yifan-c: Ah. Then the if block above should throw IllegalArgumentException too? e.g. [4, 4) is empty 

```java
        if (startInclusive == endExclusive)
        {
            return startInclusive;
        }
```

jberragan: It currently does this:

```
if (startInclusive == endExclusive)
{
    return startInclusive;
}
```

I don't particularly mind either way, current approach seems more permissive.

PR #71, opened 2024-08-05
CASSANDRA-19815: Decouple Cassandra types from Spark types so Cassandra types can be u…

jberragan:
…sed independently from Spark


cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlField.java
```diff
@@ -35,19 +35,16 @@
...
-import org.apache.spark.sql.catalyst.expressions.GenericInternalRow;
-import org.apache.spark.sql.types.DataType;
 import org.jetbrains.annotations.NotNull;
 
 @SuppressWarnings({ "WeakerAccess", "unused" })
 public class CqlField implements Serializable, Comparable<CqlField>
 {
     private static final long serialVersionUID = 42L;
 
+    private static final Comparator<String> STRING_COMPARATOR = String::compareTo;
```
yifan-c: Only `STRING_COMPARATOR` is `private`.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlField.java
```diff
@@ -201,6 +184,11 @@ public interface CqlCollection extends CqlType
         CqlField.CqlType type();
 
         CqlField.CqlType type(int position);
+
+        default boolean isComplex()
```
yifan-c, line 194: add javadoc

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlField.java
```diff
@@ -222,12 +210,17 @@ public interface CqlTuple extends CqlCollection
     {
         ByteBuffer serializeTuple(Object[] values);
 
-        Object[] deserializeTuple(ByteBuffer buffer, boolean isFrozen);
+        Object[] deserializeTuple(TypeConverter converter, ByteBuffer buffer, boolean isFrozen);
```
yifan-c: Please add javadoc for the API methods touched in this patch.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlField.java
```diff
@@ -330,14 +329,39 @@ public CqlType type()
...
+     * @param buffer    raw ByteBuffer
+     * @param isFrozen  true if the Cassandra type is frozen
+     * @return deserialized object converted to custom type.
+     */
+    public Object deserializeToType(TypeConverter converter, ByteBuffer buffer, boolean isFrozen)
+    {
+        return type().deserializeToType(converter, buffer, isFrozen);
+    }
+
+    public Object deserializeToJava(ByteBuffer buffer)
```
yifan-c: nit: `deserializeToJavaType`

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/TypeConverter.java
```diff
@@ -0,0 +1,40 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cassandra.spark.data;
+
+import org.jetbrains.annotations.NotNull;
+
+public interface TypeConverter
```
yifan-c, line 24: how about `SparkTypeConverter`?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/TypeConverter.java
```diff
@@ -0,0 +1,40 @@
...
+{
+    /**
+     * Converts Cassandra native value to desired equivalent.
```
yifan-c: not sure what does it mean with "Cassandra native value". Is the `Object value` indeed java type value?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlField.java
```diff
@@ -330,14 +329,39 @@ public CqlType type()
...
+
+    /**
+     * Deserialize raw ByteBuffer from Cassandra type and convert to a new type using the TypeConverter.
+     *
+     * @param converter custom TypeConverter that maps Cassandra type to some other type.
+     * @param buffer    raw ByteBuffer
+     * @param isFrozen  true if the Cassandra type is frozen
+     * @return deserialized object converted to custom type.
+     */
+    public Object deserializeToType(TypeConverter converter, ByteBuffer buffer, boolean isFrozen)
```
yifan-c, line 351: nit: `deserializeToSparkType` to better distinguish?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlField.java
```diff
@@ -473,6 +462,11 @@ public void write(Kryo kryo, Output output, CqlField field)
...
     public static UnsupportedOperationException notImplemented(CqlType type)
     {
-        return new UnsupportedOperationException(type.toString() + " type not implemented");
+        return notImplemented(type.toString());
+    }
+
+    public static UnsupportedOperationException notImplemented(String type)
+    {
+        return new UnsupportedOperationException(type + " type not implemented");
     }
```
yifan-c, line 477: The exception message is very much about type only. Can you make it a `private` method? To avoid being used in the other context.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/sparksql/SparkCellIterator.java
```diff
@@ -325,7 +329,10 @@ private void maybeRebuildClusteringKeys(@NotNull ByteBuffer columnNameBuf)
         {
             Object newObj = deserialize(field, ByteBufferUtils.extractComponent(columnNameBuf, index++));
             Object oldObj = values[field.position()];
-            if (newRow || oldObj == null || newObj == null || !field.equals(newObj, oldObj))
+            // Historically, we compare equality of clustering keys using the Spark types
+            // to determine if we have moved to a new 'row'. We could also compare using the Cassandra types
+            // or the raw ByteBuffers before converting to Spark types.
+            if (newRow || oldObj == null || newObj == null || !sparkSqlTypeConverter.toSparkType(field.type()).equals(newObj, oldObj))
```
yifan-c: Question: is the comment a todo (for moving to compare using Cassandra types/raw bytes)?

yifan-c (commented):
submit my review so far. I will continue the review in the other day.

yifan-c:
This flaky test `org.apache.cassandra.spark.EndToEndTests#testLargeBlobExclude` might be related. https://app.circleci.com/pipelines/github/yifan-c/cassandra-analytics/92/workflows/733d92a6-613e-4a00-ac3e-406111219d8a/jobs/853/tests

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/sparksql/SparkCellIterator.java
jberragan: Not necessarily, comparing with Spark types has worked fine for now. Comparing the raw bytes could be a bit faster. It is just an assumption that was intrinsic but not fully called out.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/TypeConverter.java
jberragan: `TypeConverter` is designed to be generic and de-coupled from Spark. `SparkSqlTypeConverter` extends `TypeConverter`.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlField.java
jberragan: `TypeConverter` is designed to be generic and not tied to Spark, so it could be some other type.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlField.java
jberragan: It is used in `NotImplementedTraits` and some tests which is why it is `public`, I can re-work if it's a blocker.

jberragan:
Green CI: https://app.circleci.com/pipelines/github/jberragan/cassandra-analytics/64/workflows/3eebf8bb-d4f2-43dc-ae0b-325b4772543e

jberragan:
Latest CircleCI: https://app.circleci.com/pipelines/github/jberragan/cassandra-analytics/71/workflows/9235a1e4-b44f-4c91-8691-ca92f70a9122

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/sparksql/SparkCellIterator.java
```diff
@@ -325,7 +329,10 @@ private void maybeRebuildClusteringKeys(@NotNull ByteBuffer columnNameBuf)
         {
             Object newObj = deserialize(field, ByteBufferUtils.extractComponent(columnNameBuf, index++));
             Object oldObj = values[field.position()];
-            if (newRow || oldObj == null || newObj == null || !field.equals(newObj, oldObj))
+            // Historically, we compare equality of clustering keys using the Spark types
+            // to determine if we have moved to a new 'row'. We could also compare using the Cassandra types
+            // or the raw ByteBuffers before converting to Spark types  - this mightb be slightly more performant.
+            if (newRow || oldObj == null || newObj == null || !sparkSqlTypeConverter.toSparkType(field.type()).equals(newObj, oldObj))
```
jberragan: TODO: convert CqlType to SparkType once in the constructor and re-use.

jberragan:
Latest CircleCI: https://app.circleci.com/pipelines/github/jberragan/cassandra-analytics/73/workflows/3f5e406a-98cd-44e0-9b2a-7001975742e6

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlField.java
yifan-c: Not a blocker. Just want to avoid any confusion. As long as they are used in the context of "type" only, it is good.

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/BinaryTraits.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.cassandra.bridge.BigNumberConfig;
+import org.apache.cassandra.spark.data.CqlField;
+import org.apache.cassandra.spark.utils.ByteBufferUtils;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.catalyst.expressions.GenericInternalRow;
+import org.apache.spark.sql.types.DataType;
+import org.apache.spark.sql.types.DataTypes;
+import org.jetbrains.annotations.NotNull;
+
+public interface BinaryTraits extends SparkType
```
yifan-c: I was confused by the name. At the first sight, I thought it was a `scala` source code, since `trait` is the scala's interface. 
Can you please not use the word `trait` for java interface? This patch add many traits.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/data/CqlFieldComparators.java
```diff
@@ -38,52 +39,52 @@
 public class CqlFieldComparators extends VersionRunner
 {
```
yifan-c: Test class not named with "Test"

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/reader/DataTypeSerializationTests.java
```diff
@@ -152,151 +153,153 @@ public void testUUID()
...
+            assertEquals(Decimal.apply(Long.MIN_VALUE), bridge.decimal().deserializeToType(bridge.typeConverter(),
+                                                                                           bridge.decimal().serialize(BigDecimal.valueOf(Long.MIN_VALUE))));
+            assertEquals(Decimal.apply(Integer.MAX_VALUE), bridge.decimal().deserializeToType(bridge.typeConverter(),
+                                                                                              bridge.decimal()
+                                                                                                    .serialize(BigDecimal.valueOf(Integer.MAX_VALUE))));
+            assertEquals(Decimal.apply(Integer.MIN_VALUE), bridge.decimal().deserializeToType(bridge.typeConverter(),
+                                                                                              bridge.decimal()
+                                                                                                    .serialize(BigDecimal.valueOf(Integer.MIN_VALUE))));
```
yifan-c: not easy to read. Can you extract a helper method?  For example, 

```
    Object toDecimal(CassandraBridge bridge, long value)
    {
        return bridge.decimal().deserializeToType(bridge.typeConverter(),
                                                  bridge.decimal().serialize(BigDecimal.valueOf(value)));
    }
```


If it looks good, can you update the other places with the same pattern?

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/SparkSqlTypeConverter.java
```diff
@@ -0,0 +1,73 @@
...
+    default Object toTestRowType(CqlField cqlField, Object value)
+    {
+        return toTestRowType(cqlField.type(), value);
+    }
+
+    default Object toTestRowType(CqlField.CqlType cqlType, Object value)
+    {
+        return toSparkType(cqlType).toTestRowType(value);
+    }
+}
```
yifan-c: javadoc for all methods and the interface.

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/Ascii.java
```diff
@@ -0,0 +1,30 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cassandra.spark.data.converter.types;
+
+public class Ascii implements StringTraits
+{
+    public static final org.apache.cassandra.spark.data.converter.types.Ascii INSTANCE = new org.apache.cassandra.spark.data.converter.types.Ascii();
```
yifan-c: The class qualifier is redundant. 

```suggestion
    public static final Ascii INSTANCE = new Ascii();
```

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/BigInt.java
```diff
@@ -0,0 +1,30 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cassandra.spark.data.converter.types;
+
+public class BigInt implements LongTraits
+{
+    public static final org.apache.cassandra.spark.data.converter.types.BigInt INSTANCE = new org.apache.cassandra.spark.data.converter.types.BigInt();
```
yifan-c: redundant class qualifier

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/Blob.java
```diff
@@ -0,0 +1,30 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cassandra.spark.data.converter.types;
+
+public class Blob implements BinaryTraits
+{
+    public static final org.apache.cassandra.spark.data.converter.types.Blob INSTANCE = new org.apache.cassandra.spark.data.converter.types.Blob();
```
yifan-c: and here, and many other classes in the package

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/Boolean.java
```diff
@@ -0,0 +1,60 @@
...
+package org.apache.cassandra.spark.data.converter.types;
+
+import org.apache.cassandra.bridge.BigNumberConfig;
+import org.apache.cassandra.spark.data.CqlField;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.catalyst.expressions.GenericInternalRow;
+import org.apache.spark.sql.types.DataType;
+import org.apache.spark.sql.types.DataTypes;
+
+public class Boolean implements SparkType
```
yifan-c: I would avoid naming it as `Boolean`. At the callsites, it only causes confusion with `java.lang.Boolean`

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/Double.java
```diff
@@ -0,0 +1,61 @@
...
+package org.apache.cassandra.spark.data.converter.types;
+
+import org.apache.cassandra.bridge.BigNumberConfig;
+import org.apache.cassandra.spark.data.CqlField;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.catalyst.expressions.GenericInternalRow;
+import org.apache.spark.sql.types.DataType;
+import org.apache.spark.sql.types.DataTypes;
+
+public class Double implements SparkType
```
yifan-c: Please rename the class.

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/Float.java
```diff
@@ -0,0 +1,61 @@
...
+package org.apache.cassandra.spark.data.converter.types;
+
+import org.apache.cassandra.bridge.BigNumberConfig;
+import org.apache.cassandra.spark.data.CqlField;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.catalyst.expressions.GenericInternalRow;
+import org.apache.spark.sql.types.DataType;
+import org.apache.spark.sql.types.DataTypes;
+
+public class Float implements SparkType
```
yifan-c: Rename this class to avoid conflict with `java.lang.Float`

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/Duration.java
```diff
@@ -0,0 +1,30 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cassandra.spark.data.converter.types;
+
+public class Duration implements NotImplementedTraits
```
yifan-c: It potentially causes confusion with `java.time.Duration`, but I am not strong on renaming it.

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/complex/List.java
```diff
@@ -0,0 +1,45 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cassandra.spark.data.converter.types.complex;
+
+import org.apache.cassandra.spark.data.CqlField;
+import org.apache.cassandra.spark.data.converter.SparkSqlTypeConverter;
+
+public class List implements ListTrait
```
yifan-c: Not a fan of the name. To easy to be confused with `java.util.List`. Same for "Map" and "Set"

yifan-c (changes requested):
Submit my review so far. Mostly on code style / naming

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/BinaryTraits.java
jberragan: I'll go ahead and remove it but I think trait or mixin is a common term not necessarily tied to any particular language

yifan-c: Thank you

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/reader/DataTypeSerializationTests.java
jberragan: I made the update here: https://github.com/apache/cassandra-analytics/pull/71/commits/b47f6fe3d4d04a1b77a5a6f2635c9d599a70c488

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/TypeConverter.java
```diff
@@ -0,0 +1,42 @@
...
+package org.apache.cassandra.spark.data;
+
+import org.jetbrains.annotations.NotNull;
+
+public interface TypeConverter
+{
+    default Object convert(CqlField.CqlType cqlType, @NotNull Object value)
+    {
+        return convert(cqlType, value, false);
+    }
```
yifan-c: This method is only used for test. Can we remove it?

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/complex/SparkTuple.java
```diff
@@ -0,0 +1,137 @@
...
+    public Object toSparkSqlType(Object value, boolean isFrozen)
+    {
+        if (value instanceof ByteBuffer)
+        {
+            // Need to deserialize first, e.g. if tuple is frozen inside collections
+            return toSparkSqlType(tuple.deserializeToJavaType((ByteBuffer) value), isFrozen);
+        }
+        else
+        {
+            Object[] array = (Object[]) value;
```
yifan-c, line 84: nit: How about checking the value type is indeed `Object[]`? And throw IllegalArgumentException, instead of cast error. Maybe there could be some util method to cast types with error message that provides better context on error.

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/complex/ListFeatures.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.catalyst.expressions.GenericInternalRow;
+import org.apache.spark.sql.catalyst.util.ArrayData;
+import org.apache.spark.sql.catalyst.util.GenericArrayData;
+import org.apache.spark.sql.types.DataType;
+import org.apache.spark.sql.types.DataTypes;
+import org.jetbrains.annotations.NotNull;
+import scala.collection.mutable.WrappedArray;
+
+interface ListFeatures extends CollectionFeatures
```
yifan-c: why interface? Would abstract class be a better fit?

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/complex/ListFeatures.java
```diff
@@ -0,0 +1,102 @@
...
+        return row.getList(position).stream()
+                  .map(element -> sparkType().toTestRowType(element))
+                  .collect(collector());
+    }
+
+    @SuppressWarnings({ "unchecked", "RedundantCast" }) // redundant cast to (Object[]) is required
+    @Override
+    default Object toTestRowType(Object value)
+    {
+        return Stream.of((Object[]) ((WrappedArray<Object>) value).array())
```
yifan-c: Can you add the explain on why the redundant cast is required? Since `array()` returns `Object[]` already.

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/complex/SparkUdt.java
```diff
@@ -0,0 +1,172 @@
...
+    private GenericInternalRow udtToSparkSqlType(Object value, boolean isFrozen)
+    {
+        if (value instanceof ByteBuffer)
+        {
+            // Need to deserialize first, e.g. if UDT is frozen inside collections
+            return udtToSparkSqlType(udt.deserializeUdt(converter, (ByteBuffer) value, isFrozen));
+        }
+        else
+        {
+            return udtToSparkSqlType((java.util.Map<String, Object>) value);
```
yifan-c: nit: add a comment to explain where the `Map<String, Object>` is coming from. (org.apache.cassandra.spark.data.complex.CqlUdt#deserializeUdt)

I do not think the full qualified class name is required for map.

yifan-c (commented):
Sorry that the review has taken a long time. I think those are my final comments.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/TypeConverter.java
jberragan: ok

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/complex/ListFeatures.java
jberragan: I can probably move these into `SparkList` as only `SparkList` implements (and `SparkSet` extends `SparkList`)

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/complex/ListFeatures.java
jberragan: `array()` returns `Object`

yifan-c: Interesting.. I can see that `def array: Array[T]` returns arrays..

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/sparksql/SparkCellIterator.java
```diff
@@ -325,7 +337,10 @@ private void maybeRebuildClusteringKeys(@NotNull ByteBuffer columnNameBuf)
         {
             Object newObj = deserialize(field, ByteBufferUtils.extractComponent(columnNameBuf, index++));
             Object oldObj = values[field.position()];
-            if (newRow || oldObj == null || newObj == null || !field.equals(newObj, oldObj))
+            // Historically, we compare equality of clustering keys using the Spark types
+            // to determine if we have moved to a new 'row'. We could also compare using the Cassandra types
+            // or the raw ByteBuffers before converting to Spark types  - this mightb be slightly more performant.
```
frankgh:
```suggestion
            // or the raw ByteBuffers before converting to Spark types  - this might be slightly more performant.
```

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/SparkDate.java
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.spark.sql.catalyst.expressions.GenericInternalRow;
+import org.apache.spark.sql.types.DataType;
+import org.apache.spark.sql.types.DataTypes;
+import org.jetbrains.annotations.NotNull;
+
+public class SparkDate implements SparkType
+{
+    public static final SparkDate INSTANCE = new SparkDate();
+
+    private SparkDate()
```
frankgh, line 34: can we annotate methods with override here?

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/SparkInet.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public Object toTestRowType(Object value)
+    {
+        if (value instanceof byte[])
+        {
+            try
+            {
+                return InetAddress.getByAddress((byte[]) value);
+            }
+            catch (final UnknownHostException e)
```
frankgh: NIT remove final modifier
```suggestion
            catch (UnknownHostException e)
```

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/complex/SparkMap.java
```diff
@@ -0,0 +1,140 @@
...
+    {
+        return mapToSparkSqlType((java.util.Map<Object, Object>) value, isFrozen);
+    }
+
+    private ArrayBasedMapData mapToSparkSqlType(final java.util.Map<Object, Object> map, boolean isFrozen)
+    {
+        final Object[] keys = new Object[map.size()];
+        final Object[] values = new Object[map.size()];
+        int position = 0;
+        for (final java.util.Map.Entry<Object, Object> entry : map.entrySet())
```
frankgh: NIT Remove final qualifiers
```suggestion
        Object[] keys = new Object[map.size()];
        Object[] values = new Object[map.size()];
        int position = 0;
        for (java.util.Map.Entry<Object, Object> entry : map.entrySet())
```

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/complex/SparkMap.java
```diff
@@ -0,0 +1,140 @@
...
+    public Object sparkSqlRowValue(GenericInternalRow row, int position)
+    {
+        final MapData map = row.getMap(position);
+        final ArrayData keys = map.keyArray();
+        final ArrayData values = map.valueArray();
+        final java.util.Map<Object, Object> result = new HashMap<>(keys.numElements());
+        for (int element = 0; element < keys.numElements(); element++)
+        {
+            final Object key = keyType().toTestRowType(keys.get(element, keyType().dataType()));
+            final Object value = valueType().toTestRowType(values.get(element, valueType().dataType()));
```
frankgh: Ditto
```suggestion
        MapData map = row.getMap(position);
        ArrayData keys = map.keyArray();
        ArrayData values = map.valueArray();
        java.util.Map<Object, Object> result = new HashMap<>(keys.numElements());
        for (int element = 0; element < keys.numElements(); element++)
        {
            Object key = keyType().toTestRowType(keys.get(element, keyType().dataType()));
            Object value = valueType().toTestRowType(values.get(element, valueType().dataType()));
```

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/complex/SparkTuple.java
```diff
@@ -0,0 +1,141 @@
...
+                array[index] = array[index] == null ? null : sparkType(index).toSparkSqlType(array[index], isFrozen);
+            }
+            return new GenericInternalRow(array);
+        }
+    }
+
+    @Override
+    public Object sparkSqlRowValue(GenericInternalRow row, int position)
+    {
+        final InternalRow tupleStruct = row.getStruct(position, size());
```
frankgh: NIT unnecessary final modifier
```suggestion
        InternalRow tupleStruct = row.getStruct(position, size());
```

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/complex/SparkTuple.java
```diff
@@ -0,0 +1,141 @@
...
+        return IntStream.range(0, size())
+                        .boxed()
+                        .map(index -> sparkType(index).toTestRowType(tupleStruct.get(index, sparkType(index).dataType())))
+                        .toArray();
+    }
+
+    @Override
+    public Object sparkSqlRowValue(Row row, int position)
+    {
+        final Row tupleStruct = row.getStruct(position);
```
frankgh: NIT unnecessary final modifier
```suggestion
        Row tupleStruct = row.getStruct(position);
```

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/complex/SparkTuple.java
```diff
@@ -0,0 +1,141 @@
...
+                        .filter(index -> !tupleStruct.isNullAt(index))
+                        .map(index -> sparkType(index).toTestRowType(tupleStruct.get(index)))
+                        .toArray();
+    }
+
+    @Override
+    public Object toTestRowType(Object value)
+    {
+        final GenericRowWithSchema tupleRow = (GenericRowWithSchema) value;
+        final Object[] tupleResult = new Object[tupleRow.size()];
```
frankgh: NIT unnecessary final modifier
```suggestion
        GenericRowWithSchema tupleRow = (GenericRowWithSchema) value;
        Object[] tupleResult = new Object[tupleRow.size()];
```

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/complex/SparkUdt.java
```diff
@@ -0,0 +1,179 @@
...
+                                                     converter.sparkSqlType(field.type(), bigNumberConfig),
+                                                     true))
+           .toArray(StructField[]::new)
+        );
+    }
+
+    @Override
+    public Object sparkSqlRowValue(GenericInternalRow row, int position)
+    {
+        final InternalRow struct = row.getStruct(position, size());
```
frankgh: NIT unnecessary final modifier
```suggestion
        InternalRow struct = row.getStruct(position, size());
```

cassandra-analytics-spark-converter/src/main/java/org/apache/cassandra/spark/data/converter/types/complex/SparkUdt.java
```diff
@@ -0,0 +1,179 @@
...
+                        .collect(Collectors.toMap(
+                        index -> field(index).name(),
+                        index -> sparkType(index).toTestRowType(struct.get(index, sparkType(index).dataType())
+                        )));
+    }
+
+    @Override
+    public Object sparkSqlRowValue(Row row, int position)
+    {
+        final Row struct = row.getStruct(position);
```
frankgh: NIT unnecessary final modifier
```suggestion
        Row struct = row.getStruct(position);
```

cassandra-bridge/build.gradle
```diff
@@ -40,6 +40,7 @@ publishing {
 
 dependencies {
     api(project(':cassandra-analytics-common'))
+    api(project(':cassandra-analytics-spark-converter'))
```
frankgh, line 43: does this belong here? it feels the bridge should be spark agnostic. We should be able to take the bridge and use it elsewhere without having to bring the spark bits

cassandra-four-zero-bridge/build.gradle
```diff
@@ -32,6 +32,7 @@ dependencies {
     compileOnly project(":cassandra-analytics-common")
     compileOnly project(":cassandra-four-zero-types")
     compileOnly(project(':cassandra-bridge'))
+    compileOnly project(":cassandra-analytics-spark-converter")
```
frankgh, line 35: same question here, it feels like spark dependency does not belong here

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
```diff
@@ -176,16 +176,18 @@ public CassandraBridgeImplementation()
...
 
-    @Override
     public CassandraTypes cassandraTypes()
     {
         return CassandraTypesImplementation.INSTANCE;
     }
 
+    public SparkSqlTypeConverter typeConverter()
+    {
+        return SparkSqlTypeConverterImplementation.INSTANCE;
```
frankgh, line 188: this should be a param to the bridge implementation . When Spark is needed we should have this spark instance here. but if we don't need spark, we should be able to use a different impl

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/data/converter/SparkSqlTypeConverterImplementation.java
```diff
@@ -0,0 +1,191 @@
...
+import org.apache.cassandra.spark.data.converter.types.complex.SparkUdt;
+import org.apache.spark.sql.Row;
+import org.apache.spark.sql.catalyst.expressions.GenericInternalRow;
+import org.apache.spark.sql.types.DataType;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * SparkSqlTypeConverter implementation that maps Cassandra 4.0 types to SparkSQL data types.
+ */
+public class SparkSqlTypeConverterImplementation implements SparkSqlTypeConverter
```
frankgh, line 67: this feels does not belong in this project. bridges should be agnostic to spark

frankgh (commented):
Looks good in general, but I have a few comments

cassandra-bridge/build.gradle
jberragan: I think bridge is the connecting point between Cassandra and Spark. It already depends on the `spark-core`, `spark_sql`  jars.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
jberragan: We could remove the Spark dependency from `cassandra-bridge` but it would be a lot more work and probably better to do outside this PR.

frankgh (approved):
+1 discussed offline about removing the spark dependency and decided that it can be done as part of a follow up

PR #72, opened 2024-08-09
CASSANDRA-19821: prevent double closing sstable writer

yifan-c:


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -184,6 +184,10 @@ public WriteResult write(Iterator<Tuple2<DecoratedKey, Object[]>> sourceIterator
             Range<BigInteger> currentRange = subRanges.get(currentRangeIndex);
             while (dataIterator.hasNext())
             {
+                if (streamSession != null)
+                {
+                    streamSession.throwIfLastStreamFailed();
```
frankgh, line 189: I feel this check belongs to the `maybeSwitchToNewStreamSession` method?

i.e.
```suggestion
        if (streamSession != null)
        {
            streamSession.throwIfLastStreamFailed();

            if (streamSession.getTokenRange().equals(currentRange))
            {
                return;
            }

            // Schedule data to be sent if we are processing a batch that has not been scheduled yet.
            // Complete existing writes (if any) before the existing stream session is closed
            flushAsync(taskContext.partitionId());
        }
```

frankgh (commented):
Looks good, I left one small comment

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
yifan-c: The `maybeSwitch` method would allow to get rid of the null check. However, I think it should only handle switching session. Throwing stream failure from the method may look confusing at the stack trace. Like, it is not error when switching session, most likely there is no session switching. 
I feel it is more clear to (re)throw the failure from RecordWriter#write.

frankgh: fair enough. My concern is doing the check on a tight loop for every iteration. I guess we could think of ways of just checking every x writes with the downside of doing extra work if an error occurred during the previous streaming session. My suggestion does not address the concern, it was more a style change.

frankgh: Discussed offline about the overhead being minimal. We are good on this change

frankgh (approved):
+1 Looks good to me

PR #73, opened 2024-08-12
CASSANDRA-19827: Add job_timeout_seconds writer option

yifan-c:
Option to specify the timeout in seconds for bulk write jobs. By default, it is disabled.
When JOB_TIMEOUT_SECONDS is specified, a job exceeding the timeout is:
- successful when the desired consistency level is met
- a failure otherwise

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
@@ -121,7 +121,7 @@ public class BulkSparkConf implements Serializable
     public final boolean skipExtendedVerify;
     public final WriteMode writeMode;
     public final int commitThreadsPerInstance;
-    public final int importCoordinatorTimeoutMultiplier;
+    public final double importCoordinatorTimeoutMultiplier;
     public boolean quoteIdentifiers;
     protected final String keystorePassword;
     protected final String keystorePath;
@@ -141,6 +141,7 @@ public class BulkSparkConf implements Serializable
     protected final StorageClientConfig storageClientConfig;
     protected final DataTransportInfo dataTransportInfo;
     protected final int jobKeepAliveMinutes;
+    protected final long jobTimeoutSeconds;
     // An optional unique identifier supplied by customer. The jobId is different from restoreJobId that is used internally.
     // The value is null when absent
     protected final String configuredJobId;
@@ -182,7 +183,7 @@ public class BulkSparkConf implements Serializable
         // else fall back to props, and then default if neither specified
         this.useOpenSsl = getBoolean(USE_OPENSSL, true);
         this.ringRetryCount = getInt(RING_RETRY_COUNT, DEFAULT_RING_RETRY_COUNT);
-        this.importCoordinatorTimeoutMultiplier = getInt(IMPORT_COORDINATOR_TIMEOUT_MULTIPLIER, 2);
+        this.importCoordinatorTimeoutMultiplier = getDouble(IMPORT_COORDINATOR_TIMEOUT_MULTIPLIER, 2.0);
         this.ttl = MapUtils.getOrDefault(options, WriterOptions.TTL.name(), null);
         this.timestamp = MapUtils.getOrDefault(options, WriterOptions.TIMESTAMP.name(), null);
         this.quoteIdentifiers = MapUtils.getBoolean(options, WriterOptions.QUOTE_IDENTIFIERS.name(), false, "quote identifiers");
@@ -216,6 +217,7 @@ public class BulkSparkConf implements Serializable
             throw new IllegalArgumentException(String.format("Invalid value for the '%s' Bulk Writer option (%d). It cannot be less than the minimum %s",
                                                              WriterOptions.JOB_KEEP_ALIVE_MINUTES, jobKeepAliveMinutes, MINIMUM_JOB_KEEP_ALIVE_MINUTES));
         }
+        this.jobTimeoutSeconds = MapUtils.getLong(options, WriterOptions.JOB_TIMEOUT_SECONDS.name(), -1L);
         this.configuredJobId = MapUtils.getOrDefault(options, WriterOptions.JOB_ID.name(), null);
 
         validateEnvironment();
@@ -497,6 +499,17 @@ public class BulkSparkConf implements Serializable
     public int getJobKeepAliveMinutes()
     {
         return jobKeepAliveMinutes;
+    }
+
+    public long getJobTimeoutSeconds()
+    {
+        return jobTimeoutSeconds;
+    }
+
+    protected double getDouble(String settingName, double defaultValue)
+    {
+        String finalSetting = getSettingNameOrDeprecatedName(settingName);
+        return conf.getDouble(finalSetting, defaultValue);
     }
 
     protected int getInt(String settingName, int defaultValue)
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
@@ -183,7 +183,7 @@ public class BulkSparkConf implements Serializable
         // else fall back to props, and then default if neither specified
         this.useOpenSsl = getBoolean(USE_OPENSSL, true);
         this.ringRetryCount = getInt(RING_RETRY_COUNT, DEFAULT_RING_RETRY_COUNT);
-        this.importCoordinatorTimeoutMultiplier = getDouble(IMPORT_COORDINATOR_TIMEOUT_MULTIPLIER, 2.0);
+        this.importCoordinatorTimeoutMultiplier = getDouble(IMPORT_COORDINATOR_TIMEOUT_MULTIPLIER, 0.5);
         this.ttl = MapUtils.getOrDefault(options, WriterOptions.TTL.name(), null);
         this.timestamp = MapUtils.getOrDefault(options, WriterOptions.TIMESTAMP.name(), null);
         this.quoteIdentifiers = MapUtils.getBoolean(options, WriterOptions.QUOTE_IDENTIFIERS.name(), false, "quote identifiers");
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
@@ -66,7 +66,8 @@ public class CassandraBulkSourceRelation extends BaseRelation implements Inserta
     private final JavaSparkContext sparkContext;
     private final Broadcast<BulkWriterContext> broadcastContext;
     private final BulkWriteValidator writeValidator;
-    private HeartbeatReporter heartbeatReporter;
+    private final SimpleTaskScheduler simpleTaskScheduler;
+    private volatile ImportCompletionCoordinator importCoordinator = null; // value is only set when using S3_COMPAT
     private long startTimeNanos;
 
     @SuppressWarnings("RedundantTypeArguments")
@@ -78,7 +79,7 @@ public class CassandraBulkSourceRelation extends BaseRelation implements Inserta
         this.broadcastContext = sparkContext.<BulkWriterContext>broadcast(writerContext);
         ReplicaAwareFailureHandler<RingInstance> failureHandler = new ReplicaAwareFailureHandler<>(writerContext.cluster().getPartitioner());
         this.writeValidator = new BulkWriteValidator(writerContext, failureHandler);
-        onCloudStorageTransport(ignored -> this.heartbeatReporter = new HeartbeatReporter());
+        this.simpleTaskScheduler = new SimpleTaskScheduler();
     }
 
     @Override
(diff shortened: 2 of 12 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/blobupload/StorageClient.java
```diff
@@ -216,7 +215,8 @@ private List<CompletedPart> uploadPartsOfBundle(String key, String uploadId,
             {
                 int loopPartNumber = partNumber;
                 ByteBuffer buffer = chunks.next();
-                AsyncRequestBody body = AsyncRequestBody.fromBytes(ByteBufferUtils.getArray(buffer));
+                // get a copy of the remaining ByteBuffer; for the last chunk, the cloned ByteBuffer is resized accordingly.
+                AsyncRequestBody body = AsyncRequestBody.fromRemainingByteBuffer(buffer);
```
yifan-c, line 219: `AsyncRequestBody.fromBytes(ByteBufferUtils.getArray(buffer))` copies data 2 times.. 
1. copy from buffer to byte array
2. copy from byte array to another byte array in `AsyncRequestBody.fromBytes`

By switching to `AsyncRequestBody.fromRemainingByteBuffer`, it copies once, i.e. reduce 1 data copy

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -182,7 +183,7 @@ public BulkSparkConf(SparkConf conf, Map<String, String> options)
         // else fall back to props, and then default if neither specified
         this.useOpenSsl = getBoolean(USE_OPENSSL, true);
         this.ringRetryCount = getInt(RING_RETRY_COUNT, DEFAULT_RING_RETRY_COUNT);
-        this.importCoordinatorTimeoutMultiplier = getInt(IMPORT_COORDINATOR_TIMEOUT_MULTIPLIER, 2);
+        this.importCoordinatorTimeoutMultiplier = getDouble(IMPORT_COORDINATOR_TIMEOUT_MULTIPLIER, 2.0);
```
JeetKunDoug: Given our experience with this setting and the fact that it's incredibly conservative, what are your thoughts on making it `0.5`... this is only used _after_ we've reached the desired consistency level and is a hedge against later repairs running, so it may be worth reducing.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -243,7 +244,7 @@ private void persist(@NotNull JavaPairRDD<DecoratedKey, Object[]> sortedRDD, Str
         {
             try
             {
-                onCloudStorageTransport(ignored -> heartbeatReporter.close());
+                onCloudStorageTransport(ignored -> simpleTaskScheduler.close());
```
JeetKunDoug: Should this still be called within `onCloudStorageTransport` now that we're always scheduling a timeout?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -434,4 +435,22 @@ private void abortRestoreJob(TransportContext.CloudStorageTransportContext conte
...
     }
+
+    private void maybeScheduleTimeout()
+    {
+        long timeoutSeconds = writerContext.job().jobTimeoutSeconds();
+        if (timeoutSeconds != -1)
+        {
+            long timeoutMillis = TimeUnit.SECONDS.toMillis(timeoutSeconds);
+            LOGGER.info("Scheduled job timeout. timeoutSeconds={}", timeoutSeconds);
+            simpleTaskScheduler.schedule("Job timeout", timeoutMillis, () -> {
```
JeetKunDoug: Suggestion - make `schedule` and `schedulePeriodic` take `Duration` in stead of a Long that is just named timeoutMillis.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/ImportCompletionCoordinator.java
```diff
@@ -165,6 +165,11 @@ public void waitForCompletion()
         }
     }
 
+    public boolean hasConsistencyLevelReached()
```
JeetKunDoug: maybe `hasReachedConsistencyLevel`?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/ImportCompletionCoordinator.java
```diff
@@ -219,28 +224,42 @@ private void addCompletionMonitor(CompletableFuture<?> future)
...
-                long timeToAllSatisfiedNanos = System.nanoTime() - waitStartNanos;
-                long timeout = estimateTimeout(timeToAllSatisfiedNanos);
+                long nowNanos = System.nanoTime();
+                long timeToAllSatisfiedNanos = nowNanos - waitStartNanos;
+                long elapsedNanos = nowNanos - startTimeNanos;
+                long timeoutNanos = estimateTimeoutNanos(timeToAllSatisfiedNanos, elapsedNanos,
+                                                         job.importCoordinatorTimeoutMultiplier(),
+                                                         minSliceSize, maxSliceSize,
+                                                         job.jobTimeoutSeconds());
                 LOGGER.info("The specified consistency level of the job has been satisfied. " +
```
JeetKunDoug: Realize this should probably be inside the `if (timeoutNanos > 0)` block, as otherwise we'll be logging some negative number of seconds here.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/ImportCompletionCoordinator.java
```diff
@@ -258,23 +277,41 @@ private void await()
...
+        if (jobTimeoutSeconds != -1)
+        {
+            long remainingIdealTimeoutNanos = TimeUnit.SECONDS.toNanos(jobTimeoutSeconds) - elapsedNanos;
+            if (remainingIdealTimeoutNanos <= 0)
+            {
+                // ideal timeout has passed, and we have already achieved the desired consistency level.
+                // Do not wait any longer
+                return 0;
+            }
+            timeoutNanos = Math.min(timeoutNanos, remainingIdealTimeoutNanos);
```
JeetKunDoug: I'd remove the usage of the term "ideal" here as we've essentially just called it "job timeout" everywhere else.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SimpleTaskScheduler.java
```diff
@@ -114,6 +123,25 @@ public synchronized void close()
         }
     }
 
+    private boolean isClosed()
+    {
+        if (isClosed)
+        {
+            LOGGER.info("SimpleTaskScheduler is already closed");
```
JeetKunDoug: Do we need this at `info` level for a boolean method that is _supposed_ to return if the scheduler is closed or not? Seems like, if it's necessary at all, it should be `debug`?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SimpleTaskScheduler.java
```diff
@@ -34,53 +34,62 @@
...
-        ScheduledFuture<?> fut = scheduler.scheduleWithFixedDelay(new NoThrow(name, heartBeat),
-                                                                  heartBeatIntervalMillis, // initial delay
-                                                                  heartBeatIntervalMillis, // delay
+
+        ScheduledFuture<?> fut = scheduler.scheduleWithFixedDelay(new NoThrow(name, task),
+                                                                  taskIntervalMillis, // initial delay
+                                                                  taskIntervalMillis, // delay
                                                                   TimeUnit.MILLISECONDS);
-        scheduledHeartbeats.put(name, fut);
+        scheduledTasks.put(name, fut);
```
JeetKunDoug, line 81: Should we be concerned about overwriting an existing scheduled task? Given the name is just a string, it'd be easy to potentially create multiple tasks with the same name, at which point only the last one would be in the scheduled task map, but others would be running and no longer reachable anywhere (so they can't be retrieved/cancelled later) ...

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SimpleTaskScheduler.java
```diff
@@ -34,53 +34,62 @@
...
-        ThreadFactory tf = ThreadUtil.threadFactory("Heartbeat reporter");
+        ThreadFactory tf = ThreadUtil.threadFactory("bulk-write-simple-task-scheduler");
         this.scheduler = Executors.newSingleThreadScheduledExecutor(tf);
-        this.scheduledHeartbeats = new HashMap<>();
+        this.scheduledTasks = new HashMap<>();
         this.isClosed = false;
     }
 
-    public synchronized void schedule(String name, long heartBeatIntervalMillis, Runnable heartBeat)
+    public synchronized void schedule(String name, long delayMillis, Runnable task)
```
JeetKunDoug: See comments above about replacing long delay values with Durations.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/ImportCompletionCoordinatorTest.java
```diff
@@ -237,6 +240,51 @@ void testAwaitShouldPassWithStuckSliceWhenClSatisfied()
                      "Each replica set should have a slice gets cancelled due to making no progress");
     }
 
+    @Test
+    void testAwaitShouldBlockUntilClSatisfiedWhenIdealTimeoutIsLow()
```
JeetKunDoug: Same comment about removing references to `Ideal`

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/ImportCompletionCoordinatorTest.java
```diff
@@ -237,6 +240,51 @@ void testAwaitShouldPassWithStuckSliceWhenClSatisfied()
...
+                     "Each replica set should have a slice gets cancelled due to making no progress");
+    }
+
+    @Test
+    void testAwaitShouldBlockUntilIdealTimeoutExceeds()
+    {
+        // it should produce a really large estimate; the coordinator ignores it anyway due to small ideal timeout
+        when(mockJobInfo.importCoordinatorTimeoutMultiplier()).thenReturn(1000.0);
+        long idealTimeout = 5L;
+        when(mockJobInfo.jobTimeoutSeconds()).thenReturn(idealTimeout); // await at most 10 seconds
```
JeetKunDoug: Comment (10 seconds) and value of `idealTimeout` (5 seconds) seem to not agree.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/SimpleTaskSchedulerTest.java
```diff
@@ -29,34 +29,34 @@
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
-public class HeartbeatReporterTest
+public class SimpleTaskSchedulerTest
 {
-    private HeartbeatReporter heartbeatReporter = new HeartbeatReporter();
+    private SimpleTaskScheduler simpleTaskScheduler = new SimpleTaskScheduler();
     private String heartbeatName = "test-heartbeat";
```
JeetKunDoug:
```suggestion
    private String taskName = "test-task";
```

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/SimpleTaskSchedulerTest.java
```diff
@@ -29,34 +29,34 @@
...
     @AfterEach
     public void teardown()
     {
-        heartbeatReporter.unschedule(heartbeatName);
+        simpleTaskScheduler.unschedule(heartbeatName);
     }
 
     @Test
-    public void testScheduleHeartbeat()
+    public void testSchedulePeriodicHeartbeat()
```
JeetKunDoug:
```suggestion
    public void testSchedulePeriodic()
```

JeetKunDoug (changes requested):
Thanks for working through this new timeout feature... looking good. I do have a few nits/suggestions, but nothing major.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
yifan-c: make perfect sense

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
yifan-c: good catch

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/ImportCompletionCoordinator.java
yifan-c: ok. I will break the message into 2 parts. The first is to confirm that CL has reached. The second is to log the wait time.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SimpleTaskScheduler.java
yifan-c: debug sounds correct. I added an info log message when it is actually closed.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SimpleTaskScheduler.java
yifan-c: scheduling with the same name would be rejected from the _synchronized_ methods, which do that check 
```
        if (isClosed() || isScheduled(name))
        {
            return;
        }
```

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/ImportCompletionCoordinatorTest.java
yifan-c: Deleted all "ideal" text

JeetKunDoug (commented):
+1 (nb) - Looks great! Thanks for the improvement.

PR #74, opened 2024-08-15
CASSANDRA-19836: Fix NPE when writing UDT values

yifan-c:
When UDT field values are set to null, the bulk writer throws NPE

Patch by Yifan Cai; Reviewed by TBD for CASSANDRA-19836

### DIFF
--- a/cassandra-four-zero-types/src/main/java/org/apache/cassandra/spark/data/CqlType.java
+++ b/cassandra-four-zero-types/src/main/java/org/apache/cassandra/spark/data/CqlType.java
@@ -30,6 +30,8 @@ import org.apache.cassandra.db.marshal.AbstractType;
 import org.apache.cassandra.serializers.TypeSerializer;
 import org.apache.spark.sql.Row;
 import org.apache.spark.sql.catalyst.expressions.GenericInternalRow;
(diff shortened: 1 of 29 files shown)


cassandra-four-zero-types/src/main/java/org/apache/cassandra/spark/data/CqlType.java
```diff
@@ -106,12 +108,35 @@ public Object sparkSqlRowValue(Row row, int position)
...
         throw CqlField.notImplemented(this);
     }
 
+    /**
+     * Set nullable inner value at the position for UDTs or Tuples
+     * @param udtValue udtValue to update
+     * @param position position in the vdtValue to set
+     * @param value nullable value to set
+     */
+    public final void setNullableInnerValue(SettableByIndexData<?> udtValue, int position, @Nullable Object value)
```
JeetKunDoug: NIT: I'd actually rename the original `setInnerValue` to `setInnerValueInternal` and keep this one called `setInnerValue`.

JeetKunDoug (approved):
Just the one NIT on naming public vs. internal implementation of the `setInnerValue` method, otherwise looks good. +1 (nb)

PR #75, opened 2024-08-22
CASSANDRA-19842: Consistency level check incorrectly passes when majo…

yifan-c:
…rity of the replica set is unavailable for write

Patch by Yifan Cai; Reviewed by TBD for CASSANDRA-19842

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
@@ -164,7 +164,13 @@ public class BulkSparkConf implements Serializable
         this.skipExtendedVerify = MapUtils.getBoolean(options, WriterOptions.SKIP_EXTENDED_VERIFY.name(), true,
                                                       "skip extended verification of SSTables by Cassandra");
         this.consistencyLevel = ConsistencyLevel.CL.valueOf(MapUtils.getOrDefault(options, WriterOptions.BULK_WRITER_CL.name(), "EACH_QUORUM"));
-        this.localDC = MapUtils.getOrDefault(options, WriterOptions.LOCAL_DC.name(), null);
+        String dc = MapUtils.getOrDefault(options, WriterOptions.LOCAL_DC.name(), null);
+        if (!consistencyLevel.isLocal() && dc != null)
+        {
+            LOGGER.warn("localDc is present for non-local consistency level {} specified in writer options. Correcting localDc to null", consistencyLevel);
+            dc = null;
+        }
+        this.localDC = dc;
         this.numberSplits = MapUtils.getInt(options, WriterOptions.NUMBER_SPLITS.name(), DEFAULT_NUM_SPLITS, "number of splits");
         this.sstableDataSizeInMiB = resolveSSTableDataSizeInMiB(options);
         this.commitBatchSize = MapUtils.getInt(options, WriterOptions.COMMIT_BATCH_SIZE.name(), DEFAULT_COMMIT_BATCH_SIZE, "commit batch size");
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriteValidator.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriteValidator.java
@@ -20,18 +20,17 @@
 package org.apache.cassandra.spark.bulkwriter;
 
 import java.math.BigInteger;
-import java.util.AbstractMap;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
-import com.google.common.collect.Multimap;
 import com.google.common.collect.Range;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.cassandra.spark.bulkwriter.token.ReplicaAwareFailureHandler;
 import org.apache.cassandra.spark.bulkwriter.token.TokenRangeMapping;
+import org.apache.cassandra.spark.exception.ConsistencyNotSatisfiedException;
 
 public class BulkWriteValidator
 {
@@ -57,8 +56,8 @@ public class BulkWriteValidator
                                         String phase,
                                         JobInfo job)
     {
-        Collection<AbstractMap.SimpleEntry<Range<BigInteger>, Multimap<RingInstance, String>>> failedRanges =
-        failureHandler.getFailedEntries(tokenRangeMapping, job.getConsistencyLevel(), job.getLocalDC());
+        List<ReplicaAwareFailureHandler<RingInstance>.ConsistencyFailurePerRange> failedRanges =
+        failureHandler.getFailedRanges(tokenRangeMapping, job.getConsistencyLevel(), job.getLocalDC());
 
         if (failedRanges.isEmpty())
         {
@@ -66,18 +65,11 @@ public class BulkWriteValidator
         }
         else
         {
-            String message = String.format("Failed to load %s ranges with %s for job %s in phase %s.",
+            String message = String.format("Failed to write %s ranges with %s for job %s in phase %s.",
                                            failedRanges.size(), job.getConsistencyLevel(), job.getId(), phase);
             logger.error(message);
-            failedRanges.forEach(failedRange ->
-                                 failedRange.getValue()
-                                            .keySet()
-                                            .forEach(instance ->
-                                                     logger.error("Failed {} for {} on {}",
-                                                                  phase,
-                                                                  failedRange.getKey(),
-                                                                  instance.toString())));
-            throw new RuntimeException(message);
+            logFailedRanges(logger, phase, failedRanges);
+            throw new ConsistencyNotSatisfiedException(message);
         }
     }
 
@@ -117,6 +109,21 @@ public class BulkWriteValidator
         });
     }
 
+    private static void logFailedRanges(Logger logger, String phase,
+                                        List<ReplicaAwareFailureHandler<RingInstance>.ConsistencyFailurePerRange> failedRanges)
+    {
+        for (ReplicaAwareFailureHandler<RingInstance>.ConsistencyFailurePerRange failedRange : failedRanges)
+        {
+            failedRange.failuresPerInstance.forEachInstance((instance, errors) -> {
+                logger.error("Failed in phase {} for {} on {}. Failure: {}",
+                             phase,
+                             failedRange.range,
+                             instance.toString(),
+                             errors);
+            });
+        }
+    }
+
     public void updateFailureHandler(Range<BigInteger> failedRange, RingInstance instance, String reason)
     {
         failureHandler.addFailure(failedRange, instance, reason);
@@ -124,8 +131,16 @@ public class BulkWriteValidator
 
     public void validateClOrFail(TokenRangeMapping<RingInstance> tokenRangeMapping)
     {
-        // Updates failures by looking up instance metadata
-        updateInstanceAvailability();
+        validateClOrFail(tokenRangeMapping, true);
+    }
+
+    public void validateClOrFail(TokenRangeMapping<RingInstance> tokenRangeMapping, boolean refreshInstanceAvailability)
+    {
+        if (refreshInstanceAvailability)
+        {
+            // Updates failures by looking up instance metadata
+            updateInstanceAvailability();
+        }
         // Fails if the failures violate consistency requirements
         validateClOrFail(tokenRangeMapping, failureHandler, LOGGER, phase, job);
     }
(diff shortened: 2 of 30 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -381,67 +383,79 @@ private InstanceAvailability determineInstanceAvailability(RingInstance instance
...
-        Set<RingInstance> blockedInstances;
+        Supplier<TokenRangeReplicasResponse> topologySupplier = () -> {
+            try
+            {
+                return getTokenRangesAndReplicaSets();
+            }
+            catch (ExecutionException | InterruptedException exception)
+            {
+                LOGGER.error("Failed to get token ranges, ", exception);
+                throw new RuntimeException(exception);
```
JeetKunDoug: I'd recommend moving the try/catch into `getTokenRangesAndReplicaSets` and throwing a new exception extending `AnalyticsException` here - that way, in the call itself, you can just pass `this::getTokenRangesAndReplicaSets` as the first parameter, which makes this feel a bit more consistent with the rest of the code.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
```diff
@@ -99,12 +99,10 @@ public boolean isLocal()
             public boolean checkConsistency(Set<String> writeReplicas,
                                             Set<String> pendingReplicas,
                                             Set<String> replacementInstances,
-                                            Set<String> blockedInstances,
```
frankgh, line 102: I'm glad to see the blocked instances being removed from this checks, but it seems like replacement instances is also problematic? we only seem to use it for CL.ALL and the usage seems incorrect too.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ReplicaAwareFailureHandler.java
```diff
@@ -157,55 +211,82 @@ private boolean validateConsistency(TokenRangeMapping<? extends CassandraInstanc
...
+
+        Set<String> writeReplicas = overlappingRanges.entrySet()
+                                                     .stream()
+                                                     .flatMap(e -> e.getValue().stream())
+                                                     .map(CassandraInstance::ipAddressWithPort)
+                                                     .collect(Collectors.toSet());
+        return filterOutReplicas(instances, writeReplicas);
+    }
+
+    private Set<String> filterOutReplicas(Set<String> source, Set<String> validSet)
```
frankgh: maybe detail what is happening here and why it's important

frankgh (commented):
I've added some comments. I think we may need to evaluate CL checks more closely

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
```diff
@@ -56,35 +56,38 @@ default void ensureNetworkTopologyStrategy(ReplicationFactor replicationFactor,
...
      */
-    boolean checkConsistency(Set<String> writeReplicas,
-                             Set<String> pendingReplicas,
-                             Set<String> replacementInstances,
-                             Set<String> blockedInstances,
-                             Set<String> failedInstanceIps,
-                             String localDC); // todo: simplify the parameter list. not all are required in impl
+    boolean checkWriteConsistency(Set<String> writeReplicas,
+                                  Set<String> pendingReplicas,
+                                  Set<String> failedReplicas);
```
JeetKunDoug: The documentation and method name here should very clearly state that it only checks consistency _for one DC_ and it is expected that multi-DC consistency levels like EACH_QUORUM/LOCAL_QUORUM will need to have the instances filtered properly before this method is called, and likely also call it multiple times to actually determine the success/failure... I'm not really crazy about this fact, as I think part of our problem with this class in general has been the use of primitives like Set<String>, String, etc. to pass information around where the types don't enforce any kind of discipline as to what the expected input is.  I'd recommend we use similar parameters to `canBeSatisfied` which much more clearly show what the input is, but with both succeeded and failed instances passed... in fact, it's possible that this could end up being 1 method, where you just pass an empty list for failed instances to do the `canBeSatisfied` check (although providing failed instances to that check would likely also be useful now that I think of it). 

More generally, it feels like the logic in ReplicaAwareFailureHandler that deals with the validating consistency (single DC, multi-DC, splitting failures into DC groups, etc.) really belongs in ConsistencyLevel.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
```diff
@@ -56,35 +56,38 @@ default void ensureNetworkTopologyStrategy(ReplicationFactor replicationFactor,
                                     cl.name() + " only make sense for NetworkTopologyStrategy keyspaces");
     }
 
+    default int quorum(Set<String> group)
```
JeetKunDoug: This is unused, as the static `quorum` method below is the only usage of `quorum` in the codebase.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriteValidator.java
```diff
@@ -117,6 +109,21 @@ public static void updateFailureHandler(CommitResult commitResult,
...
     }
 
+    private static void logFailedRanges(Logger logger, String phase,
+                                        List<ReplicaAwareFailureHandler<RingInstance>.ConsistencyFailurePerRange> failedRanges)
+    {
+        for (ReplicaAwareFailureHandler<RingInstance>.ConsistencyFailurePerRange failedRange : failedRanges)
+        {
+            for (RingInstance instance : failedRange.failuresPerInstance.instances())
+            {
+                logger.error("Failed in phase {} for {} on {}",
```
JeetKunDoug: I know this was just a refactoring of code, but logging the **reason** for failure here would be super-useful.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/StreamSessionConsistencyTest.java
```diff
@@ -82,34 +86,41 @@ public static Collection<Object[]> data()
...
+    private void setup(ConsistencyLevel.CL consistencyLevel)
     {
         digestAlgorithm = new XXHash32DigestAlgorithm();
         tableWriter = new MockTableWriter(folder);
         writerContext = new MockBulkWriterContext(TOKEN_RANGE_MAPPING, "cassandra-4.0.0", consistencyLevel);
         transportContext = (TransportContext.DirectDataBulkWriterContext) writerContext.transportContext();
     }
 
+    @Test
+    public void test() throws IOException, ExecutionException, InterruptedException
```
JeetKunDoug: Why does this test exist separate from the parameterized one? Can you please add both some documentation and a more descriptive method name than `test` here so if it fails it's understandable what it's testing and why it is different than the other tests?

JeetKunDoug (commented):
Partial review so far... my biggest concern is really about the split in logic between ReplicaAwareFailureHandler and ConsistencyLevel, which makes the code much more difficult to reason about.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
yifan-c: yeah. it is refactored and forgot to remove.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/StreamSessionConsistencyTest.java
yifan-c: It was there to debug a specific input set. Removed it.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/ConsistencyLevel.java
yifan-c: largely refactored the classes.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -164,7 +164,13 @@ public BulkSparkConf(SparkConf conf, Map<String, String> options)
         this.skipExtendedVerify = MapUtils.getBoolean(options, WriterOptions.SKIP_EXTENDED_VERIFY.name(), true,
                                                       "skip extended verification of SSTables by Cassandra");
         this.consistencyLevel = ConsistencyLevel.CL.valueOf(MapUtils.getOrDefault(options, WriterOptions.BULK_WRITER_CL.name(), "EACH_QUORUM"));
-        this.localDC = MapUtils.getOrDefault(options, WriterOptions.LOCAL_DC.name(), null);
+        String dc = MapUtils.getOrDefault(options, WriterOptions.LOCAL_DC.name(), null);
```
JeetKunDoug, line 167: 👍

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -381,67 +392,68 @@ private InstanceAvailability determineInstanceAvailability(RingInstance instance
...
                                        .stream()
+                                       .filter(m -> m.state().equalsIgnoreCase(InstanceState.REPLACING.name()))
                                        .map(RingInstance::new)
-                                       .filter(this::instanceIsBlocked)
                                        .collect(Collectors.toSet());
 
-            Set<String> blockedIps = blockedInstances.stream().map(i -> i.ringInstance().address())
-                                                     .collect(Collectors.toSet());
+        Set<RingInstance> blockedInstances;
+        blockedInstances = response.replicaMetadata()
```
JeetKunDoug: NIT: Since we're iterating over the entire `replicaMetadata()` collection and converting them all to `RingInstance` instances here, maybe move this up above the `replacementInstances` assignment and hold on to the complete list of instances? That way we only create them once, and can get rid of the `.values().stream()` duplication, and can filter on the instance state in the RingInstance rather than doing the `equalsIgnoreCase/.name` call in the filter of replacing instances?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -227,10 +230,18 @@ protected String getCurrentKeyspaceSchema() throws Exception
...
         CassandraContext context = getCassandraContext();
-        return context.getSidecarClient().tokenRangeReplicas(new ArrayList<>(context.getCluster()), conf.keyspace).get();
+        try
+        {
+            return context.getSidecarClient().tokenRangeReplicas(new ArrayList<>(context.getCluster()), conf.keyspace).get();
+        }
+        catch (ExecutionException | InterruptedException exception)
+        {
+            LOGGER.error("Failed to get token ranges", exception);
+            throw new SidecarApiCallException("Failed to get token ranges", exception);
```
frankgh: can we add the keyspace information here?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -478,21 +490,21 @@ private Map<String, Set<String>> getPendingReplicas(TokenRangeReplicasResponse r
...
+                                .filter(e -> e.getValue().stream() // todo: transformToHostWithoutPort is called twice for entries
                                               .anyMatch(v -> pendingReplicas.contains(transformToHostWithoutPort(v))))
                                 .collect(Collectors.toMap(Map.Entry::getKey,
                                                           e -> e.getValue().stream()
                                                                 .filter(v -> pendingReplicas.contains(transformToHostWithoutPort(v)))
                                                                 .collect(Collectors.toSet())));
     }
 
-    private String transformToHostWithoutPort(String v)
+    private static String transformToHostWithoutPort(String v)
```
frankgh, line 500: not ipv6 friendly?

frankgh (commented):
The PR looks good. I've added a couple of additional comments

JeetKunDoug (commented):
Comments so far. Will complete in the morning - I think in general things look good so if you and Francisco are comfortable with the changes (including the new tests we discussed off-line) feel free to move forward.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
yifan-c: talked offline. decide to not touch it as it is unrelated with the patch.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
yifan-c: done.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
yifan-c: done

frankgh (approved):
+1 No further comments from my end

JeetKunDoug (approved):
I think there are some additional testing improvements we can do, but this looks great. We can follow up with more test coverage when we look at further refactoring this code.

Thanks Yifan

PR #78, opened 2024-09-05
CASSANDRA-19901: Refactor TokenRangeMapping to use proper types instead of Strings

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSANDRA-19901


scripts/build-dtest-jars.sh
```diff
@@ -118,4 +118,6 @@ else
         exit ${RETURN}
     fi
   done
+  # always delete the Cassandra source after dtest.jar is built to avoid confusing IDE
```
frankgh, line 121: can we alternatively exclude these files from the IDE's indexing?

scripts/build-sidecar.sh
```diff
@@ -77,4 +77,6 @@ else
   fi
   git clean -fd
   ./gradlew -Pversion=${SIDECAR_BUILD_VERSION} -Dmaven.repo.local=${SIDECAR_JAR_DIR} publishToMavenLocal
+  # Delete sidecar source after publishing to avoid confusing IDE
+  rm -rf "${SIDECAR_BUILD_DIR}"
```
frankgh, line 81: can we alternatively exclude these files from the IDE's indexing?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/TokenRangeMapping.java
```diff
@@ -22,55 +22,131 @@
...
+
+        if (LOGGER.isDebugEnabled())
+        {
+            LOGGER.debug("Fetched token-ranges with dcs={}, write_replica_count={}, pending_replica_count={}",
+                         writeReplicasByDC.keySet(),
+                         writeReplicasByDC.values().stream().flatMap(Collection::stream).collect(Collectors.toSet()).size(),
+                         pendingReplicasByDC.values().stream().flatMap(Collection::stream).collect(Collectors.toSet()).size());
+        }
+
+        Map<String, ReplicaMetadata> replicaMetadata = response.replicaMetadata();
```
frankgh: unused?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/TokenRangeMapping.java
```diff
@@ -22,55 +22,131 @@
...
-    private final transient RangeMap<BigInteger, List<Instance>> replicasByTokenRange;
-    private final transient Multimap<Instance, Range<BigInteger>> tokenRangeMap;
-    private final transient Map<String, Set<String>> writeReplicasByDC;
-    private final transient Map<String, Set<String>> pendingReplicasByDC;
-    private final transient List<ReplicaMetadata> replicaMetadata;
+    private final transient Set<I> allInstances;
+    private final transient RangeMap<BigInteger, List<I>> replicasByTokenRange;
+    private final transient Multimap<I, Range<BigInteger>> tokenRangeMap;
+    private final transient Map<String, Set<I>> writeReplicasByDC;
+    private final transient Map<String, Set<I>> pendingReplicasByDC;
```
frankgh: I don't think it's worth keeping the replicas by DC, we always end up flattening. Why not just keep the set instead?

frankgh (commented):
Looks good in general, I've added some comments.

scripts/build-dtest-jars.sh
yifan-c: discussed offline. Source code is no longer needed once dtest.jar is built. We are good with file deletion.

frankgh (approved):
+1

PR #79, opened 2024-09-10
CASSANDRA-19909: Add writer options COORDINATED_WRITE_CONFIG to define coordinated write to multiple Cassandra clusters

yifan-c:
The option specifies the configuration (in JSON) for coordinated write. See org.apache.cassandra.spark.bulkwriter.coordinatedwrite.CoordinatedWriteConf. When the option is present, SIDECAR_CONTACT_POINTS, SIDECAR_INSTANCES and LOCAL_DC are ignored if they are present.

Patch by Yifan Cai; Reviewed by TBD for CASSANDRA-19909

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
@@ -42,6 +42,8 @@ import org.slf4j.LoggerFactory;
 
 import org.apache.cassandra.sidecar.client.SidecarInstance;
 import org.apache.cassandra.spark.bulkwriter.blobupload.StorageClientConfig;
+import org.apache.cassandra.spark.bulkwriter.coordinatedwrite.CoordinatedWriteConf;
+import org.apache.cassandra.spark.bulkwriter.coordinatedwrite.CoordinatedWriteConf.SimpleClusterConf;
 import org.apache.cassandra.spark.bulkwriter.token.ConsistencyLevel;
 import org.apache.cassandra.spark.bulkwriter.util.SbwKryoRegistrator;
 import org.apache.cassandra.spark.common.SidecarInstanceFactory;
@@ -148,7 +150,9 @@ public class BulkSparkConf implements Serializable
     protected int ringRetryCount;
     // create sidecarInstances from sidecarContactPointsValue and effectiveSidecarPort
     private final String sidecarContactPointsValue; // It takes comma separated values
-    private transient Set<? extends SidecarInstance> sidecarContactPoints; // not serialized
+    private transient Set<SidecarInstance> sidecarContactPoints; // not serialized
+    private final String coordinatedWriteConfJson;
+    private transient CoordinatedWriteConf coordinatedWriteConf; // it is transient; deserialized from coordinatedWriteConfJson in executors
 
     public BulkSparkConf(SparkConf conf, Map<String, String> options)
     {
@@ -223,7 +227,8 @@ public class BulkSparkConf implements Serializable
         }
         this.jobTimeoutSeconds = MapUtils.getLong(options, WriterOptions.JOB_TIMEOUT_SECONDS.name(), -1L);
         this.configuredJobId = MapUtils.getOrDefault(options, WriterOptions.JOB_ID.name(), null);
-
+        this.coordinatedWriteConfJson = MapUtils.getOrDefault(options, WriterOptions.COORDINATED_WRITE_CONF.name(), null);
+        this.coordinatedWriteConf = buildCoordinatedWriteConf(); // must only call the build method at this step, after the related fields have been resolved
         validateEnvironment();
     }
 
@@ -263,7 +268,7 @@ public class BulkSparkConf implements Serializable
         });
     }
 
-    protected Set<? extends SidecarInstance> buildSidecarContactPoints()
+    protected Set<SidecarInstance> buildSidecarContactPoints()
     {
         String[] split = Objects.requireNonNull(sidecarContactPointsValue, "Unable to build sidecar instances from null value")
                                 .split(",");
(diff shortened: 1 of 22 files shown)


cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/CoordinatedWriteConfTest.java
```diff
@@ -0,0 +1,111 @@
...
+import org.apache.cassandra.sidecar.client.SidecarInstance;
+import org.apache.cassandra.sidecar.client.SidecarInstanceImpl;
+import org.apache.cassandra.spark.bulkwriter.coordinatedwrite.CoordinatedWriteConf.ClusterConfProvider;
+import org.apache.cassandra.spark.bulkwriter.coordinatedwrite.CoordinatedWriteConf.SimpleClusterConf;
+import org.apache.cassandra.spark.bulkwriter.token.ConsistencyLevel.CL;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+class CoordinatedWriteConfTest
```
frankgh, line 40: maybe have a case where parsing fails. We should know how to tell the end user what's incorrectly configured in the config

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConfTest.java
```diff
@@ -275,6 +277,70 @@ void testSidecarContactPoints()
...
+    void testReadCoordinatedWriteConfFails()
+    {
+        Map<String, String> options = copyDefaultOptions();
+        String coordinatedWriteConfJsonNoLocalDc = "{\"cluster1\":" +
+                                                   "{\"sidecarContactPoints\":[\"instance-1:9999\",\"instance-2:9999\",\"instance-3:9999\"]}}";
+
+        options.put(WriterOptions.COORDINATED_WRITE_CONF.name(), coordinatedWriteConfJsonNoLocalDc);
+        assertThatThrownBy(() -> new BulkSparkConf(sparkConf, options))
+        .isExactlyInstanceOf(IllegalArgumentException.class)
+        .hasMessage("Coordinated write only supports S3_COMPACT mode");
```
frankgh:
```suggestion
        .hasMessage("Coordinated write only supports S3_COMPAT mode");
```

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConfTest.java
```diff
@@ -275,6 +277,70 @@ void testSidecarContactPoints()
         .isNull();
     }
 
+    @Test
+    void testReadCoordinatedWriteConfFails()
```
frankgh, line 281: Maybe let's have a test with an invalid JSON representation for the `WriterOptions.COORDINATED_WRITE_CONF` option?

frankgh (commented):
Some initial comments

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/CoordinatedWriteConfTest.java
yifan-c: I believe the test case is already there.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConfTest.java
yifan-c: I believe the test case is already there.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -223,7 +227,8 @@ public BulkSparkConf(SparkConf conf, Map<String, String> options)
         }
         this.jobTimeoutSeconds = MapUtils.getLong(options, WriterOptions.JOB_TIMEOUT_SECONDS.name(), -1L);
         this.configuredJobId = MapUtils.getOrDefault(options, WriterOptions.JOB_ID.name(), null);
-
+        this.coordinatedWriteConfJson = MapUtils.getOrDefault(options, WriterOptions.COORDINATED_WRITE_CONFIG.name(), null);
+        this.coordinatedWriteConf = buildCoordinatedWriteConf(); // must only call the build method at this step, after the related fields have been resolved
```
JeetKunDoug: If, instead, you passed the necessary parameters to `buildCoordinatedWriteConf` you could ensure the order was preserved, rather than just having a comment here. Move the build method to CoordinatedWriteConf as a static method and call it where necessary.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -282,6 +287,63 @@ Set<? extends SidecarInstance> sidecarContactPoints()
...
+        if (coordinatedWriteConf == null)
+        {
+            coordinatedWriteConf = buildCoordinatedWriteConf();
+        }
+
+        return coordinatedWriteConf;
+    }
+
+    @Nullable
+    protected CoordinatedWriteConf buildCoordinatedWriteConf()
```
JeetKunDoug: See above - this logic I think really belongs in CoordinatedWriteConf.build() or something.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/CoordinatedWriteConf.java
```diff
@@ -0,0 +1,177 @@
...
+                                                                "ClusterConfProvider value is not instance of " + clusterConfType));
+        return (Map<String, T>) clusters;
+    }
+
+    public String toJson() throws JsonProcessingException
+    {
+        return OBJECT_MAPPER.writeValueAsString(clusters);
+    }
+
+    public interface ClusterConfProvider
```
JeetKunDoug: `ClusterConfProvider` feels a lot like this should just be a functional interface that provides a ClusterConf object via a `get` method (similar to `Supplier`s). Isn't this just `ClusterConf`?

JeetKunDoug (commented):
Looking good - a few more comments.

PR #80, opened 2024-09-11
CASSANDRA-19910: Support data partitioning for multiple clusters coordinated write

yifan-c:


yifan-c:
I will rebase once https://github.com/apache/cassandra-analytics/pull/79 is merged

yifan-c:
> I will rebase once #79 is merged

#79 is merged. Rebase is done

cassandra-analytics-common/build.gradle
```diff
@@ -41,20 +41,21 @@ publishing {
...
-    testImplementation("org.junit.jupiter:junit-jupiter-params:${project.junitVersion}")
-    testImplementation("org.junit.jupiter:junit-jupiter-engine:${project.junitVersion}")
+    testImplementation("com.google.guava:guava:${guavaVersion}")
+    testImplementation("com.fasterxml.jackson.core:jackson-annotations:${jacksonVersion}")
+    testImplementation("com.fasterxml.jackson.core:jackson-core:${jacksonVersion}")
+    testImplementation("com.fasterxml.jackson.core:jackson-databind:${jacksonVersion}")
+    testImplementation("org.junit.jupiter:junit-jupiter-api:${junitVersion}")
+    testImplementation("org.junit.jupiter:junit-jupiter-params:${junitVersion}")
+    testImplementation("org.junit.jupiter:junit-jupiter-engine:${junitVersion}")
+    testImplementation("org.assertj:assertj-core:${assertjCoreVersion}")
```
yifan-c, line 58: the actual change is this line, adding assertj-core dependency. Other lines in the file are changed to unify the dependency declaration style.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlTable.java
```diff
@@ -39,7 +39,7 @@
 @SuppressWarnings({ "WeakerAccess", "unused" })
 public class CqlTable implements Serializable
 {
-    public static final long serialVersionUID = 42L;
+    private static final long serialVersionUID = 1018995207366817661L;
```
yifan-c, line 42: 42 is the ultimate answer, but likely a common one. Changed to a random value.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/token/TokenRangeMapping.java
```diff
@@ -113,12 +139,10 @@ Multimap<I, Range<BigInteger>> tokenRangesByInstance(List<ReplicaInfo> writeRepl
     }
 
     public TokenRangeMapping(Partitioner partitioner,
-                             ReplicationFactor replicationFactor,
```
yifan-c, line 116: Since `TokenRangeMapping` now can be consolidated, it no longer makes sense to contain ReplicationFactor, which is tie to cluster. 
The `TokenRangeMapping` or topology is only to describe the ring layout. `ReplicationFactor` is for consistency coordination. It is not part of `TokenRangeMapping` at the first place.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/token/TokenRangeMappingTest.java
```diff
@@ -105,16 +101,145 @@ void testCreateTokenRangeMappingWithPending()
         }
         TokenRangeMapping<RingInstance> topology = TokenRangeMapping.create(() -> response,
                                                                             () -> Partitioner.Murmur3Partitioner,
-                                                                            () -> rf, RingInstance::new);
+                                                                            RingInstance::new);
         assertThat(topology.pendingInstances()).hasSize(pendingCount);
     }
 
+    @Test
+    void testConsolidateListOfTokenRangeMappingFails()
```
JeetKunDoug, line 109: Is it possible to add some tests (or modify the replica building logic) to cover the entire token range of the partitioner so we have wrap-around ranges in these tests please?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/CassandraClusterInfoGroup.java
```diff
@@ -0,0 +1,278 @@
...
+    public Partitioner getPartitioner()
+    {
+        Map<String, Partitioner> aggregated = applyOnEach(ClusterInfo::getPartitioner);
+        Set<Partitioner> partitioners = EnumSet.copyOf(aggregated.values());
+        if (partitioners.size() != 1)
+        {
+            throw new IllegalStateException("Clusters are not running with the same partitioner kind. Found partitioners: " + aggregated);
+        }
+
+        return MapUtils.firstEntry(aggregated).getValue();
```
JeetKunDoug:
```suggestion
        return aggregated.get(clusterInfos.get(0).clusterId());
```

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/MapUtils.java
```diff
@@ -279,4 +282,19 @@ public static <T> T resolveDeprecated(Map<String, String> options, String option
...
         return deprecatedOptionValue == null ? resolver.apply(null) : deprecatedOptionValue;
     }
+
+    /**
+     * Get the first map entry from the map
+     *
+     * @return the first map entry, if there are at least one entry in the map
+     * @throws NoSuchElementException if the map is emtpy
+     */
+    public static <K, T> Map.Entry<K, T> firstEntry(@NotNull Map<K, T> map)
```
JeetKunDoug: This is somewhat misleading, as _most_ maps have no concept of order. What exactly is it trying to achieve, and is "first" really what we mean here (in which case you probably want to have the parameter to be a `SortedMap`, but from the usage in CassandraClusterInfoGroup it seems to be more like `anyEntry`, but I think you can just remove this all together and just do something like I recommend in the usage below.

Otherwise, I'd rename this `anyEntry`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/MultiClusterInfoProvider.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.List;
+
+import org.apache.cassandra.spark.bulkwriter.ClusterInfo;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Provider for multiple ClusterInfo and lookup
+ */
+public interface MultiClusterInfoProvider
```
JeetKunDoug, line 31: NIT: this is just `MultiClusterInfo` like the previous PR - no need for `Provider` in the interface name.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/CassandraClusterInfoGroup.java
```diff
@@ -0,0 +1,300 @@
...
+            TimeSkewResponse response = clusterInfo.getTimeSkew(instancesOfCluster);
+            long d = Math.abs(response.currentTime - localNow);
+            if (Math.abs(response.currentTime - localNow) > maxDiff)
+            {
+                maxDiff = d;
+                largestSkew = response;
+            }
+        }
+        return largestSkew;
+    }
```
JeetKunDoug: If we add an `allInstances` method to `clusterInfo` this becomes _much simpler_:
```suggestion
    public TimeSkewResponse getTimeSkew(List<RingInstance> instances)
    {
        return applyOnEach(clusterInfo ->
                           clusterInfo.getTimeSkew(clusterInfo.allInstances()))
               .values()
               .stream()
               .max(Comparator.comparingLong(tsr -> Math.abs(tsr.currentTime - System.currentTimeMillis())))
               .orElseThrow(() -> new RuntimeException("Could not load time skew data from any cluster."));
    }
```

I'd say the overhead of the `applyOnEach` is worth the code clarity by removing the size check at the beginning, but up to you.

`allInstances` is a one-liner in `ClusterInfo`:

```java
    public List<RingInstance> allInstances()
    {
        return new ArrayList<>(getTokenRangeReplicas().allInstances());
    }
```

and in `ClusterInfoGroup`:

``` java
    public List<RingInstance> allInstances()
    {
        return applyOnEach(ci -> ci.allInstances())
               .values().stream().flatMap(Collection::stream)
               .collect(Collectors.toUnmodifiableList());
    }
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/CassandraClusterInfoGroup.java
```diff
@@ -0,0 +1,300 @@
...
+import org.apache.cassandra.spark.bulkwriter.WriteAvailability;
+import org.apache.cassandra.spark.bulkwriter.RingInstance;
+import org.apache.cassandra.spark.bulkwriter.token.TokenRangeMapping;
+import org.apache.cassandra.spark.data.ReplicationFactor;
+import org.apache.cassandra.spark.data.partitioner.Partitioner;
+import org.apache.cassandra.spark.utils.MapUtils;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class CassandraClusterInfoGroup implements ClusterInfo, MultiClusterInfoProvider
```
JeetKunDoug, line 53: This class needs unit tests

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/sparksql/CassandraDataSink.java
```diff
@@ -73,7 +71,7 @@ public BaseRelation createRelation(@NotNull SQLContext sqlContext,
             case Append:
                 // Initialize the job group ID for later use if we need to cancel the job
                 // TODO: Can we get a more descriptive "description" in here from the end user somehow?
-                BulkWriterContext writerContext = createBulkWriterContext(
+                BulkWriterContext writerContext = factory().createBulkWriterContext(
```
JeetKunDoug, line 72: Realize it didn't happen in this PR, but the comments above this line has somehow never been kept with the code that they actually apply to - can you please move them back to the line where we actually call `setJobGroup` and remove the TODO at this point?

JeetKunDoug (commented):
Good stuff! I'm hoping we can add some additional tests around the coordinated write context, and a few other refactoring/quesitons.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/token/TokenRangeMappingTest.java
yifan-c: The `TokenRangeMapping` covers the entire ring and contains only the unwrapped ranges already.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/CassandraClusterInfoGroup.java
yifan-c: ok

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/MapUtils.java
yifan-c: ok. removing the code. It is not as useful.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/MultiClusterInfoProvider.java
yifan-c: I think `Provider` qualifies here. The interface is to provide the getters for clusters.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/CassandraClusterInfoGroup.java
yifan-c: The suggested implementation is not identical.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/CassandraClusterInfoGroup.java
yifan-c: Done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/CassandraClusterInfoGroup.java
yifan-c: I think it makes better sense to drop the parameter in the method. Let me know.

```
public TimeSkewResponse timeSkew()
```

yifan-c: I pushed https://github.com/apache/cassandra-analytics/pull/80/commits/45c8e0bebdfccd1a6071213556fbd986771a7a92 to address this comment

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/CassandraClusterInfoGroup.java
```diff
@@ -161,40 +160,22 @@ public void checkBulkWriterIsEnabledOrThrow()
...
-            if (Math.abs(response.currentTime - localNow) > maxDiff)
-            {
-                maxDiff = d;
-                largestSkew = response;
-            }
-        }
-        return largestSkew;
+        return clusterInfos.stream()
+                           .map(clusterInfo -> clusterInfo.timeSkew(range))
+                            // Find the timeSkew with the lowest remote currentTime, i.e. largest difference with the local current time.
```
JeetKunDoug: This isn't really right, as it would be possible for the current time on a remote host to be much _greater_ than the current time on the host running the task. We really need to do the whole "take the absolute value of the difference between the times" thing, not just look for the min currentTime.

I realize also that for multi-cluster scenarios, the other field in the TimeSkewResponse (`allowableSkewInMinutes`) can, in fact, be different (although in theory I'd doubt that it would be). We should really grab 2 things:
1) The time skew response with the maximum _difference_ between localNow and the remote time, where localNow is calculated each time you get a response back (per my previous suggestion)
2) The _minimum_ `allowableSkewInMinutes` value

And then craft the "combined" time skew response from those two things.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/CassandraClusterInfoGroupTest.java
```diff
@@ -0,0 +1,222 @@
...
+        CassandraClusterInfoGroup group = mockClusterGroup(2, index -> {
+            Map<RingInstance, WriteAvailability> availability = (index % 2 == 0) ? cluster1Availability : cluster2Availability;
+            CassandraClusterInfo clusterInfo = mockClusterInfo("cluster" + index);
+            when(clusterInfo.clusterWriteAvailability()).thenReturn(availability);
+            return clusterInfo;
+        });
+        assertThat(group.clusterWriteAvailability())
+        .describedAs("clusterWriteAvailability retrieved from group contains entries from both clusters")
+        .hasSize(2)
+        .containsValues(WriteAvailability.AVAILABLE, WriteAvailability.AVAILABLE);
```
JeetKunDoug: I always feel better when doing a test like this if I have 2 _different_ values that I'm combining, so maybe change one of the two clusters' ring instance to `UNAVAILABLE_DOWN` and change this assert?

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/CassandraClusterInfoGroupTest.java
```diff
@@ -0,0 +1,222 @@
...
+        .isExactlyInstanceOf(IllegalStateException.class)
+        .hasMessage("Cluster versions are not compatible. lowest=4.0.0 and highest=5.0.0");
+    }
+
+    @Test
+    void testCheckBulkWriterIsEnabledOrThrow()
+    {
+        CassandraClusterInfoGroup badGroup = mockClusterGroup(2, index -> {
+            CassandraClusterInfo clusterInfo = mockClusterInfo("cluster" + index);
+            if (index == 0)
```
JeetKunDoug: Similar to the "make sure I have different values", it would be good to test a non-zero index here (or randomize the index you use) - otherwise, aggregating this information and just calling the `checkBulkWriterIsEnabledOrThrow()` method would both pass the test.

JeetKunDoug (changes requested):
A few more comments but looking good. LMK when you've addressed these and I can take a final look and approve.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/CassandraClusterInfoGroup.java
yifan-c: Talked offline. We are going to refactor the method to do the validation, instead of exposing a response object.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/coordinatedwrite/CassandraClusterInfoGroupTest.java
yifan-c: yeah.. I think I had 2 state in my mind, but somehow they are all `AVAILABLE` in the written test. Maybe copy paste error. Good call.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfoTest.java
```diff
@@ -0,0 +1,102 @@
...
+    @Test
+    void testTimeSkewTooLarge()
+    {
+        Instant localNow = Instant.now();
+        int allowanceMinutes = 10;
+        Instant remoteNow = localNow.plus(Duration.ofMinutes(11)); // 11 > allowanceMinutes
+        CassandraClusterInfo ci = mockClusterInfoForTimeSkewTest(allowanceMinutes, remoteNow);
+        assertThatThrownBy(() -> ci.validateTimeSkewWithLocalNow(Range.openClosed(BigInteger.valueOf(10), BigInteger.valueOf(20)), localNow))
+        .isExactlyInstanceOf(TimeSkewTooLargeException.class);
+    }
```
JeetKunDoug: NIT: I'm not crazy about tests that have no asserts, even if it's really just "this doesn't throw" - a suggestion to make the test just a bit more explicit about what it's testing:

```suggestion
 
    public static final Range<BigInteger> RANGE = Range.openClosed(BigInteger.valueOf(10), BigInteger.valueOf(20));

    @Test
    void testTimeSkewAcceptable()
    {
        Instant localNow = Instant.now();
        int allowanceMinutes = 10;
        Instant remoteNow = localNow.plus(Duration.ofMinutes(1));
        CassandraClusterInfo ci = mockClusterInfoForTimeSkewTest(allowanceMinutes, remoteNow);
        assertThatNoException().describedAs("Acceptable time skew should validate without exception")
                               .isThrownBy(() -> ci.validateTimeSkewWithLocalNow(RANGE, localNow));
    }

    @Test
    void testTimeSkewTooLarge()
    {
        Instant localNow = Instant.now();
        int allowanceMinutes = 10;
        Instant remoteNow = localNow.plus(Duration.ofMinutes(11)); // 11 > allowanceMinutes
        CassandraClusterInfo ci = mockClusterInfoForTimeSkewTest(allowanceMinutes, remoteNow);
        assertThatException().describedAs("Time skew with too large a value should throw TimeSkewTooLargeException")
                             .isThrownBy(() -> ci.validateTimeSkewWithLocalNow(RANGE, localNow))
                             .isExactlyInstanceOf(TimeSkewTooLargeException.class);
    }
```

JeetKunDoug (approved):
Good stuff - thanks for working through all of my nits :)

PR #82, opened 2024-09-12
CASSANDRA-19917: Improve build-dependencies.sh script to set CASSANDRA_USE_JDK11 if an…

jberragan:
…alytics is running on JDK11


frankgh (commented):
Can we update the circleci scripts that use `CASSANDRA_USE_JDK11` to make sure it works out of the box?

PR #91, opened 2024-10-16
CASSANDRA-19994: Add dataTransferApi and TwoPhaseImportCoordinator for coordinated write

yifan-c:


cassandra-analytics-core-example/src/main/java/org/apache/cassandra/spark/example/ExampleStorageTransportExtension.java
```diff
@@ -145,17 +145,19 @@ private void refreshTokens()
...
                                                                 "writeSessionToken-" + tokenCount),
+                                         "readRegion",
                                          new StorageCredentials(
                                          "readAccessKeyId-" + tokenCount,
                                          "readSecretKey-" + tokenCount,
                                          "readSessionToken-" + tokenCount));
     }
 
     @Override
-    public void onStageSucceeded(String clusterId, long objectsCount, long rowsCount, long elapsedMillis)
```
JeetKunDoug, line 158: Did we lose object count all together, or did it move somewhere else?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/CassandraCloudStorageTransportContext.java
```diff
@@ -73,24 +76,28 @@ public CassandraCloudStorageTransportContext(@NotNull BulkWriterContext bulkWrit
         this.storageTransportConfiguration = storageTransportExtension.getStorageConfiguration();
         Objects.requireNonNull(storageTransportConfiguration,
                                "Storage configuration cannot be null in order to upload to cloud");
-        this.dataTransferApi = createBlobDataTransferApi();
+        this.storageClient = new StorageClient(storageTransportConfiguration, conf.getStorageClientConfig());
+        this.dataTransferApi = createBlobDataTransferApi(storageClient);
+        Preconditions.checkState(!jobInfo.isCoordinatedWriteEnabled()
+                                 || dataTransferApi instanceof CoordinatedCloudStorageDataTransferApi,
+                                 "CoordinatedCloudStorageDataTransferApi must be created when coordinated write is enabled");
```
JeetKunDoug: This chunk of code makes me believe that maybe the dataTransferApi should own creating the storage client, or even just own the storage client itself, so some external thing (the transport context in this case) doesn't have to verify that this precondition is met.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/CloudStorageDataTransferApiFactory.java
```diff
@@ -0,0 +1,71 @@
...
+public class CloudStorageDataTransferApiFactory
+{
+    public static final CloudStorageDataTransferApiFactory INSTANCE = new CloudStorageDataTransferApiFactory();
+
+    /**
+     * Create CloudStorageDataTransferApi based on the actual runtime type of ClusterInfo
+     * @return CoordinatedCloudStorageDataTransferApi if using coordinated write, i.e. with CassandraClusterInfoGroup;
+     *         otherwise, CloudStorageDataTransferApiImpl
+     */
+    public CloudStorageDataTransferApi createDataTransferApi(StorageClient storageClient,
```
JeetKunDoug, line 39: Could `ClusterInfo` instances not just do this on their own? I'm not crazy about places where we're doing if statements on the results of`instanceof` calls, as it seems you should be able to delegate this to the base class of the thing you're checking instanceof on.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterContainer.java
```diff
@@ -0,0 +1,185 @@
...
+ * @param <T> value type
+ */
+public class MultiClusterContainer<T> implements Serializable, MultiClusterSupport<T>
+{
+    private static final long serialVersionUID = 8387168256773834417L;
+
+    // used by coordinated write
+    private final Map<String, T> byCluster = new ConcurrentHashMap<>();
+    // used by non-coordinated write
+    private T single = null;
```
JeetKunDoug: Do we really need to have a special-case `single` here or could a non-coordinated write just be a write to a single cluster that has some static constant for the `clusterId` value?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/sparksql/CassandraDataSink.java
```diff
@@ -105,4 +102,11 @@ protected BulkWriterContextFactory factory()
     {
         return new BulkWriterContextFactory();
     }
+
+    // Util method to convert from Scala map to Java map. FQCN is used for code clarity.
```
JeetKunDoug: Why not just call the ScalaConversionUtils method directly?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/transports/storage/extensions/StorageTransportConfiguration.java
```diff
@@ -29,71 +27,93 @@
...
 
-    public String getReadBucket()
+    /**
+     * @param clusterId cluster id. Cluster id must present for coordinated write; otherwise, it is null
+     * @return a map of access tokens used to authenticate to the storage transport
+     */
+    public StorageCredentialPair getStorageCredentialPair(@Nullable String clusterId)
     {
-        return readBucket;
+        StorageAccessConfiguration readAccess = readAccessConfigurations.getValueOrNull(clusterId);
```
JeetKunDoug: Should we return `null` if there's no read access value? When would we possibly want this to not fail?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/transports/storage/extensions/StorageTransportConfiguration.java
```diff
@@ -29,71 +27,93 @@
...
-public class StorageTransportConfiguration implements Serializable
+public class StorageTransportConfiguration
 {
-    private static final long serialVersionUID = -8164878804296039585L;
-    private final String writeBucket;
-    private final String writeRegion;
-    private final String readBucket;
-    private final String readRegion;
     private final String prefix;
     private final Map<String, String> tags;
```
JeetKunDoug: Realize this is already in here, but can we name "tags" something slightly more descriptive like "objectTags" or "tagsToApply"...

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraJobInfo.java
```diff
@@ -50,6 +50,10 @@ public ConsistencyLevel getConsistencyLevel()
     @Override
     public String getLocalDC()
     {
+        if (isCoordinatedWriteEnabled())
+        {
+            throw new UnsupportedOperationException("Read localDc per cluster from coordinatedWriteConf, when coordinated write is enabled");
```
frankgh: I'm not understanding the message here.
```suggestion
            throw new UnsupportedOperationException("Remove the 'LOCAL_DC' writer option when using coordinated writes");
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CoordinatedCloudStorageDataTransferApi.java
```diff
@@ -0,0 +1,236 @@
...
+    @Override
+    public void forEach(BiConsumer<String, CloudStorageDataTransferApiImpl> action)
+    {
+        forEachInternal(ignored -> true, action);
+    }
+
+    @Override
+    public BundleStorageObject uploadBundle(StorageCredentials writeCredentials, Bundle bundle) throws S3ApiCallException
+    {
+        return dataTransferApis.getAnyValueOrThrow().uploadBundle(writeCredentials, bundle);
```
JeetKunDoug, line 100: I wonder if there aren't really two different interfaces in the dataTransferApi - on that is called on all of these `getAnyValueOrThrow` and a different one that's called for each cluster in the methods with a `forEach` in them? the `getAnyValue` thing seems weird, even though I vaguely understand why it's going on.

Not sure how exactly I'd split/refactor it but it may be worth looking at.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CoordinatedCloudStorageDataTransferApi.java
```diff
@@ -0,0 +1,236 @@
...
+            sidecarClient.executeRequestAsync(requestContext).get();
+        }
+        catch (Exception exception)
+        {
+            handleInterruption(exception);
+            throw new SidecarApiCallException("Failed to create restore job slice. " +
+                                              "restoreJobId=" + jobInfo.getRestoreJobId() +
+                                              " sliceId=" + createSliceRequestPayload.sliceId(),
+                                              exception);
+        }
```
JeetKunDoug, line 187: This method, minus the extra exception handling, is almost identical to the one in `CloudStorageDataTransferApiImpl` - could you not find a way to simply call that one from here, maybe adding the exception handling there so you could just call that one?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CoordinatedCloudStorageDataTransferApi.java
```diff
@@ -0,0 +1,236 @@
...
+        Predicate<String> allowAllClusters = id -> false;
+        forEachInternal(allowAllClusters, (ignored, api) -> dataTransferApiConsumer.accept(api));
+    }
+
+    // run dataTransferApiAction on the clusters that are not filtered by clusterIdFilter
+    private void forEachInternal(Predicate<String> clusterIdFilter,
+                                 BiConsumer<String, CloudStorageDataTransferApiImpl> dataTransferApiAction)
+    {
+        dataTransferApis.forEach((clusterId, dataTransferApi) -> {
+            if (clusterIdFilter.test(clusterId))
```
JeetKunDoug: Usually `filter`-type functions _include_ the items that _match_ the predicate, not the other way around. Can you swap this so that the predicate is used in a similar way to other filters like `Stream.filter`? In fact, there are two places where this is used, and each seems to think that it is a different behavior.

This one returns true for every one (which means it never runs):
```java
   @Override
    public void forEach(BiConsumer<String, CloudStorageDataTransferApiImpl> action)
    {
        forEachInternal(ignored -> true, action);
    }
```

This one returns false (so will run for all clusters):
```java
    private void forEachInternal(Consumer<CloudStorageDataTransferApiImpl> dataTransferApiConsumer)
    {
        Predicate<String> allowAllClusters = id -> false;
        forEachInternal(allowAllClusters, (ignored, api) -> dataTransferApiConsumer.accept(api));
    }
```

JeetKunDoug (commented):
Submitting comments so far. More to come.

Thanks for putting the effort into this - it's a big lift and obviously a lot of work!

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/transports/storage/extensions/StorageTransportConfiguration.java
frankgh: This method can return null, but we don't do null checks below when accessing the region/storageCredentials.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterContainer.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.function.BiConsumer;
+import java.util.function.UnaryOperator;
+
+import com.google.common.base.Preconditions;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A container to hold value per cluster. Each value is identified by its belonging cluster id.
```
frankgh:
```suggestion
 * A container to hold values per cluster. Each value is identified by its belonging cluster id.
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterContainer.java
```diff
@@ -0,0 +1,185 @@
...
+ * However, the container does not permit holding values for both multi-cluster case and single cluster case;
+ * otherwise, {@link IllegalStateException} is thrown.
+ * @param <T> value type
+ */
+public class MultiClusterContainer<T> implements Serializable, MultiClusterSupport<T>
+{
+    private static final long serialVersionUID = 8387168256773834417L;
+
+    // used by coordinated write
+    private final Map<String, T> byCluster = new ConcurrentHashMap<>();
```
frankgh: I think we should be able to avoid doing multiple null checks by having some sort of sentinel value that denotes single cluster.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterSupport.java
```diff
@@ -0,0 +1,69 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Support storing values per cluster, iteration and lookup
+ * @param <T> value type
+ */
+public interface MultiClusterSupport<T>
+{
+    /**
+     * @return count of all clusters
```
frankgh: NIT
```suggestion
     * @return the total number of clusters
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/transports/storage/extensions/CredentialChangeListener.java
```diff
@@ -36,5 +37,23 @@ public interface CredentialChangeListener
...
+     * for the Spark job to have time to rotate credentials without interruption.
+     * These tokens should be provided with plenty of time for the job to distribute them to
+     * the consumers of the storage transport endpoint to update their tokens before expiration.
+     *
+     * @param jobId     the unique identifier for the job
+     * @param clusterId the unique identifier for the cluster.
+     *                  When its value is null, it behaves the same as {@link #onCredentialsChanged(String, StorageCredentialPair)}
+     * @param newTokens a map of access tokens used to authenticate to the storage transport
+     */
+    void onCredentialsChanged(String jobId, @Nullable String clusterId, StorageCredentialPair newTokens);
```
frankgh, line 58: I am not seeing any call sites for this method

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/sparksql/CassandraDataSink.java
```diff
@@ -105,4 +102,11 @@ protected BulkWriterContextFactory factory()
...
+    private static java.util.Map<String, String> toJavaMap(scala.collection.immutable.Map<String, String> map)
+    {
+        // preserve the type arguments as required by jdk 1.8
+        return ScalaConversionUtils.<String, String>mapAsJavaMap(map);
```
frankgh: any reason why we no longer inline this? was it causing issues?

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/CassandraAnalyticsSimpleTest.java
```diff
@@ -45,12 +47,17 @@
...
-    void runSampleJob(Integer ttl, Long timestamp)
+    void runSampleJob(Integer ttl, Long timestamp, QualifiedName tableName)
     {
+        createTestTable(tableName, CREATE_TEST_TABLE_STATEMENT);
+
+        // The table is created in cassandra. However, sidecar keyspace endpoint relies on java driver to refresh schema,
+        // which introduces certain delay (1 second refresh interval).
+        // Wait for a while here to ensure sidecar is aware of schema.
+        Uninterruptibles.sleepUninterruptibly(2, TimeUnit.SECONDS);
+
```
frankgh: That's why we have `initializeSchemaForTest` which runs before sidecar starts. When sidecar starts, we will have all the schemas ready. Ideally we should pre-create all the schemas we need before actually running any test

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterContainerTest.java
```diff
@@ -0,0 +1,161 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cassandra.spark.bulkwriter.cloudstorage.coordinated;
+
+import java.util.Arrays;
+
+import com.google.common.collect.ImmutableMap;
```
frankgh: I'd prefer to avoid using google common's where realistically possible
```suggestion
```

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterContainerTest.java
```diff
@@ -0,0 +1,161 @@
...
+        container.setValue(null, new Value());
+        assertThat(container.getValueOrNull(null)).isNotNull();
+        assertThat(container.getValueOrNull("cluster1")).isNull();
+    }
+
+    @Test
+    void testReadAndWriteForMultiClusters()
+    {
+        MultiClusterContainer<Value> container = new MultiClusterContainer<>();
+        container.addAll(ImmutableMap.of("cluster1", new Value()));
```
frankgh:
```suggestion
        container.addAll(Collections.singletonMap("cluster1", new Value()));
```

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterContainerTest.java
```diff
@@ -0,0 +1,161 @@
...
+    void testSetValueFailsDueToConflict()
+    {
+        MultiClusterContainer<Value> singleContainer = new MultiClusterContainer<>();
+        singleContainer.setValue(null, new Value());
+
+        assertThatThrownBy(() -> singleContainer.setValue("cluster1", new Value()))
+        .isExactlyInstanceOf(IllegalStateException.class)
+        .hasMessage("Cannot set value for non-null cluster when the container is used for non-coordinated-write");
+
+        assertThatThrownBy(() -> singleContainer.addAll(ImmutableMap.of("cluster1", new Value())))
```
frankgh:
```suggestion
        assertThatThrownBy(() -> singleContainer.addAll(Collections.singletonMap("cluster1", new Value())))
```

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterContainerTest.java
```diff
@@ -0,0 +1,161 @@
...
+        .isExactlyInstanceOf(IllegalStateException.class)
+        .hasMessage("Cannot set value for non-null cluster when the container is used for non-coordinated-write");
+
+        MultiClusterContainer<Value> multiClusterContainer = new MultiClusterContainer<>();
+        multiClusterContainer.addAll(ImmutableMap.of("cluster1", new Value()));
```
frankgh:
```suggestion
        multiClusterContainer.addAll(Collections.singletonMap("cluster1", new Value()));
```

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterContainerTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+import java.util.Arrays;
+
+import com.google.common.collect.ImmutableMap;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+class MultiClusterContainerTest
```
frankgh, line 32: missing unit tests for `getAnyValueOrThrow` and `forEach` should be easy to test and increase coverage?

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterContainerTest.java
```diff
@@ -0,0 +1,161 @@
...
+        assertThat(container.size()).isZero();
+
+        container.setValue("cluster1", new Value());
+        assertThat(container.size()).isOne();
+
+        container.setValue("cluster2", new Value());
+        assertThat(container.size()).isEqualTo(2);
+
+        container = new MultiClusterContainer<>();
+        container.setValue(null, new Value());
```
frankgh, line 161: small nit, maybe set the value twice to ensure we always have one entry when we update the null key twice?
```suggestion
        container.setValue(null, new Value());
        container.setValue(null, new Value());
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
frankgh, line 1: We use `import org.jetbrains.annotations.NotNull;` consistently across the project, can we change it here?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -259,6 +267,21 @@ private void persist(@NotNull JavaPairRDD<DecoratedKey, Object[]> sortedRDD, Str
...
+    {
+        // create the importBarrier for non-coordinated-write mode.
+        // for coordinated write, the import barrier is created when starting the job
+        if (!writerContext.job().isCoordinatedWriteEnabled())
+        {
+            importBarrier = ImportCompletionCoordinator.of(startTimeNanos, writerContext, context.dataTransferApi(),
+                                                           writeValidator, resultsAsCloudStorageStreamResults,
+                                                           context.transportExtensionImplementation(), this::cancelJob);
+        }
+        Preconditions.checkState(importBarrier != null, "importBarrier is not initialized");
```
frankgh: NIT
```suggestion
        Objects.requireNonNull(importBarrier, "importBarrier is not initialized");
```

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/KryoSerializationTests.java
```diff
@@ -371,25 +374,74 @@ public void testSslConfig()
...
 
-        StorageTransportConfiguration deserialized;
-        try (Output out = serialize(config))
+        testSerDeserStorageTransportConfiguration(config);
+    }
+
+    @Test
+    public void testStorageTransportConfigurationWithMultiClusters()
+    {
+        MultiClusterContainer<StorageAccessConfiguration> readAccesses = new MultiClusterContainer<>();
```
frankgh: do we use this variable at all (`readAccesses`) for serde validation?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/transports/storage/extensions/TransportExtensionUtils.java
```diff
@@ -17,29 +17,26 @@
...
-     * @return the ClusterInfo associated with the clusterId, or null if not found
-     */
-    @Nullable
-    ClusterInfo cluster(@NotNull String clusterId);
+    public static void validateReceivedJobId(String receivedJobId, JobInfo jobInfo)
+    {
+        String actualJobId = jobInfo.getId();
+        if (!Objects.equals(receivedJobId, actualJobId))
+        {
+            throw new IllegalStateException("Received jobId does not match with the actual one. Received: " + receivedJobId
```
frankgh: NIT:
```suggestion
            throw new IllegalStateException("Received jobId does not match. Received: " + receivedJobId
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/transports/storage/extensions/StorageTransportHandler.java
```diff
@@ -53,79 +55,43 @@ public StorageTransportHandler(TransportContext.CloudStorageTransportContext tra
...
     {
-        validateReceivedJobId(jobId);
-        if (Objects.equals(transportContext.transportConfiguration().getStorageCredentialPair(), newCredentials))
+        validateReceivedJobId(jobId, jobInfo);
+        Preconditions.checkState(!jobInfo.isCoordinatedWriteEnabled() || clusterId != null,
+                                 "ClusterId cannot be null for coordinated write enabled jobs");
+        if (Objects.equals(transportContext.transportConfiguration().getStorageCredentialPair(clusterId), newCredentials))
         {
-            LOGGER.info("The received new credential is the same as the existing one. Skip updating.");
+            LOGGER.info("The received new credential is the same as the existing one. Skip updating. clusterId={}", clusterId);
```
frankgh: NIT:
```suggestion
            LOGGER.info("The received credentials have not changed. Skip updating. clusterId={}", clusterId);
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraCoordinatedBulkWriterContext.java
```diff
@@ -48,6 +49,10 @@ public CassandraCoordinatedBulkWriterContext(@NotNull BulkSparkConf conf,
         super(conf, structType, sparkDefaultParallelism);
         Preconditions.checkArgument(conf.isCoordinatedWriteConfigured(),
                                     "Cannot create CassandraCoordinatedBulkWriterContext without CoordinatedWrite configuration");
+        // Redundant check, since isCoordinatedWriteConfigured implies using S3_COMPAT mode already.
+        // Having it here for clarity.
+        Preconditions.checkArgument(conf.getTransportInfo().getTransport() == DataTransport.S3_COMPAT,
+                                    "CassandraCoordinatedBulkWriterContext only be created with " + DataTransport.S3_COMPAT);
```
frankgh:
```suggestion
                                    "CassandraCoordinatedBulkWriterContext can only be created with " + DataTransport.S3_COMPAT);
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
```diff
@@ -266,4 +276,14 @@ private RuntimeException toRuntimeException(ClusterInfo clusterInfo, Throwable c
...
         return new RuntimeException("Failed to perform action on cluster: " + clusterInfo.clusterId(), cause);
     }
+
+    private static void validateClusterInfoList(List<ClusterInfo> clusterInfos)
+    {
+        Preconditions.checkArgument(clusterInfos != null && !clusterInfos.isEmpty(),
+                                    "clusterInfos cannot be null or empty");
+        for (ClusterInfo ci : clusterInfos)
+        {
+            Preconditions.checkArgument(ci.clusterId() != null, "Found clusterInfo without clusterId defined");
```
frankgh: there's no easy way to identify which ClusterInfo is missing the clusterId, so it might be hard to debug.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/ImportCompletionCoordinator.java
```diff
@@ -362,7 +383,7 @@ private void handleFailedSliceInstance(RingInstance instance,
         catch (RuntimeException rte)
         {
             // record the first failure and cancel queued futures.
-            firstFailure.completeExceptionally(rte);
+            setImportFailure(new ImportFailedException(rte));
```
frankgh: can we use propagate here?
```suggestion
            setImportFailure(ImportFailedException.propagate(rte));
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CoordinatedCloudStorageDataTransferApi.java
```diff
@@ -0,0 +1,236 @@
...
+        QualifiedTableName qualifiedTableName = jobInfo.qualifiedTableName();
+        CreateRestoreJobSliceRequest request = new CreateRestoreJobSliceRequest(qualifiedTableName.keyspace(),
+                                                                                qualifiedTableName.table(),
+                                                                                jobInfo.getRestoreJobId(),
+                                                                                createSliceRequestPayload);
+        RetryPolicy retryPolicy = new CloudStorageDataTransferApiImpl.ExecutorCreateSliceRetryPolicy(sidecarClient);
+        RequestContext requestContext = sidecarClient.requestBuilder().retryPolicy(retryPolicy).request(request).build();
+        try
+        {
+            sidecarClient.executeRequestAsync(requestContext).get();
```
frankgh, line 178: you probably don't want this to wait indefinitely, can we upper bound it to something?
```suggestion
            sidecarClient.executeRequestAsync(requestContext).get(1, TimeUnit.MINUTES);
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CoordinatedCloudStorageDataTransferApi.java
```diff
@@ -0,0 +1,236 @@
...
+        });
+    }
+
+    @VisibleForTesting
+    RestoreJobProgressResponsePayload restoreJobProgress(CloudStorageDataTransferApiImpl dataTransferApi,
+                                                         RestoreJobProgressRequestParams requestParams)
+    {
+        try
+        {
+            return dataTransferApi.sidecarClient().restoreJobProgress(requestParams).get();
```
frankgh, line 245: upper bound it ?
```suggestion
            return dataTransferApi.sidecarClient().restoreJobProgress(requestParams).get(1, TimeUnit.MINUTE);
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -69,7 +78,7 @@ public class CassandraBulkSourceRelation extends BaseRelation implements Inserta
     private final Broadcast<BulkWriterContext> broadcastContext;
     private final BulkWriteValidator writeValidator;
     private final SimpleTaskScheduler simpleTaskScheduler;
-    private volatile ImportCompletionCoordinator importCoordinator = null; // value is only set when using S3_COMPAT
+    private volatile ImportBarrier importBarrier = null; // value is only set when using S3_COMPAT
```
frankgh: does this need to be volatile? Do we have a different thread updating / reading this value?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/CloudStorageDataTransferApi.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * Called from task level to create a restore slice.
+     * The request retries until the slice is created (201) or retry has exhausted.
+     *
+     * @param sidecarInstance           the sidecar instance where we will create the slice
+     * @param createSliceRequestPayload the payload to create the slice
+     * @throws SidecarApiCallException when an error occurs during the slice creation
+     */
+    void createRestoreSliceFromExecutor(SidecarInstance sidecarInstance,
```
frankgh, line 57: we have `org.apache.cassandra.spark.bulkwriter.cloudstorage.coordinated.CoordinatedCloudStorageDataTransferApiExtension#createRestoreSliceFromExecutor` I wonder why we need both?

frankgh (commented):
Looks good so far, I have some comments/suggestions on the PR

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/ImportBarrier.java
```diff
@@ -0,0 +1,48 @@
...
+     * Check whether the import operation has succeeded. The method does not block.
+     * @return true if import has succeeded. When import is pending or failed, it returns false
+     */
+    boolean succeeded();
+
+    /**
+     * Check whether the import operation has failed. The method does not block.
+     * @return ImportFailedException if import has failed; null otherwise
+     */
+    ImportFailedException importFailure();
```
JeetKunDoug: NIT: This could just be called `failure`

cassandra-analytics-core-example/src/main/java/org/apache/cassandra/spark/example/ExampleStorageTransportExtension.java
yifan-c: Object count and row count are still there. Reported from the `onAllObjectsPersisted` method, which is called right before STAGE_READY. 
The change only removes the duplication, so that the code is simplified a bit.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterContainer.java
yifan-c: Refactored to get rid of single

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/sparksql/CassandraDataSink.java
yifan-c: I think I was mostly not happy with having `Map<String, String> parameters` for `createRelation`. It is a scala map, but can be easily confused. I changed the parameter to have its FQCN. The method is not really required and removed.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/transports/storage/extensions/StorageTransportConfiguration.java
yifan-c: good catch. It was an oversight after the refactoring. Fixed.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/transports/storage/extensions/StorageTransportConfiguration.java
yifan-c: objectTags sounds good to me.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CoordinatedCloudStorageDataTransferApi.java
yifan-c: `getAnyValueOrThrow` is for the operations that can be performed on a single target and do not need broadcasting.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CoordinatedCloudStorageDataTransferApi.java
yifan-c: Yeah. Good catch. 
They are cancelling each other. I have updated the code.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraJobInfo.java
yifan-c: Yeah. The message is not suitable for project internal callsites. 

I have updated the message to "Method not supported, when coordinated write is enabled. Call-sites should get localDc from #coordinatedWriteConf() method"

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterContainer.java
yifan-c: done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/transports/storage/extensions/CredentialChangeListener.java
yifan-c: Yeah. Implementation of StorageTransportExtension should call it.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterContainerTest.java
yifan-c: What is the motivation?

frankgh: Because where possible we should leverage JDK APIs to avoid having a tight dependency with a 3rd party library, which is not controlled by the Analytics library, but rather it is controlled by the Spark environment where the job runs.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterContainerTest.java
yifan-c: done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
yifan-c: good catch.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/KryoSerializationTests.java
yifan-c: no. they are unused and not needed. Removed

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
yifan-c: Good call. I can move the validation to the location where the original input is accessible.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CoordinatedCloudStorageDataTransferApi.java
yifan-c: I would like to avoid the interference from future timing out. 
The request itself has retry policy configured to terminate in a bounded time. I see the sidecar client api desire some refactoring to be more straightforward/intuitive.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
yifan-c: no. removed. Cannot remember why it was declared this way. (volatile is not added in this patch)

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/CloudStorageDataTransferApi.java
yifan-c: The only difference is one takes a specific SidecarInstance as input, the other one (coordinated) can pick a random instance.

scripts/build-sidecar.sh
```diff
@@ -20,23 +20,32 @@
...
+echo $JVM_VERSION
+if [ "${JVM_VERSION}" == "1.8.0" ]; then
+  SIDECAR_BUILD_VERSION="1.0.0-jdk8-analytics"
+else
+  SIDECAR_BUILD_VERSION="1.0.0-analytics"
+fi
+
+# Build from local sidecar repo. For instance,
+# LOCAL_SIDECAR_REPO=/PATH/TO/cassandra-sidecar ./scripts/build-sidecar.sh
+if [ "x${LOCAL_SIDECAR_REPO}" != "x" ]; then
```
yifan-c, line 42: Allows to build from local sidecar repository.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/CassandraCloudStorageTransportContext.java
yifan-c: The check is a confidence check and redundant. I can remove it. I will see if any meaningful refactoring can be done here to clarify the entities relationship.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -161,7 +161,11 @@ public BulkSparkConf(SparkConf conf, Map<String, String> options)
         this.userProvidedSidecarPort = sidecarPortFromOptions.isPresent() ? sidecarPortFromOptions.get() : getOptionalInt(SIDECAR_PORT).orElse(-1);
         this.effectiveSidecarPort = this.userProvidedSidecarPort == -1 ? DEFAULT_SIDECAR_PORT : this.userProvidedSidecarPort;
         this.sidecarContactPointsValue = resolveSidecarContactPoints(options);
-        this.sidecarContactPoints = sidecarContactPoints();
+        // Coordinated-write builds its own sidecar contact points, skip when enabled.
+        if (!options.containsKey(WriterOptions.COORDINATED_WRITE_CONFIG.name()))
+        {
+            this.sidecarContactPoints = sidecarContactPoints();
+        }
```
JeetKunDoug: Given the `sidecarContactPoints()` method lazily builds these, and nothing touches the internal member, maybe just drop the whole chunk of code here rather than wrapping it in an if statement?
```suggestion
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -201,29 +211,30 @@ private void persist(@NotNull JavaPairRDD<DecoratedKey, Object[]> sortedRDD, Str
                 // Some token ranges might fail on instances, but the CL is still satisfied at this step
                 writeValidator.updateFailureHandler(streamResults);
 
-                List<BlobStreamResult> resultsAsBlobStreamResults = streamResults.stream()
-                                                                                 .map(BlobStreamResult.class::cast)
-                                                                                 .collect(Collectors.toList());
+                // todo: assert that coordinated write mode should not produce any stream error?
```
JeetKunDoug: Do we think this needs to be done? If so, please figure out how, otherwise remove the todo

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/ImportBarrier.java
```diff
@@ -0,0 +1,48 @@
...
+package org.apache.cassandra.spark.bulkwriter.cloudstorage;
+
+import org.apache.cassandra.spark.exception.ImportFailedException;
+
+/**
+ * A barrier to block for import operations.
+ * It is responsible to ensure that all bulk written data is imported into Cassandra cluster in the consistency safe way.
+ * {@link ImportFailedException} is thrown upon import failure, due to various reasons like timeout, consistency level not satisfied, etc.
+ */
+public interface ImportBarrier
```
JeetKunDoug: Both implementations of this interface are Import _Coordinators_ - can we name this similarly (and of course rename the members that hold them) as having two different names for this concept is confusing.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -381,32 +422,71 @@ void onDirectTransport(Consumer<TransportContext.DirectDataBulkWriterContext> co
     private void createRestoreJob(TransportContext.CloudStorageTransportContext context)
     {
         StorageTransportConfiguration conf = context.transportConfiguration();
-        RestoreJobSecrets secrets = conf.getStorageCredentialPair().toRestoreJobSecrets(conf.getReadRegion(),
-                                                                                        conf.getWriteRegion());
+        RestoreJobSecrets secrets = conf.getStorageCredentialPair(null)
```
JeetKunDoug, line 441: Using `null` as a key like this really bugs me - I can live with it for now but I think we need to think about how the direct/s3/s3-coordinated modes work and try to reorganize this code.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -201,29 +211,30 @@ private void persist(@NotNull JavaPairRDD<DecoratedKey, Object[]> sortedRDD, Str
...
-                                                             .mapToInt(res -> res.createdRestoreSlices.size())
-                                                             .sum();
+                List<CloudStorageStreamResult> resultsAsCloudStorageStreamResults = streamResults.stream()
+                                                                                                 .map(CloudStorageStreamResult.class::cast)
+                                                                                                 .collect(Collectors.toList());
+
+                int objectCount = resultsAsCloudStorageStreamResults.stream()
+                                                                    .mapToInt(res -> res.objectCount)
+                                                                    .sum();
                 // report the number of objects persisted on s3
```
JeetKunDoug: NIT: This comment isn't really right any more - it's objects + rows...

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -192,7 +202,7 @@ private void persist(@NotNull JavaPairRDD<DecoratedKey, Object[]> sortedRDD, Str
             boolean hasClusterTopologyChanged = writeResults.stream().anyMatch(WriteResult::isClusterResizeDetected);
 
             onCloudStorageTransport(context -> {
-                LOGGER.info("Waiting for Cassandra to complete import slices. rows={} bytes={} cluster_resized={}",
+                LOGGER.info("Waiting for Cassandra to complete import slices. rowCount={} totalBytes={} hasClusterTopologyChanged={}",
```
JeetKunDoug: Please refactor all of the code in this lambda into a separate method to clean up `persist` a bit - maybe `doCloudImport` or something like that?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/CassandraCloudStorageTransportContext.java
```diff
@@ -35,9 +35,11 @@
 import org.apache.cassandra.spark.bulkwriter.RingInstance;
 import org.apache.cassandra.spark.bulkwriter.SortedSSTableWriter;
 import org.apache.cassandra.spark.bulkwriter.TransportContext;
+import org.apache.cassandra.spark.bulkwriter.cloudstorage.coordinated.CoordinatedCloudStorageDataTransferApi;
 import org.apache.cassandra.spark.bulkwriter.token.ReplicaAwareFailureHandler;
 import org.apache.cassandra.spark.transports.storage.extensions.StorageTransportConfiguration;
 import org.apache.cassandra.spark.transports.storage.extensions.StorageTransportExtension;
+import org.apache.parquet.Preconditions;
```
JeetKunDoug:
```suggestion
import com.google.common.base.Preconditions;
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/CloudStorageStreamSession.java
```diff
@@ -168,23 +172,30 @@ protected StreamResult doFinalizeStream()
...
+                                                                             createdRestoreSlices,
+                                                                             sliceCount,
+                                                                             sstableWriter.rowCount(),
+                                                                             sstableWriter.bytesWritten());
         LOGGER.info("StreamResult: {}", streamResult);
-        // If the number of successful createSliceRequests cannot satisfy the configured consistency level,
-        // an exception is thrown and the task is failed. Spark might retry the task.
-        BulkWriteValidator.validateClOrFail(tokenRangeMapping, failureHandler, LOGGER, WRITE_PHASE, writerContext.job(), writerContext.cluster());
+        // CL validation is only required for non-coordinated write.
+        // For coordinated write, executors only need to make sure the bundles are uploaded and restore slice is created on sidecar.
```
JeetKunDoug, line 192: Could we not fail to create the necessary restore slices here and, if so, wouldn't we want to retry the task at this point by letting Spark's retry mechanism take over? If you don't do it here, it's impossible to retry later as the input dataset has been consumed at that point (unless we start persisting it in our own code, which would change the memory/disk requirements fairly significantly I think).

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/CloudStorageStreamSession.java
```diff
@@ -233,40 +244,59 @@ void sendBundle(Bundle bundle, boolean hasRefreshedCredentials)
...
+            });
+            // no need to add to the createdRestoreSlices. Driver polls the progress of the entire restore job.
+        }
+        else
         {
+            CreateSliceRequestPayload slicePayload = toCreateSliceRequestPayload(bundleStorageObject);
+            // Create slices on all replicas; remove the _failed_ replica if operation fails
+            replicas.removeIf(replica -> !tryCreateRestoreSlicePerReplica(replica, slicePayload));
+            // exit early if all replicas has failed, i.e. CL cannot be satisfied for sure.
+            if (replicas.isEmpty())
```
JeetKunDoug, line 268: Like above, would it be possible to actually check the CL here and fail early?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
```diff
@@ -61,10 +70,41 @@ public class CassandraClusterInfoGroup implements ClusterInfo, MultiClusterInfoP
...
     private transient volatile TokenRangeMapping<RingInstance> consolidatedTokenRangeMapping;
 
-    public CassandraClusterInfoGroup(List<ClusterInfo> clusterInfos)
+    /**
+     * Creates {@link CassandraClusterInfoGroup} with the list of {@link ClusterInfo} from {@link BulkSparkConf} and validation
+     * The validation ensures non-empty list of {@link ClusterInfo}, where all objects have non-empty and unique clusterId
+     * @param conf bulk write conf
+     * @return new {@link CassandraClusterInfoGroup} instance
+     */
+    public static CassandraClusterInfoGroup createFromBulkSparkConf(BulkSparkConf conf)
```
JeetKunDoug: NIT: We normally name things like this just `from*` in stead of `createFrom`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
```diff
@@ -61,10 +70,41 @@ public class CassandraClusterInfoGroup implements ClusterInfo, MultiClusterInfoP
...
+    {
+        CoordinatedWriteConf coordinatedWriteConf = conf.coordinatedWriteConf();
+        Preconditions.checkArgument(coordinatedWriteConf != null,
+                                    "CoordinatedWriteConf must present for CassandraCoordinatedBulkWriterContext");
```
JeetKunDoug:
```suggestion
                                    "In order to create an instance of CassandraCoordinatedBulkWriterContext, you must provide the appropriate coordinated write configuration by setting the `COORDINATED_WRITE_CONFIG` writer option.");
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterContainer.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * @return any value in the container. Null is returned if the container is empty.
+     */
+    @Nullable
+    public T getAnyValue()
+    {
+        T v = null;
+        if (!byCluster.isEmpty())
```
JeetKunDoug, line 65: Now that we have the whole sentinel value for null, is there any case where `byCluster` would be empty? Wouldn't you always at least have the `null` key for non-coordinated writes? If so, we should be able to get rid of most/all of these checks.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/TwoPhaseImportCoordinator.java
```diff
@@ -0,0 +1,267 @@
...
+        LOGGER.info("Received StageReady signal for coordinated write. Notifying Sidecars. jobId={}", jobId);
+        stageReady.complete(RestoreJobStatus.STAGE_READY);
+    }
+
+    @Override
+    public void onApplyReady(String jobId)
+    {
+        validateReceivedJobId(jobId, job);
+        LOGGER.info("Received ApplyReady signal for coordinated write. Notifying Sidecars. jobId={}", jobId);
+        importReady.complete(RestoreJobStatus.IMPORT_READY);
```
JeetKunDoug: Can we be consistent with the Stage/Apply terminology here?

JeetKunDoug (commented):
More comments - will finish up tonight.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/TwoPhaseImportCoordinator.java
```diff
@@ -0,0 +1,267 @@
...
+/**
+ * Import coordinator that implements the two phase import for coordinated write.
+ * 1. It waits for the stageReady signal from StorageTransportExtension and update the restore job status on all clusters to STAGE_READY
+ * 2. It, then, polls the stage progress of each cluster and notifies StorageTransportExtension on completion
+ * 3. It waits for the applyReady signal from StorageTransportExtension and update the restore job status on all clusters to APPLY_READY
+ * 4. It polls the apply/import progress of each cluster
+ * 5. Finally, the two phase import completes.
+ * The procedure is programed in {@link #awaitInternal()}
+ */
+public class TwoPhaseImportCoordinator implements ImportBarrier, CoordinationSignalListener
```
JeetKunDoug: Can we call this `CoordinatedImportCoordinator` maybe? Try to keep the same name for all of the coordinated import stuff. Also, "Two Phase" implies something that we don't support (an actual commit).

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/TwoPhaseImportCoordinator.java
```diff
@@ -0,0 +1,267 @@
...
+import org.apache.cassandra.spark.bulkwriter.cloudstorage.ImportBarrier;
+import org.apache.cassandra.spark.exception.ConsistencyNotSatisfiedException;
+import org.apache.cassandra.spark.exception.ImportFailedException;
+import org.apache.cassandra.spark.transports.storage.extensions.CoordinationSignalListener;
+import org.apache.cassandra.spark.transports.storage.extensions.StorageTransportExtension;
+
+import static org.apache.cassandra.spark.transports.storage.extensions.TransportExtensionUtils.validateReceivedJobId;
+
+/**
+ * Import coordinator that implements the two phase import for coordinated write.
```
JeetKunDoug: Maybe "staged import" vs. "two phase import" (see below for context)

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/sparksql/CassandraDataSink.java
```diff
@@ -56,47 +55,45 @@ public String shortName()
...
                 try
                 {
+                    writerContext = factory().createBulkWriterContext(sqlContext.sparkContext(),
+                                                                      ScalaConversionUtils.<String, String>mapAsJavaMap(parameters),
+                                                                      data.schema());
                     JobInfo jobInfo = writerContext.job();
-                    String description = "Cassandra Bulk Load for table " + jobInfo.qualifiedTableName();
                     CassandraBulkSourceRelation relation = new CassandraBulkSourceRelation(writerContext, sqlContext);
                     // Initialize the job group ID for later use if we need to cancel the job
                     // TODO: Can we get a more descriptive "description" in here from the end user somehow?
```
JeetKunDoug, line 79: After several years, maybe it's time to get rid of this TODO?
```suggestion
```

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/KryoSerializationTests.java
```diff
@@ -371,25 +374,65 @@ public void testSslConfig()
...
-        );
+        ImmutableMap.of("tag1", "tagVal1", "tag2", "tagVal2"));
 
-        StorageTransportConfiguration deserialized;
-        try (Output out = serialize(config))
+        testSerDeserStorageTransportConfiguration(config);
+    }
+
+    @Test
+    public void testStorageTransportConfigurationWithMultiClusters()
```
JeetKunDoug, line 394: Another "after this PR refactoring note" - maybe not every class should have its Kryo tests in this one test class?

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/CloudStorageDataTransferApiFactoryTest.java
```diff
@@ -0,0 +1,61 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.RETURNS_DEEP_STUBS;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+class CloudStorageDataTransferApiFactoryTest
+{
+    @Test
+    void testCreateCloudStorageDataTransferApi()
```
JeetKunDoug, line 43: NIT: Please split this into 3 tests that each test one part of this and have more descriptive method names.

JeetKunDoug (commented):
Last few comments - lots of great stuff here. Thanks Yifan!

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
yifan-c: yeah. not needed. removed.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
yifan-c: addressed with warning log for the unexpected state and avoid interrupting the write.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/ImportBarrier.java
yifan-c: fair. Renamed both to barrier.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
yifan-c: ok. Adding a todo.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
yifan-c: getting rid of the comment. The method name and log message are self-explanatory.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/CloudStorageStreamSession.java
yifan-c: Creating restore slice and uploading to S3 can both fail. We want to retry the such task by throwing the RuntimeException from `sendBundle` method. 

The `if` statement here is to skip CL validation by the task executor. When in coordinated write mode, CL validation is performed by sidecar.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/MultiClusterContainer.java
yifan-c: The value container can be empty, when no entries are populated. Empty check is still required to get any value from the map (, since it returns the first value from `values()`).

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/TwoPhaseImportCoordinator.java
yifan-c: ok.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/TwoPhaseImportCoordinator.java
yifan-c: It is two phase import, not the [two-phase commit protocol](https://en.wikipedia.org/wiki/Two-phase_commit_protocol).

That being said, the names do share a good amount of similarities. I am renaming it to `CoordinatedImportBarrier`.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/TwoPhaseImportCoordinator.java
yifan-c: There are indeed 2 phases for the coordinated write procedure. I do not see how avoiding mentioning two phase is any better. If the concern is the atomicity guarantee, the clarification can be added somewhere, e.g. at `WriterOptions#COORDINATED_WRITE_CONFIG`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/sparksql/CassandraDataSink.java
yifan-c: I would prefer addressing in a separate patch.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/KryoSerializationTests.java
yifan-c: I think they are placed all in one class because of the below code block for convenience. 

I agree that we can separate the test cases into their own test classes, for many benefits.

```java
    static
    {
        new KryoRegister().registerClasses(KRYO);
        new SbwKryoRegistrator().registerClasses(KRYO);
    }
```

cassandra-analytics-core-example/src/main/java/org/apache/cassandra/spark/example/ExampleStorageTransportExtension.java
```diff
@@ -169,13 +169,13 @@ public void onStageFailed(String clusterId, Throwable cause)
     }
 
     @Override
-    public void onApplySucceeded(String clusterId, long elapsedMillis)
+    public void onImportSucceeded(String clusterId, long elapsedMillis)
     {
         LOGGER.info("Job {} has all objects applied at cluster {} after {}ms", jobId, clusterId, elapsedMillis);
```
JeetKunDoug, line 174: NIT: Where we changed the method name from Apply -> Import we should also change the logging text. This goes for all instances of onImportSucceeded/onImportFailed please.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -42,10 +42,10 @@
 import org.apache.cassandra.spark.bulkwriter.cloudstorage.CloudStorageDataTransferApi;
 import org.apache.cassandra.spark.bulkwriter.cloudstorage.CloudStorageStreamResult;
 import org.apache.cassandra.spark.bulkwriter.cloudstorage.ImportBarrier;
-import org.apache.cassandra.spark.bulkwriter.cloudstorage.ImportCompletionCoordinator;
+import org.apache.cassandra.spark.bulkwriter.cloudstorage.ImportCompletionBarrier;
```
JeetKunDoug: I really think "coordinator" is a better term than "barrier" - I realize that you may end up with "CoordinatedImportCoordinator" which sounds repetitive, but it better reflects what the class is doing I think as it does more than just wait... it is the thing that orchestrates the imports on a single cluster or multiple clusters.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
```diff
@@ -270,16 +240,62 @@ private void persist(@NotNull JavaPairRDD<DecoratedKey, Object[]> sortedRDD, Str
...
+                    rowCount,
+                    totalBytesWritten,
+                    hasClusterTopologyChanged);
+
+        List<StreamError> allErrors = streamResults.stream().flatMap(r -> r.failures.stream()).collect(Collectors.toList());
+        if (writerContext.job().isCoordinatedWriteEnabled())
+        {
+            if (!allErrors.isEmpty())
+            {
+                // log a warning and move on
```
JeetKunDoug: Why move on in this case? If it's expected to be empty but isn't would that not be an error that should cause the job to fail early vs. continuing?

JeetKunDoug (commented):
A couple more small things, but looking good. Thanks for addressing the previous feedback.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkSourceRelation.java
yifan-c: ok. I will throw illegal state to be strict.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriteValidator.java
```diff
@@ -87,14 +87,12 @@ public void setPhase(String phase)
...
-    public synchronized void updateFailureHandler(List<? extends StreamResult> results)
+    public synchronized void updateFailureHandler(List<StreamError> streamErrors)
     {
-        results.stream()
-               .flatMap(res -> res.failures.stream())
-               .forEach(err -> {
-                   LOGGER.info("Populate stream error from tasks. {}", err);
-                   failureHandler.addFailure(err.failedRange, err.instance, err.errMsg);
-               });
+        streamErrors.forEach(err -> {
```
frankgh, line 92: 👍

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CoordinatedCloudStorageDataTransferApi.java
frankgh: yeah, that's a fair point. I guess it's not easy to distinguish whether there's a defined upperbound. Maybe something to keep in mind for a v2 of the client API

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/ImportCompletionCoordinator.java
```diff
@@ -42,31 +43,36 @@
...
 import org.apache.cassandra.util.ThreadUtil;
 
 import static org.apache.cassandra.clients.Sidecar.toSidecarInstance;
-import static org.apache.cassandra.spark.bulkwriter.blobupload.CreatedRestoreSlice.ConsistencyLevelCheckResult.NOT_SATISFIED;
-import static org.apache.cassandra.spark.bulkwriter.blobupload.CreatedRestoreSlice.ConsistencyLevelCheckResult.SATISFIED;
+import static org.apache.cassandra.spark.bulkwriter.cloudstorage.CreatedRestoreSlice.ConsistencyLevelCheckResult.NOT_SATISFIED;
+import static org.apache.cassandra.spark.bulkwriter.cloudstorage.CreatedRestoreSlice.ConsistencyLevelCheckResult.SATISFIED;
 
-public final class ImportCompletionCoordinator
+public final class ImportCompletionCoordinator implements ImportCoordinator
```
frankgh, line 67: Does it make sense to call this class something like `TimeBoundedImportCoordinator` ?

frankgh: Maybe the javadoc should explain what are the terminal conditions for the coordinated import?

frankgh (approved):
Some minor comments, but this looks good. Thanks for working through the comments.

PR #92, opened 2024-11-07
CASSANDRA-20066: Expose detailed bulk write failure message for better insight

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSANDRA-20066

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriteValidator.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriteValidator.java
@@ -38,6 +38,7 @@ import org.apache.cassandra.spark.exception.ConsistencyNotSatisfiedException;
 public class BulkWriteValidator
 {
     private static final Logger LOGGER = LoggerFactory.getLogger(BulkWriteValidator.class);
+    private static final int ERROR_MESSAGE_MAX_LENGTH = 1024 * 64;
 
     private final ClusterInfo cluster;
     private final ReplicaAwareFailureHandler<RingInstance> failureHandler;
@@ -68,10 +69,11 @@ public class BulkWriteValidator
         }
         else
         {
-            String message = String.format("Failed to write %s ranges with %s for job %s in phase %s.",
+            String header = String.format("Failed to write %s ranges with %s for job %s in phase %s. ",
                                            failedRanges.size(), job.getConsistencyLevel(), job.getId(), phase);
+            String errorDetails = aggregateErrors(failedRanges);
+            String message = header + errorDetails;
             logger.error(message);
-            logFailedRanges(logger, phase, failedRanges);
             throw new ConsistencyNotSatisfiedException(message);
         }
     }
@@ -110,19 +112,24 @@ public class BulkWriteValidator
         });
     }
 
-    private static void logFailedRanges(Logger logger, String phase,
-                                        List<ReplicaAwareFailureHandler<RingInstance>.ConsistencyFailurePerRange> failedRanges)
+    // aggregate the stream errors in order to provide a better insight on failure
+    private static String aggregateErrors(List<ReplicaAwareFailureHandler<RingInstance>.ConsistencyFailurePerRange> failedRanges)
     {
+        StringBuilder sb = new StringBuilder();
         for (ReplicaAwareFailureHandler<RingInstance>.ConsistencyFailurePerRange failedRange : failedRanges)
         {
-            failedRange.failuresPerInstance.forEachInstance((instance, errors) -> {
-                logger.error("Failed in phase {} for {} on {}. Failure: {}",
-                             phase,
-                             failedRange.range,
-                             instance.toString(),
-                             errors);
-            });
+            for (Map.Entry<RingInstance, Collection<String>> entry : failedRange.failuresPerInstance.entrySet())
+            {
+                sb.append("Instance: ").append(entry.getKey()).append("All failures: ").append(entry.getValue());
+                if (sb.length() > ERROR_MESSAGE_MAX_LENGTH)
+                {
+                    sb.setLength(ERROR_MESSAGE_MAX_LENGTH - 3);
+                    sb.append("...");
+                    return sb.toString();
+                }
+            }
         }
+        return sb.toString();
     }
 
     public void updateFailureHandler(Range<BigInteger> failedRange, RingInstance instance, String reason)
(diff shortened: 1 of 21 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriteValidator.java
```diff
@@ -110,19 +112,24 @@ public static void updateFailureHandler(CommitResult commitResult,
...
-            failedRange.failuresPerInstance.forEachInstance((instance, errors) -> {
-                logger.error("Failed in phase {} for {} on {}. Failure: {}",
-                             phase,
-                             failedRange.range,
-                             instance.toString(),
-                             errors);
-            });
+            for (Map.Entry<RingInstance, Collection<String>> entry : failedRange.failuresPerInstance.entrySet())
+            {
+                sb.append("Instance: ").append(entry.getKey()).append(" All failures: ").append(entry.getValue());
```
frankgh: should we add line breaks?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriteValidator.java
```diff
@@ -110,19 +112,24 @@ public static void updateFailureHandler(CommitResult commitResult,
...
-    private static void logFailedRanges(Logger logger, String phase,
-                                        List<ReplicaAwareFailureHandler<RingInstance>.ConsistencyFailurePerRange> failedRanges)
+    // aggregate the stream errors in order to provide a better insight on failure
+    private static String aggregateErrors(List<ReplicaAwareFailureHandler<RingInstance>.ConsistencyFailurePerRange> failedRanges)
     {
+        StringBuilder sb = new StringBuilder();
         for (ReplicaAwareFailureHandler<RingInstance>.ConsistencyFailurePerRange failedRange : failedRanges)
         {
-            failedRange.failuresPerInstance.forEachInstance((instance, errors) -> {
-                logger.error("Failed in phase {} for {} on {}. Failure: {}",
```
JeetKunDoug, line 119: Not thrilled with the fact that we lost the logging completely - I'd build the error message/cap its length, but continue logging everything, rather than just building the error and returning it.

scripts/build-dtest-jars.sh
```diff
@@ -52,7 +52,7 @@ else
   if [[ "$CLEAN" == "true" ]]; then
     echo "Clean up $DTEST_JAR_DIR"
     rm -rf "$DTEST_JAR_DIR/cassandra-build"
-    rm "$DTEST_JAR_DIR/dtest*.jar"
+    rm -rf "$DTEST_JAR_DIR/dtest*.jar"
```
JeetKunDoug: Why? you should only be deleting the jar _files_ so what does `-r` do here?

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/correctness/BulkWriteCorruptionTest.java
```diff
@@ -53,22 +56,55 @@
...
+    }
+
+    private static final QualifiedName QUALIFIED_NAME = new QualifiedName(TEST_KEYSPACE, "test_write_corruption");
 
     static
     {
-        // Intercepts SortedSSTableWriter#validateSSTables to corrupt file on purpose.
         // Install the class rebase the earliest, before JVM loads the class
-        BBHelperInterceptSortedSSTableWriterValidateSSTables.install();
+        BBHelperFileCorrupter.install();
```
JeetKunDoug: NIT: Why not just have 2 different byte buddy helper classes, rather than introducing the mode enum/if statements in two different intercepted methods?

JeetKunDoug (changes requested):
Thanks for the patch - I like the idea of pushing some of these errors into the main exception given the "fun" folks have debugging Spark jobs, but I'd like to not lose the complete logging at the driver... and a few other small questions.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriteValidator.java
yifan-c: The logging is moved to `validateClOrFail`. However, I should probably not truncate the message here. Making an update.

scripts/build-dtest-jars.sh
yifan-c: Because it failed to remove files. I got this error.
```
+ rm '.../cassandra-analytics/dependencies/dtest*.jar'
rm: .../cassandra-analytics/dependencies/dtest*.jar: No such file or directory
```

However, adding `-rf` is not the correct fix. It just hides the error. 

The correct `rm` command is `rm "$DTEST_JAR_DIR"/dtest*.jar`

Thanks for bringing it up!

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/correctness/BulkWriteCorruptionTest.java
yifan-c: Yep. It sounds better.

yifan-c: Made the change and I realized why having the switch. 

Each test case only expect corruption at specific locations. Without the switch, `testDataTransferCorruption` has the files corrupted on disk and fail.

JeetKunDoug: I think there may be a way around this, but it seems like it requires some more knowledge of ByteBuddy, and I don't have that knowledge right now, so I don't want to hold up this PR.

JeetKunDoug (approved):
:shipit: !

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriteValidator.java
```diff
@@ -110,19 +112,27 @@ public static void updateFailureHandler(CommitResult commitResult,
...
-                             phase,
-                             failedRange.range,
-                             instance.toString(),
-                             errors);
-            });
+            for (Map.Entry<RingInstance, Collection<String>> entry : failedRange.failuresPerInstance.entrySet())
+            {
+                logger.error("Failed in phase {} for {} on {}. Failure: {}", phase, failedRange.range, entry.getKey(), entry.getValue());
+                if (sb.length() > ERROR_MESSAGE_MAX_LENGTH)
```
frankgh: should be >=

frankgh (commented):
Small comments, but looks good in general. Thanks for the additional tests also

frankgh (approved):
+1

PR #97, opened 2025-01-29
CASSANDRA-20259: Refactor to decouple RowIterator and CellIterator fr…

jberragan:
…om Spark so bulk reads can be performed outside of Spark


jberragan:
CircleCI: https://app.circleci.com/pipelines/github/jberragan/cassandra-analytics/139/workflows/86288cbe-015e-42e4-9a78-de98d3559404

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/CellIterator.java
```diff
@@ -0,0 +1,385 @@
...
+    private boolean closed = false;
+    private Cell next = null;
+    private long previousTimeNanos;
+
+    protected final int partitionId;
+    protected final int firstProjectedValueColumnPositionOrZero;
+    protected final boolean hasProjectedValueColumns;
+    protected final TypeConverter typeConverter;
+
+    public interface ScannerSupplier
```
frankgh, line 76: can we add javadocs for interfaces?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/FullRowBuilder.java
```diff
@@ -133,16 +135,13 @@ public void onCell(Cell cell)
     @Override
     public int fieldIndex(String name)
     {
-        List<CqlField> fields = cqlTable.fields();
-        return IntStream.range(0, fields.size())
-                        .filter(i -> Objects.equals(fields.get(i).name(), name))
-                        .findFirst()
-                        .orElse(-1);
+        return Optional.ofNullable(cqlTable.getField(name))
```
frankgh, line 138: This looks safe, but I am a bit concerned the position might not always be == the index of the field. I'm thinking of the case where you do a select c, a, b  when the table schema was created with a, b, c. This is the only concern I'd have with this change. I can spend some time looking into whether this could change the behavior, but if you've already spent some time on this, I'd rather not. I assume some test would have caught this anyway.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/sparksql/SparkCellIterator.java
```diff
@@ -19,115 +19,50 @@
...
-        stats.openedSparkCellIterator();
-        firstProjectedValueColumnPositionOrZero = maybeGetPositionOfFirstProjectedValueColumnOrZero();
-
-        sparkSqlTypeConverter = dataLayer.typeConverter();
-        sparkTypes = new SparkType[cqlTable.numFields()];
+        this.sparkTypes = new SparkType[cqlTable.numFields()];
         for (int index = 0; index < cqlTable.numFields(); index++)
         {
-            sparkTypes[index] = sparkSqlTypeConverter.toSparkType(cqlTable.field(index).type());
+            this.sparkTypes[index] = ((SparkSqlTypeConverter) this.typeConverter).toSparkType(cqlTable.field(index).type());
```
frankgh: NIT : do the casting of the type converter outside of the for loop and assign it to a local variable. To avoid casting type casting on every iteration

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/CellIterator.java
```diff
@@ -0,0 +1,385 @@
...
+        scanner = scannerSupplier.get(partitionId, partitionKeyFilters, columnFilter);
+        long openTimeNanos = System.nanoTime() - startTimeNanos;
+        LOGGER.info("Opened CompactionScanner runtimeNanos={}", openTimeNanos);
+        stats.openedCompactionScanner(openTimeNanos);
+        rowData = scanner.data();
+        stats.openedSparkCellIterator();
+        firstProjectedValueColumnPositionOrZero = maybeGetPositionOfFirstProjectedValueColumnOrZero();
+    }
+
+    public boolean hasProjectedValueColumns()
```
frankgh, line 133: it looks like this method is no longer used, should we remove it?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/CellIterator.java
```diff
@@ -0,0 +1,385 @@
...
+            return true;
+        }
+
+        // Finished so close
+        next = null;
+        close();
+        return false;
+    }
+
+    public String decodeString(@Nullable ByteBuffer buffer) throws CharacterCodingException
```
frankgh: NIT: can we scope it to protected?
```suggestion
    protected String decodeString(@Nullable ByteBuffer buffer) throws CharacterCodingException
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/sparksql/AbstractSparkRowIterator.java
```diff
@@ -92,66 +78,21 @@ private static boolean useColumnFilter(@Nullable StructType requiredSchema, CqlT
...
-            // Keep reading more cells until we read the entire row
-        } while (builder.hasMoreCells() && it.hasNextThrows());
-
-        // Build row and reset builder for next row
-        row = builder.build();
-        builder.reset();
-
-        stats.nextRow();
-        return true;
+        return new PartialRowBuilder<>(requiredColumns, cqlTable, hasProjectedValueColumns, this::mapper);
```
frankgh: NIT:

```
        return new PartialRowBuilder<>(Objects.requireNonNull(requiredColumns, "requiredColumns must be non-null for PartialRowBuilder"),
                                       cqlTable, hasProjectedValueColumns, this::mapper);
```

or 

```
        Objects.requireNonNull(requiredColumns, "requiredColumns must be non-null for PartialRowBuilder");
        return new PartialRowBuilder<>(requiredColumns, cqlTable, hasProjectedValueColumns, this::mapper);
```

frankgh (commented):
Looks good, I have a few comments and small minor suggestions.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/FullRowBuilder.java
jberragan: It's an important point, for the `FullRowBuilder` the SparkSQL column order is the same as the Cassandra CQL column order.  In `CqlTable` the `fields` list is sorted by `position` anyhow so there should be no change in behaviour between this and the previous code.

I'm not entirely sure how this O(n) search through `fields` came in here, but we can use the  CqlTable to lookup the position with the map.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/CellIterator.java
jberragan: Ok

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/FullRowBuilder.java
frankgh: As long as out of order queries where columns are not in the order of the underlying defined schemas work correctly, I have no concerns.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/RowIterator.java
```diff
@@ -0,0 +1,148 @@
...
+                          Stats stats,
+                          @Nullable String[] requiredColumns,
+                          Function<RowBuilder<T>, RowBuilder<T>> decorator)
+    {
+        this.stats = stats;
+        this.it = it;
+        this.openTimeNanos = System.nanoTime();
+        this.cqlTable = it.cqlTable;
+        this.requiredColumns = requiredColumns;
+        this.hasProjectedValueColumns = it.hasProjectedValueColumns;
```
yifan-c:
```suggestion
        this.hasProjectedValueColumns = it.hasProjectedValueColumns();
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/sparksql/SparkCellIterator.java
jberragan: Makes sense

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/CellIterator.java
```diff
@@ -0,0 +1,386 @@
...
+            scanner.advanceToNextColumn();
+
+            // Skip partition e.g. if token is outside of Spark worker token range
+            if (skipPartition)
+            {
+                continue;
+            }
+
+            // Deserialize clustering keys - if moved to new CQL row - and update 'values' Object[] array
+            ByteBuffer columnNameBuf = Objects.requireNonNull(rowData.getColumnName(), "ColumnName buffer in Rid is null, this is unexpected");
```
yifan-c: Rid is now `RowData`. Can you update the message?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/CellIterator.java
```diff
@@ -0,0 +1,386 @@
...
+        {
+            stats.skippedPartitionInIterator(rowData.getPartitionKey(), rowData.getToken());
+            return;
+        }
+
+        // Or new partition, so deserialize partition keys and update 'values' array
+        readPartitionKey(typeConverter, rowData.getPartitionKey(), cqlTable, this.values, stats);
+    }
+
+    public abstract boolean isInPartition(int partitionId, BigInteger token, ByteBuffer partitionKey);
```
yifan-c, line 303: Can you group the abstract methods in this file together and add java doc for them?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/FullRowBuilder.java
```diff
@@ -19,34 +19,36 @@
...
-    Object[] result;
-    int count;
+    protected final int numColumns;
+    protected final int numCells;
+    protected final boolean hasProjectedValueColumns;
+    protected int extraColumns;
+    protected Object[] result;
+    protected int count;
     private final CqlTable cqlTable;
+    protected final Function<Object[], T> mapper;
```
yifan-c: nit: a more descriptive name than "mapper"?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/sparksql/AbstractSparkRowIterator.java
```diff
@@ -92,66 +79,18 @@ private static boolean useColumnFilter(@Nullable StructType requiredSchema, CqlT
                        .anyMatch(field -> !requiredFields.contains(field));
     }
 
-    abstract RowBuilder newBuilder();
+    public abstract T mapper(Object[] result);
```
yifan-c: add javadoc and (nit) rename the method to be more descriptive?

frankgh (approved):
+1 Looks good to me

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/RowIterator.java
```diff
@@ -0,0 +1,143 @@
...
+
+        builder = decorator.apply(builder);
+
+        builder.reset();
+        return builder;
+    }
+
+    public abstract PartialRowBuilder<T> newPartialBuilder();
+
+    public abstract FullRowBuilder<T> newFullRowBuilder();
```
yifan-c: javadoc

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/RowBuilder.java
```diff
@@ -20,9 +20,8 @@
 package org.apache.cassandra.spark.sparksql;
 
 import org.apache.cassandra.spark.data.CqlTable;
-import org.apache.spark.sql.catalyst.expressions.GenericInternalRow;
 
-public interface RowBuilder
+public interface RowBuilder<T>
```
yifan-c, line 24: Since RowBuilder is now generic, I think the `SchemaFeature` will have to updated according to declare the generic type. Right? The current features are all decorating the spark row.

PR #98, opened 2025-01-31
Add CassandraBridge helper APIs that can be used by external tooling

jberragan:


cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
```diff
@@ -96,6 +98,12 @@ public static ReplicationStrategy getEnum(String value)
         }
     }
 
+    public static ReplicationFactor simple(int rf)
```
yifan-c: nit: how about `simpleStrategy`?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/TypeConverter.java
```diff
@@ -23,6 +23,8 @@
 
 public interface TypeConverter
 {
+    TypeConverter STUB = (cqlType, value, isFrozen) -> value;
```
yifan-c: My understanding of the term stub is a placeholder. 

The function is more like `Function.identity`, which returns the input as-is. 

Maybe we name it `IDENTITY` too?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/RowIterator.java
```diff
@@ -58,6 +65,55 @@ protected RowIterator(CellIterator it,
...
+    /**
+     * @param it              CellIterator cell iterate instance to iterate over each cell in the data.
+     * @param stats           stats instance to emit KPIs
+     * @param requiredColumns optional column filter to select a subset of columns, if null then all columns are returned.
+     * @return a basic RowIterator that transforms Object[] to a Map keyed on column name
+     */
+    public static RowIterator<Map<String, Object>> rowMapIterator(CellIterator it, Stats stats, @Nullable String[] requiredColumns)
+    {
+        final CqlTable cqlTable = it.cqlTable();
+        final boolean hasProjectedValueColumns = it.hasProjectedValueColumns();
```
yifan-c: please remove the `final`s

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/RowIterator.java
```diff
@@ -58,6 +65,55 @@ protected RowIterator(CellIterator it,
...
+                                                            .collect(Collectors.toMap(i -> requiredColumns[i], Function.identity()));
+
+                return new PartialRowBuilder<>(
+                requiredColumns, it.cqlTable(), hasProjectedValueColumns,
+                (valueArray) -> {
+                    Map<String, Object> row = new HashMap<>(requiredColumns.length);
+                    for (String field : requiredColumns)
+                    {
+                        row.put(field, valueArray[columnIndex.get(field)]);
+                    }
```
yifan-c: The `columnIndex` map is probably not needed if you populate the row this way. 

```suggestion
                    for (int i = 0; i < requiredColumns.length; i++)
                    {
                        String field = requiredColumns[i];
                        row.put(field, valueArray[i]);
                    }
```

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/RowIterator.java
```diff
@@ -58,6 +65,55 @@ protected RowIterator(CellIterator it,
...
+    public static RowIterator<Map<String, Object>> rowMapIterator(CellIterator it, Stats stats, @Nullable String[] requiredColumns)
+    {
+        final CqlTable cqlTable = it.cqlTable();
+        final boolean hasProjectedValueColumns = it.hasProjectedValueColumns();
+        return new RowIterator<>(it, stats, requiredColumns, Function.identity())
+        {
+            @Override
+            public PartialRowBuilder<Map<String, Object>> newPartialBuilder()
+            {
+                Objects.requireNonNull(requiredColumns, "requiredColumns must be non-null to use PartialRowBuilder");
```
yifan-c: nit: can you move the validation inside the `PartialRowBuilder#constructor`? The same validation is done in both here and `AbstractSparkRowIterator`

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/filters/PartitionKeyFilter.java
```diff
@@ -33,7 +33,7 @@
 import org.apache.cassandra.spark.utils.ByteBufferUtils;
 import org.jetbrains.annotations.NotNull;
 
-public final class PartitionKeyFilter implements Serializable
+public final class PartitionKeyFilter implements Serializable, Comparable<PartitionKeyFilter>
```
yifan-c, line 36: It is unclear to me about the semantics of comparing two filters. Can you explain? 
And why removing the `filter` method?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/ByteBufferUtils.java
```diff
@@ -91,6 +94,16 @@ public static byte[] getArray(ByteBuffer buffer)
         return bytes;
     }
 
+    public static String readShortLengthString(ByteBuffer buf)
+    {
+        int len = buf.getShort();
+        byte[] ar = new byte[len];
+        buf.get(ar);
+        String str = new String(ar, StandardCharsets.UTF_8);
+        buf.get();
```
yifan-c: why consume an extra byte and throw away? It is not a behavior of reading string, but to decode from `CompositeType` bytes

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/ByteBufferUtils.java
```diff
@@ -272,6 +285,28 @@ public static ByteBuffer buildPartitionKey(List<CqlField> partitionKeys, Object.
         return build(false, buffers);
     }
 
+    public static Map<String, Object> decodePartitionKey(CqlTable table, ByteBuffer partitionKey)
```
yifan-c: Add a test for it?

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/TestUtils.java
```diff
@@ -388,4 +388,16 @@ public static Range<BigInteger> range(BigInteger start, BigInteger end)
     {
         return Range.openClosed(start, end);
     }
+
+    public static void createDirectory(Path directory)
```
yifan-c, line 392: Just to mention that the same functionality is already provided by junit. 
But I am fine with defining our own testing utility.

cassandra-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridge.java
```diff
@@ -443,6 +445,198 @@ public ByteBuffer uncompress(ByteBuffer input) throws IOException
...
+
+    /**
+     * @param ssTable          SSTable instance
+     * @param partitioner      Cassandra partitioner
+     * @param minIndexInterval minIndexInterval configured in the TableMetaData
+     * @param maxIndexInterval maxIndexInterval configured in the TableMetadata
+     * @return the first and last token by attempting to read from the Summary.db file first then failing back to the Index.db.
+     * @throws IOException
+     */
+    public abstract Pair<BigInteger, BigInteger> firstLastToken(SSTable ssTable,
```
yifan-c: naming is hard.. I was wondering what is the token that is the first of the last tokens..

Anyway, there is already a method to expose `org.apache.cassandra.bridge.SSTableSummary`. This new method is sort of redundant, in the presence of `org.apache.cassandra.bridge.CassandraBridgeImplementation#getSSTableSummary`. 
Probably merge them or make it an overload method.

cassandra-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridge.java
```diff
@@ -443,6 +445,198 @@ public ByteBuffer uncompress(ByteBuffer input) throws IOException
...
+     * @param maxIndexInterval maxIndexInterval configured in the TableMetadata
+     * @param ranges           a list of token ranges
+     * @return a list boolean value if corresponding token range in `ranges` list parameter overlaps with the SSTable.
+     * The SSTable may or may not contain data for the range.
+     */
+    public abstract List<Boolean> overlaps(SSTable ssTable,
+                                           Partitioner partitioner,
+                                           int minIndexInterval,
+                                           int maxIndexInterval,
+                                           List<TokenRange> ranges) throws IOException;
```
yifan-c: It seems to be out of the scope if the range of the sstable is already exposed, since client should be able to check quite easily with the list of ranges. But I am not strongly against adding it.

cassandra-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridge.java
```diff
@@ -443,6 +445,198 @@ public ByteBuffer uncompress(ByteBuffer input) throws IOException
...
+
+    /**
+     * @param partitioner     Cassandra partitioner
+     * @param keyspace        Cassandra keyspace
+     * @param createTableStmt CQL table create statement
+     * @param partitionKeys   list of
+     * @return list of tokens corresponding to each input `partitionKeys`
+     */
+    public List<BigInteger> toTokens(Partitioner partitioner, String keyspace, String createTableStmt, List<List<String>> partitionKeys)
```
yifan-c: Nit: I am wonder if we can expose some sort of `Tokenizer` instead to convert bytebuffer to token? And the tokenizer can be composed with `encodePartitionKey` to handle the string input.

cassandra-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridge.java
```diff
@@ -443,6 +445,198 @@ public ByteBuffer uncompress(ByteBuffer input) throws IOException
...
+     * @param keyspace      keyspace name
+     * @param table         table name
+     * @param ssTable       SSTable instance
+     * @param partitionKeys list of partition keys
+     * @return list of booleans returning true if an SSTable might contain a partition key
+     * (might return false-positives but never false-negatives),
+     * corresponding to the partition keys input in `partitionKeys`.
+     * @throws IOException
+     */
+    public abstract List<Boolean> maybeContains(Partitioner partitioner,
```
yifan-c: Can you add test? 

Another concern I have is that `maybeContains` and `contains` can be confusing. Their implementations differ completely, and they have different performance characteristics. I am wondering if it is more clear to just return a `BloomFilter` object here and client check the with the `BloomFilter`. 

The `BloomFilter` would be a new class defined in the bridge subproject, not the one vended from cassandra library. It could be simply an extension to `Predicate<ByteBuffer>`.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
```diff
@@ -264,6 +276,244 @@ public CompressionUtil compressionUtil()
...
+        return maybeContains(descriptor, ssTable, partitionKeys.stream().map(iPartitioner::decorateKey).collect(Collectors.toList()));
+    }
+
+    private List<Boolean> maybeContains(Descriptor descriptor, SSTable ssTable, Collection<DecoratedKey> partitionKeys) throws IOException
+    {
+        if (partitionKeys.isEmpty())
+        {
+            return List.of();
+        }
+        BloomFilter filter = ReaderUtils.readFilter(ssTable, descriptor);
```
yifan-c: `BloomFilter` is closable, and it is not closed.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/ByteBufferUtils.java
jberragan: `CompositeType` pads with a null byte at the end. I can rename the method.

cassandra-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridge.java
jberragan: What is confusing? I can add a comment regarding performance characteristics. What benefit is there by having another `BloomFilter` class?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/filters/PartitionKeyFilter.java
jberragan: The `filter` method is unused. I added Comparable so we can sort the partition key filters in token order. I think I intended to improve `ReaderUtils.primaryIndexContainsAnyKey` to permit an improved 'any match' using the sorted order instead of performing an `anyMatch` on each filter, but I didn't get around to it in this PR.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
jberragan: `BloomFilter` is closable but it is because it inherits the Cassandra reference counting code from `SharedCloseableImpl`, but we do not use this reference counting code in Analytics . We have seen SIGSEGV errors calling closable on `SharedCloseableImpl` instances which is why `BloomFilter` is not closed anywhere in the code.

yifan-c: I see. Probably drop a comment in the code.

cassandra-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridge.java
yifan-c: The benefit of `class BloomFilter implements Predicate<ByteBuffer>` is to clarify by indicating the underlying data structure. Callsites, upon seeing `BloomFilter`, know its properties already w/o needing to read the javadoc.

That said, I am good with enriching javadoc to clarify the properties of the method.

cassandra-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridge.java
jberragan: It is for convenience, but I agree it doesn't add a great deal.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/reader/CassandraBridgeUtilTests.java
```diff
@@ -169,6 +170,61 @@ public void testOverlaps()
...
+            assertTrue(bridge.contains(partitioner, schema.keyspace, schema.table, ssTable, randomBuffers).stream().noneMatch(aBoolean -> aBoolean));
+            List<ByteBuffer> allKeys = new ArrayList<>();
+            allKeys.addAll(buffers);
+            allKeys.addAll(randomBuffers);
+            List<Boolean> exactResult = bridge.contains(partitioner, schema.keyspace, schema.table, ssTable, allKeys);
+            assertEquals(allKeys.size(), exactResult.size());
+            for (int i = 0; i < exactResult.size(); i++)
+            {
+                assertEquals(i < buffers.size(), exactResult.get(i));
+            }
```
yifan-c, line 239: nit: for this test, how about adding the assertion that given the same list of keys, for each boolean result in the result lists, if `maybeContains` returns false, `contains` must also return false?

yifan-c (commented):
There are few comments not yet addressed. But, LGTM overall.

frankgh (approved):
+1 Looks good to me

jberragan:
Green CI: https://app.circleci.com/pipelines/github/jberragan/cassandra-analytics/159/workflows/6670f141-d2c8-4973-b894-0eae13cbbccf

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/filters/PartitionKeyFilter.java
jberragan: I re-added `filter` method it is used in some Spark 2 tests.

PR #102, opened 2025-03-11
CASSANALYTICS-18: Support CQL duration type

lukasz-antoniak:
Adds support for CQL `duration` type.
JIRA link: [CASSANALYTICS-18](https://issues.apache.org/jira/browse/CASSANALYTICS-18)
Pipeline Link: [circleci.com](https://app.circleci.com/pipelines/github/lukasz-antoniak/cassandra-analytics/2/workflows/f28b4e3d-e2b2-44d9-8b77-4a92a0188913)


cassandra-four-zero-types/build.gradle
```diff
@@ -33,6 +33,7 @@ dependencies {
     compileOnly project(":cassandra-analytics-common")
     compileOnly(project(path: ':cassandra-four-zero', configuration: 'shadow'))
     compileOnly "com.esotericsoftware:kryo-shaded:${kryoVersion}"
+    compileOnly(group: "${sparkGroupId}", name: "spark-core_${scalaMajorVersion}", version: "${project.rootProject.sparkVersion}")
```
jberragan: Could you not pull in Spark here? I think you can work around by moving your `CalendarIntervalSerializer` class to `SparkDuration` and performing the conversion in `toSparkSqlType`. An effort was made last year to decouple the types from Spark, so bulk reads can be performed outside of Spark and also because CDC does not necessarily run in Spark.

jberragan:
Thanks for adding this! My one request is to keep the Spark conversion code in `SparkDuration` to keep `Duration` decoupled from Spark.

yifan-c:
Hi @lukasz-antoniak , thank you for the patch! Can you create a JIRA ticket at https://issues.apache.org/jira/projects/CASSANALYTICS/ if not yet?

cassandra-four-zero-types/build.gradle
lukasz-antoniak: FWIW, I cannot easily move logic of `Duration` to `SparkDuration`, because Cassandra serializer expects `org.apache.cassandra.cql3.Duration`. To achieve it, I would need to add Cassandra dependency to `cassandra-analytics-spark-converter` module. Is that OK?

I am not sure if `toSparkSqlType()` shall return `org.apache.cassandra.cql.Duration`, as this is C* type. I guess I might have misunderstood your suggestion.
```
@Override
public Object toSparkSqlType(@NotNull Object value, boolean isFrozen)
{
    CalendarInterval cl = (CalendarInterval) value;
    return Duration.newInstance(cl.months, cl.days, cl.microseconds * 1000);
}
```
```
Caused by: java.lang.ClassCastException: class org.apache.cassandra.cql3.Duration cannot be cast to class org.apache.spark.unsafe.types.CalendarInterval (org.apache.cassandra.cql3.Duration and org.apache.spark.unsafe.types.CalendarInterval are in unnamed module of loader 'app')
	at org.apache.spark.sql.catalyst.expressions.BaseGenericInternalRow.getInterval(rows.scala:49)
	at org.apache.spark.sql.catalyst.expressions.BaseGenericInternalRow.getInterval$(rows.scala:49)
	at org.apache.spark.sql.catalyst.expressions.GenericInternalRow.getInterval(rows.scala:195)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.processNext(Unknown Source)
	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
```

I have tried to exclude Spark dependency from `cassandra-four-zero-types` module in various ways. Best I could come up with, was to introduce a POJO `CqlDuration` that is able to map internally to `CalendarInterval`. See commit: https://github.com/apache/cassandra-analytics/pull/102/commits/df40c73b4e73728d8455609526314f01e845df13.

jberragan: Could you convert it into a an intermediate JDK type from `Duration`? Even if it's a simple wrapper like:

```
public class DurationWrapper {
    public int months;
    public int days;
    public long microseconds;
}
```

And then convert `DurationWrapper ` to `CalendarInterval` in `SparkDuration`? It just means we can consume Duration without the Spark dependency (and in CDC).

jberragan: Or even `java.time.Duration` if that works?

lukasz-antoniak: Clear. Where shall I convert `CalendarInterval` (generated by Spark) to `DurationWrapper` to correctly support write-path from Spark to Cassandra. We do have `TypeConverter`, but it is used only in the opposite direction (deserialization).

yifan-c: For the write-path, the converters are located at a different place. Check out `org.apache.cassandra.spark.bulkwriter.SqlToCqlTypeConverter`. 
They should be consolidated, but it is what it is as of today. 

A nit on the class name. How about calling it `InternalDuration` instead of `DurationWrapper`? Since it does not wrap anything.

cassandra-four-zero-types/src/main/java/org/apache/cassandra/spark/data/types/Duration.java
```diff
@@ -32,8 +46,148 @@ public String name()
...
     @Override
-    public boolean isSupported()
+    public AbstractType<?> dataType()
+    {
+        return DurationType.instance;
+    }
+
+    public <T> TypeSerializer<T> serializer()
+    {
+        return (TypeSerializer<T>) CalendarIntervalSerializer.INSTANCE;
```
yifan-c: ~I think we can continue using `DurationSerializer.instance`.~

_Update_ 
I tried out `DurationSerializer.instance` and other necessary changes. It does not work for nested type.

cassandra-four-zero-types/build.gradle
yifan-c: Disregard the renaming suggestion. I just refreshed and found out `CqlDuration` is added.

yifan-c (commented):
There are changes I would like to suggest here and there. So I just put them in a single commit, https://github.com/yifan-c/cassandra-analytics/commit/903748c20dc5f5b17d19c99797a508c70801f941
There are mainly 4 items.
1. do not introduce spark dependency to `common` subproject. We want to keep the spark dependency retained within the subprojects named with "spark".
2. refactoring `CqlDuration` to not referencing to `CalendarInterval`. I renamed the class, as _I_ think `CqlDuration` is not really a CQL data type. It is a intermediate data type used in analytics internal only. If disliking the name, please feel free to revert. 
3. `SparkDuration` and `Duration` are updated to convert to/from `InternalDuration`
4. added a TODO at `SqlToCqlTypeConverter` to implement the support at the write path. 

Unfortunately, when converting the CQL duration to Spark duration, it loses precision. It should be called out.

lukasz-antoniak:
@yifan-c, thank you very much for the review! I did not notice your commit and was doing changes myself, but then pushed your stuff to my branch. I think I have addressed point 4, so feel free to review PR once you have time. Can you also check new implementation of `testDuration()` test?

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/type/InternalDuration.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    public String toString()
+    {
+        StringBuilder builder = new StringBuilder();
+        if (this.months < 0 || this.days < 0 || this.nanoseconds < 0L)
+        {
+            builder.append('-');
+        }
```
yifan-c, line 72: CQL duration validate all non-zero fields are all negative or all positive. Since this is consuming from data that is validated by Cassandra already, I think this condition is fine.

yifan-c (approved):
Thanks again! It looks good!

yifan-c:
Did you setup circleCI for the project? If so, please comment the CI link in the JIRA.

yifan-c:
Ah. Just realized one thing. Can you update `changes.txt`?

cassandra-four-zero-types/build.gradle
jberragan: Could you remember to remove this dependency, it seems to compile without it now, after moving to `SparkDuration`.

jberragan:
If you could remember to remove this line:

https://github.com/apache/cassandra-analytics/pull/102/files#diff-a22af2fd084620cc2bec9d43f77ea8d8c050f1c59b316f9970f6b3e2b9d1ec10R36

Otherwise I'm +1 (nb)

lukasz-antoniak:
@jberragan, removed the dependency line. Sorry for missing that before. Pipeline link added. I had to move creation of `CalendarInterval` to `cassandra-analytics-spark-converter` module, because API changes between Spark 2 and 3.

@yifan-c, I have reviewed changes from this PR and they look good to me. Can you have final look? Shall I squash your commit with mine or leave 3 commits in this PR?

yifan-c:
The changes look good to me. However, the CI link you shared does not open. It gives me 404. Can you double check and shared it at https://issues.apache.org/jira/browse/CASSANALYTICS-18? 

Re: squashing. You do not need to squash them. For this repo, the convention is to merge from the PR page directly, using the `Squash and merge` button. (There is one single trunk. so why the trouble of the manual work :p)

lukasz-antoniak:
Updated CI link.

jberragan (approved):
+1 (nb)

frankgh (approved):
+1 thanks for the patch

PR #103, opened 2025-03-11
CASSANALYTICS-16: Bump Cassandra Sidecar version to 0.1.0

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSANALYTICS-16


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/Bundle.java
```diff
@@ -42,7 +42,7 @@ public class Bundle
 {
     private static final String MANIFEST_FILE_NAME = "manifest.json";
 
-    public final BigInteger startToken;
+    public final BigInteger firstToken;
```
yifan-c, line 45: The start tokens in the ranges are usually excluded, since the ranges are open-closed unless called out specifically. The token here is really the first token in the range. Renaming to clarify.

frankgh (approved):
+1

PR #106, opened 2025-04-16
CASSANALYTICS-20: CassandraDataLayer uses configuration list of IPs instead of the full ring/datacenter

serban21:
This includes some changes from #93 , so it's better to merge it only after the other one is merged.

I used `tokenPartitioner.ring().instances()`  instead of making a new request to get ring information since the tokenPartitioner is built just before this. 

An argument can be made that this is not the case for clearing snapshots, since that is happening (much) later. On the other hand if the ring had instances added/removed/replaced in between the new instances would not have the snapshots anyway, and the old instances will be removed.

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
@@ -220,6 +220,7 @@ public class CassandraDataLayer extends PartitionedDataLayer implements StartupV
         this.rfMap = rfMap;
         this.timeProvider = timeProvider;
         this.maybeQuoteKeyspaceAndTable();
+        this.initSidecarClient();
         this.initInstanceMap();
         this.startupValidate();
     }
@@ -234,7 +235,7 @@ public class CassandraDataLayer extends PartitionedDataLayer implements StartupV
 
         // Load cluster config from options
         clusterConfig = initializeClusterConfig(options);
-        initInstanceMap();
+        initSidecarClient();
 
         // Get cluster info from Cassandra Sidecar
         int effectiveNumberOfCores;
@@ -251,6 +252,7 @@ public class CassandraDataLayer extends PartitionedDataLayer implements StartupV
         {
             throw new RuntimeException(ThrowableUtils.rootCause(exception));
         }
+        initInstanceMap();
         LOGGER.info("Initialized Cassandra Bulk Reader with effectiveNumberOfCores={}", effectiveNumberOfCores);
     }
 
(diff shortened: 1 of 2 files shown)


frankgh:
@serban21 #93 has been merged, and should unblock you now. Can you rebase?

serban21:
> @serban21 #93 has been merged, and should unblock you now. Can you rebase?

Done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
```diff
@@ -115,6 +121,7 @@ protected ClientConfig(Map<String, String> options)
         this.snapshotName = MapUtils.getOrDefault(options, SNAPSHOT_NAME_KEY, "sbr_" + UUID.randomUUID().toString().replace("-", ""));
         this.datacenter = options.get(MapUtils.lowerCaseKey(DC_KEY));
         this.createSnapshot = MapUtils.getBoolean(options, CREATE_SNAPSHOT_KEY, true);
+        this.createSnapshotFilterDistinctInstances = MapUtils.getBoolean(options, CREATE_SNAPSHOT_FILTER_DISTINCT_INSTANCES_KEY, true);
```
yifan-c, line 124: `CREATE_SNAPSHOT_FILTER_DISTINCT_INSTANCES_KEY` was discussed and removed in #93

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
```diff
@@ -422,7 +424,23 @@ public boolean readIndexOffset()
...
+            instanceMap = tokenPartitioner.ring().instances().stream()
+                    .filter(instance -> datacenter == null || datacenter.equals(instance.dataCenter()))
+                    .map(instance -> instance.nodeName()).distinct()
+                    .map(nodeName -> new SidecarInstanceImpl(nodeName, sidecarClientConfig.effectivePort()))
+                    .collect(Collectors.toMap(SidecarInstance::hostname, Function.identity()));
+        }
+        else
+        {
+            instanceMap = Collections.emptyMap();
+        }
```
yifan-c, line 438: Should it fail early if `tokenPartitioner` is `null`? It is unexpected to have the `null` value. 
Alternatively, the `if-else` can be removed and only keep the `instanceMap` assignment.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
```diff
@@ -422,7 +424,23 @@ public boolean readIndexOffset()
 
     protected void initInstanceMap()
     {
-        instanceMap = clusterConfig.stream().collect(Collectors.toMap(SidecarInstance::hostname, Function.identity()));
+        if (tokenPartitioner != null)
+        {
+            instanceMap = tokenPartitioner.ring().instances().stream()
+                    .filter(instance -> datacenter == null || datacenter.equals(instance.dataCenter()))
+                    .map(instance -> instance.nodeName()).distinct()
```
yifan-c, line 431: Can we consider the port as well for the uniqueness check? Just the same as line#353.

yifan-c (commented):
Please update `CHANGES.txt` and share the circleCI link. You can sign up a free account using your github account (, which _I think_ is the easiest among all the the sign up options).

yifan-c (commented):
Forgot to comment yesterday.. It would be nice to have an integration test that sets `sidecar_contact_points` with one single IP and the job should still work on reading data.

yifan-c:
I rebased the patch and triggered CI (green).
The rebased branch: https://github.com/yifan-c/cassandra-analytics/tree/r/apache/pr106
CI: https://app.circleci.com/pipelines/github/yifan-c/cassandra-analytics?branch=r%2Fapache%2Fpr106

yifan-c:
Closing in favor of https://github.com/apache/cassandra-analytics/pull/122
Thanks @serban21 for the patch!

PR #107, opened 2025-04-16
CASSANALYTICS-5: Bump Cassandra version to 4.0.17 and integrate with …

yifan-c:
…CQLSSTableWriter

Patch by Yifan Cai; Reviewed by TBD for CASSANALYTICS-5

### DIFF
--- a/cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/bridge/SSTableWriterImplementation.java
+++ b/cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/bridge/SSTableWriterImplementation.java
@@ -19,25 +19,15 @@
 
 package org.apache.cassandra.bridge;
 
-import java.io.Closeable;
 import java.io.IOException;
-import java.nio.file.DirectoryStream;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.util.HashSet;
+import java.util.Collection;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import org.apache.cassandra.config.Config;
 import org.apache.cassandra.dht.IPartitioner;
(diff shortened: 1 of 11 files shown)


scripts/build-dtest-jars.sh
```diff
@@ -31,14 +31,14 @@ else
...
-  #   "cassandra-4.0:cassandra-4.0.12"
+  #   "cassandra-4.0:cassandra-4.0.17"
    # a hash that points to a commit on the cassandra-4.0 branch
   #   "cassandra-4.0:1f79c8492528f01bcc5f88951a1cc9e0d7265c54"
   # the cassandra-4.0 branch - used for nightly integration test runs or local testing of new features
   #   "cassandra-4.0:cassandra-4.0"
   # Due to MacOS being stuck on Bash < 4, we don't use associative arrays here.
   CANDIDATE_BRANCHES=(
-    "cassandra-4.0:cassandra-4.0.12"
+    "cassandra-4.0:cassandra-4.0.17"
```
frankgh, line 41: unrelated, but for 4.1 should we use the latest released version (4.1.8)?

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/bridge/SSTableWriterImplementation.java
```diff
@@ -69,94 +56,56 @@ public SSTableWriterImplementation(String inDirectory,
...
-            }
-        }
+    private void onSSTablesProduced(Collection<SSTableReader> sstables)
+    {
+        Objects.requireNonNull(producedSSTablesListener, "producedSSTablesListener is not set");
+        Set<SSTableDescriptor> sstableDescriptors = sstables
+                                                    .stream()
+                                                    .map(sstable -> {
+                                                        String baseFilename = baseFilename(sstable.descriptor);
+                                                        // TODO: for now, the sstableReader is closed immediately,
```
frankgh, line 95: should we do this as part of this PR? I'd argue that this is desirable. Any reason to not do it here?

frankgh (commented):
looks good in general. I left a couple of comments

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/bridge/SSTableWriterImplementation.java
yifan-c: I left the todo because the change would be more invasive and I feel they are unrelated. The step does not simply re-read the data back, but also calculates digests, etc. Some refactoring will be required.

frankgh: maybe worth filing a jira to keep better track of this change?

frankgh (approved):
+1 looks good!

PR #108, opened 2025-04-22
CASSANALYTICS-30: Add Sidecar Client

bbotella:
Adding sidecar client to remove the artifact dependency with the Apache Sidecar artifact.


cassandra-analytics-sidecar-client/src/main/java/org/apache/cassandra/clients/Sidecar.java
```diff
@@ -32,19 +32,19 @@
...
-import o.a.c.sidecar.client.shaded.common.response.NodeSettings;
-import o.a.c.sidecar.client.shaded.io.vertx.core.Vertx;
-import o.a.c.sidecar.client.shaded.io.vertx.core.VertxOptions;
+import io.vertx.core.Vertx;
+import io.vertx.core.VertxOptions;
+import org.apache.cassandra.sidecar.common.client.SidecarInstance;
+import org.apache.cassandra.sidecar.common.client.SidecarInstanceImpl;
+import org.apache.cassandra.sidecar.common.response.NodeSettings;
+import org.apache.cassandra.sidecar.vertxclient.VertxHttpClient;
+import org.apache.cassandra.sidecar.vertxclient.VertxRequestExecutor;
```
yifan-c: I think we still want to shade the sidecar client subproject and reference to the shadow jar. The idea is to embedding the sidecar client as a whole (shadow jar) w/o conflicting with any existing dependencies in analytics and spark. 

Just like what we already do for cassandra-all. See `:cassandra-four-zero`

frankgh:
We will probably need to add to the https://github.com/apache/cassandra-analytics/blob/trunk/LICENSE.txt noting that the code is copied from Sidecar. Please take a look at the bottom of LICENSE.txt

cassandra-analytics-core/build.gradle
```diff
@@ -77,8 +77,7 @@ dependencies {
     runtimeOnly(group: 'net.java.dev.jna', name: 'jna', version: "${jnaVersion}")
     runtimeOnly(group: 'net.java.dev.jna', name: 'jna-platform', version: "${jnaVersion}")
 
-    // This dependency must be built by running `scripts/build-dependencies.sh`
-    api(group: "${sidecarClientGroup}", name: "${sidecarClientName}", version: "${sidecarVersion}")
```
frankgh, line 81: there's clean up to be  done for these variables. I see it in the profiles + settings.gradle. It'd be useful to clean that up as part of this PR, since we are removing usage of them here

PR #109, opened 2025-05-01
CASSANALYTICS-49: Support UDTs inside collections

skoppu22:
Currently Cassandra bulk writer not properly supporting UDTs inside collections. Also complex structures like collections inside UDTs are not supported properly. Adding support for these cases.

Note: Tuples with UDTs are not yet supported properly, it will be taken care in another Jira/PR

Circle CI passed: https://app.circleci.com/pipelines/github/skoppu22/cassandra-analytics/19/workflows/187a5c65-b03f-495e-b29b-914dd0483642


cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/testing/TestUtils.java
```diff
@@ -36,7 +36,7 @@ public final class TestUtils
     private static final AtomicInteger TEST_TABLE_ID = new AtomicInteger(0);
 
     //    public static final int ROW_COUNT = 10_000;
-    public static final int ROW_COUNT = 1_000;
+    public static final int ROW_COUNT = 100;
```
yifan-c: can you revert the change? I assume you lower the row count for local testing.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -389,19 +391,52 @@ private Map<String, Object> getBindValuesForColumns(Map<String, Object> map, Str
...
+        if (value instanceof Map)
+        {
+            Map<Object, Object> resultMap = new HashMap<>();
+            for (Map.Entry<?, ?> entry : ((Map<?, ?>) value).entrySet())
+            {
+                resultMap.put(maybeConvertUdt(entry.getKey()), maybeConvertUdt(entry.getValue()));
+            }
+
+            return resultMap;
+        }
```
yifan-c: those 3 branches will copy the list unnecessarily if they do _not_ contain UDT in the table schema. It would produce a lot of short-lived objects. Note that there are millions or more rows to write. 

One idea to skip the unnecessary conversion is to check the table schema. 
1. `RecordWriter` has access to the table schema, e.g. `CqlTable`. 
2. `CqlTable` can mark the columns where contain UDT. 
3. Instead of calling `maybeConvertUdt` on every column in each row, first check whether `maybeConvertUdt` should be invoked for this column. Now, since you know which columns contain UDT, you can even drop "maybe" from the method name. It would clarify the code too.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteUdtTest.java
```diff
@@ -88,17 +141,298 @@ void testWriteWithNestedUdt()
...
+
+        // test null cases
+        cluster.schemaChangeIgnoringStoppedInstances(String.format("insert into %s (id) values (%d)",
+                SET_OF_UDT_SOURCE_TABLE, i++));
+        cluster.schemaChangeIgnoringStoppedInstances(String.format("insert into %s (id, udtset) values (%d, null)",
+                SET_OF_UDT_SOURCE_TABLE, i++));
+        cluster.schemaChangeIgnoringStoppedInstances(String.format("insert into %s (id, udtset) values (%d, " +
+                        "{{f1:null, f2:null, f3:null}})",
+                SET_OF_UDT_SOURCE_TABLE, i++));
+
```
yifan-c: `schemaChangeIgnoringStoppedInstances` is to change schema. It is not a proper use of the api to insert data. 

I think you want to use `cluster.coordinator(1).executeWithResult(QUERY)` instead.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteUdtTest.java
```diff
@@ -32,7 +32,7 @@
 import org.apache.spark.sql.SparkSession;
 import org.jetbrains.annotations.NotNull;
 
-import static org.apache.cassandra.testing.TestUtils.DC1_RF3;
+import static org.apache.cassandra.testing.TestUtils.DC1_RF1;
```
yifan-c: Curious about the reason of the change? If unnecessary, please revert.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteUdtTest.java
```diff
@@ -59,6 +59,59 @@ class BulkWriteUdtTest extends SharedClusterSparkIntegrationTestBase
...
+            " (innerudt map<frozen<" + TWO_FIELD_UDT_NAME + ">, frozen<" + TWO_FIELD_UDT_NAME + ">>);";
+    public static final QualifiedName UDT_WITH_MAP_OF_UDT_SOURCE_TABLE = new QualifiedName(TEST_KEYSPACE, "udt_with_map_of_udt_src");
+    public static final QualifiedName UDT_WITH_MAP_OF_UDT_DEST_TABLE = new QualifiedName(TEST_KEYSPACE, "udt_with_map_of_udt_dest");
+
+    // Table with UDT which contains either a list or set or map of UDTs inside it
+    public static final String UDT_WITH_COLLECTION_OF_UDT_TABLE_CREATE = "CREATE TABLE %s.%s (\n"
+            + "            id BIGINT PRIMARY KEY,\n"
+            + "            outerudt frozen<%s>)";
+
+
```
yifan-c, line 118: There is no test covering the scenario of UDTs in tuple. Are you going to update this patch or address in a follow-up? (I am OK with either of them)

yifan-c (commented):
Sorry for the delay. I thought I have reviewed, but apparently not :p

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/testing/TestUtils.java
skoppu22: Some tests were timing out when ran on local, especially when ran from Intellij, so had to reduce it. Can we just continue with lesser number of rows so we don't need to change this on local everytime we run?

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteUdtTest.java
skoppu22: This should be reverted, was needed only for local testing

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteUdtTest.java
skoppu22: As I mentioned in the PR description, there is lot more to be done to handle Tuples, support for Tuples will come in another PR/task

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteUdtTest.java
skoppu22: Done

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteUdtTest.java
skoppu22: - schemaChangeIgnoringStoppedInstances also uses executeWithResult internally, but it uses Monitor additionally. 
- Changed to use execute for insert and select queries

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
skoppu22: - Updated CqlTable to maintain columns with UDTs and check it before calling maybeConvertUdt
- There’s still a chance of creating unnecessary lists, sets, or maps, like udt<list<udt<set<map>>>>. We create new sets and maps without knowing if there are more UDTs inside. I tried counting UDTs in a column (inside determineColumnsWithUdts) to avoid this, but it may not work for all cases. It doesn’t differentiate when the same UDT type is used multiple times, like in map<udtType1, udtType1>, where the UDT count is 1 instead of 2. This is fine for Maps, but it could break things in non-Map cases.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlTable.java
```diff
@@ -50,6 +51,7 @@ public class CqlTable implements Serializable
     private final Set<CqlField.CqlUdt> udts;
 
     private final Map<String, CqlField> fieldsMap;
+    private final Map<String, Boolean> columnsWithUdts;
```
yifan-c: 👍 
An minor improvement is to a `Set` instead. Only the columns referencing UDT are included in the set.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlTable.java
```diff
@@ -242,6 +246,31 @@ public int indexCount()
         return indexCount;
     }
 
+    /**
+     * Check each column of the table for UDT type somewhere nested inside it and maintain map of true/false accordingly
+     * @return map of column name to true/false
+     */
```
yifan-c: 👍 thanks for adding javadocs

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlTable.java
```diff
@@ -242,6 +246,31 @@ public int indexCount()
...
+
+        return columnsWithUdts;
+    }
+
+    /**
+     * Determines if a column has UDT type somewhere nested inside it
+     * @param fieldName name of the column
+     * @return true if the column has UDT type , false otherwise
+     */
+    public boolean containsUdt(String fieldName)
```
yifan-c, line 273: nit: to be consistent with Cassandra and to better describe the column, how about naming it as `referencingUserType`?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -380,28 +382,69 @@ private Map<String, Object> getBindValuesForColumns(Map<String, Object> map, Str
...
+            }
+            else
+            {
+                map.put(columnNames[i], values[i]);
+            }
         }
         return map;
     }
 
     private Object maybeConvertUdt(Object value)
```
yifan-c, line 397: It would be nice to add java doc to describe the method, e.g. why it checks for collections in the implementation.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -380,28 +382,69 @@ private Map<String, Object> getBindValuesForColumns(Map<String, Object> map, Str
...
+            {
+                resultList.add(maybeConvertUdt(entry));
+            }
+
+            return resultList;
+        }
+
+        if (value instanceof Set)
+        {
+            Set<Object> resultList = new HashSet<>();
```
yifan-c: nit: `resultSet`?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
yifan-c: having columns referrencing UDTs flagged is good enough. Thank you for adding that!

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/testing/TestUtils.java
yifan-c: I ran the tests after reverting `ROW_COUNT`, cannot reproduce timing out. Maybe it was related to the deep copy for all columns (which is now resolved). Can you revert the change?

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteUdtTest.java
```diff
@@ -59,6 +60,61 @@ class BulkWriteUdtTest extends SharedClusterSparkIntegrationTestBase
                                                      + "           id BIGINT PRIMARY KEY,\n"
                                                      + "           nested " + NESTED_FIELD_UDT_NAME + ");";
 
+    // UDT with list, set and map in it
+    public static final String UDT_WITH_COLLECTIONS_TYPE_NAME = "udt_with_collections";
+    public static final String UDT_WITH_COLLECTIONS_TYPE_CREATE = "CREATE TYPE " + TEST_KEYSPACE + "." + UDT_WITH_COLLECTIONS_TYPE_NAME +
+            " (f1 list<text>, f2 set<int>, f3 map<int, text>);";
+
```
yifan-c, line 68: It would be nice to use QT to fuzz test the different combinations of types. Just say it out load, no need to address.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -380,28 +382,69 @@ private Map<String, Object> getBindValuesForColumns(Map<String, Object> map, Str
     {
         Preconditions.checkArgument(values.length == columnNames.length,
                                     "Number of values does not match the number of columns " + values.length + ", " + columnNames.length);
+
         for (int i = 0; i < columnNames.length; i++)
         {
-            map.put(columnNames[i], maybeConvertUdt(values[i]));
+            if (cqlTable().containsUdt(columnNames[i]))
```
frankgh: can we extract the method invocation for `cqlTable` and assign it to a variable outside of the for loop?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -380,28 +382,69 @@ private Map<String, Object> getBindValuesForColumns(Map<String, Object> map, Str
...
+            }
         }
         return map;
     }
 
     private Object maybeConvertUdt(Object value)
     {
+        if (value instanceof List)
+        {
+            List<Object> resultList = new ArrayList<>();
```
frankgh, line 401: minor, avoid creating a list when the source is empty. ie. if !value.isEmtpy() -> process

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -380,28 +382,69 @@ private Map<String, Object> getBindValuesForColumns(Map<String, Object> map, Str
...
+            for (Object entry : (List<?>) value)
+            {
+                resultList.add(maybeConvertUdt(entry));
+            }
+
+            return resultList;
+        }
+
+        if (value instanceof Set)
```
frankgh: minor , avoid creating a set when source set is empty

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -380,28 +382,69 @@ private Map<String, Object> getBindValuesForColumns(Map<String, Object> map, Str
...
+            Set<Object> resultList = new HashSet<>();
+            for (Object entry : (Set<?>) value)
+            {
+                resultList.add(maybeConvertUdt(entry));
+            }
+
+            return resultList;
+        }
+
+        if (value instanceof Map)
```
frankgh: minor avoid creating a map when the source map is empty

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/testing/TestUtils.java
frankgh: this change should be reverted. Are you seeing issues after the new change?

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/SharedClusterSparkIntegrationTestBase.java
```diff
@@ -171,28 +173,35 @@ public void checkSmallDataFrameEquality(Dataset<Row> expected, Dataset<Row> actu
...
-        queriedData.forEach(row -> actualEntries.add(rowFormatter.apply(row)));
+        Set<String> driverEntries = new HashSet<>();
+        driverData.forEach(row -> driverEntries.add(driverRowFormatter
+                .apply(row)
+                // Driver Codec writes "NULL" for null value. Spark DF writes "null".
+                .replace("NULL", "null")
+                // driver writes lists as [] and sets as {},
+                // whereas spark entries have the same type WrappedArray for both lists and sets
+                .replace('[', '{')
+                .replace(']', '}')));
```
frankgh: I think this logic should be encapsulated in the function itself, and not here.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlTable.java
skoppu22: We use the word UDT extensively all over the repo. Better to keep the name containsUdt to be inline with remaining code.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
skoppu22: cqlTable() already uses a class local variable `cqlTable`, populates it only once if it is not set and references the same in subsequent invocations. Using a local variable here in this caller only avoids the function call, but compiler must be smart enough to avoid subsequent calls I believe. Also `cqlTable` is volatile, there must be a reason for making it volatile, so better we reference the same. Infact we can reference that class local variable `cqlTable` directly here by calling cqlTable() only once in the class init, but I wouldn't do that without understanding why it was implemented this way previously.

skoppu22: I see @yifan-c added cqlTable. @yifan-c Can we call cqlTable() only once in the class init and reference the class local variable in rest of the class? I see `cqlTable` is volatile, but it is assigned only once when it is null. Does it need to be really volatile?

cassandra-analytics-integration-framework/src/main/java/org/apache/cassandra/testing/TestUtils.java
skoppu22: I am seeing quite slower runs, sometimes timeouts only when running under Intellij. Reverted this change.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/SharedClusterSparkIntegrationTestBase.java
skoppu22: Which function do you mean? We repeat the same logic for all driver entries formatters for UDTs (and also Tuples which is coming up today). So keeping it in one place instead of repeating the same in each formatter.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlTable.java
yifan-c: Sounds good. Given that you added the detailed javadoc (awesome!), I have no concern.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
frankgh: I think we want to avoid having a call to a method in a tight loop. So it's better to get a reference of this variable before the loop

yifan-c: > Infact we can reference that class local variable cqlTable directly here by calling cqlTable() only once in the class init
This sounds good. Given that the writer requires cqlTable to be initialized, the initialization can be done eagerly.

yifan-c:
Please update `CHANGES.txt`

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/SharedClusterSparkIntegrationTestBase.java
skoppu22: Moved out collections specific formatting to corresponding formatter, to avoid other tests getting impacted by chance

CHANGES.txt
```diff
@@ -1,4 +1,5 @@
 1.0.0
+ * Bulk writer needs to support collections with UDTs
```
frankgh: can we add the ticket here? 
```suggestion
 * Bulk writer needs to support collections with UDTs (CASSANALYTICS-49)
```

frankgh (approved):
+1 LGTM

PR #110, opened 2025-05-09
CASSANALYTICS-27: Support for Trie-based SSTables

lukasz-antoniak:
Status: Work in progress. Contributions welcome!

Although tests passes and users can read V5 SSTables, there are various things to fix on a code level:

- [ ] `FileSystemSSTable` or any SSTable should support re-buffering on random position to allow BTI partition index reading.
- [ ] Code uses `@VisibleForTesting` methods from C* API in `BtiReaderUtils.createFileHandle`. BTI reading components require `FileHandle` as an argument, and I am trying to create one based on `InputStream` that we receive from bridge interface.
- [ ] BTI index readers (e.g. `BtiReaderUtils`) do not expose statistics as number of bytes skipped. Due to characteristic of BTI index reading, I am not sure if this metric becomes useful.
- [x] Enable again v4 bridges. My IntelliJ had issues with debugging when same class was present in two Cassandra JARs.
- [ ] Tests encountered [CASSANDRA-19902](https://github.com/apache/cassandra-analytics/commit/384f150bf17756ed1614f7ae3170fcad2644ffa3). I have applied a patch, but had to link dtest JAR creation to my own Cassandra fork.

All places where I was no sure about the changes, are marked as `TODO(c4c5)`.


cassandra-analytics-common/src/main/java/org/apache/cassandra/cdc/api/CommitLog.java
```diff
@@ -54,7 +54,7 @@ static Optional<Pair<Integer, Long>> extractVersionAndSegmentId(@NotNull final S
             try
             {
                 final int version = matcher.group(2) == null ? 6 : Integer.parseInt(matcher.group(2));
-                if (version != 6 && version != 7)
+                if (version != 6 && version != 7 && version != 8) // TODO(c4c5): What is the difference between values?
```
yifan-c: the values correspond to the commit log version in Cassandra. 
You might already find it out. Sharing just in case. They can be found at https://github.com/apache/cassandra/blob/701495afec90a9ce0e01430be6927b729a12ec72/src/java/org/apache/cassandra/db/commitlog/CommitLogDescriptor.java#L63-L66 
version:8 is added in this patch https://issues.apache.org/jira/browse/CASSANDRA-14227

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/FileSystemSource.java
```diff
@@ -90,28 +90,30 @@ public void request(long start, long end, StreamConsumer consumer)
...
+                if (read >= 0)
                 {
                     consumer.onRead(StreamBuffer.wrap(bytes));
                     consumer.onEnd();
                 }
-                else
-                {
-                    close = true;
-                }
+                // TODO(c4c5): Make configurable, so that remote file access does not accidentally close file.
```
yifan-c: Yes. This TODO makes sense. The prior behavior is only reading the file sequentially. But for the partitions trie, random access is required. We do not want to close the file prematurely.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/FileSystemSSTable.java
```diff
@@ -63,9 +61,11 @@ protected InputStream openInputStream(FileType fileType)
         }
         try
         {
-            return useBufferingInputStream
-                   ? new BufferingInputStream<>(new FileSystemSource<>(this, fileType, filePath), stats.get())
-                   : new BufferedInputStream(new FileInputStream(filePath.toFile()));
+            // TODO(c4c5): Any input stream returned here should support re-buffering at random position.
+//            return useBufferingInputStream
```
yifan-c: +1 on this

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/CqlUtils.java
```diff
@@ -166,7 +166,7 @@ public static String extractCleanedTableSchema(@NotNull String createStatementTo
                                                    @NotNull String keyspace,
                                                    @NotNull String table)
     {
-        Pattern pattern = Pattern.compile(String.format("CREATE TABLE ?\"?%s?\"?\\.{1}\"?%s\"?[^;]*;", keyspace, table));
+        Pattern pattern = Pattern.compile(String.format("CREATE TABLE (IF NOT EXISTS)? ?\"?%s?\"?\\.{1}\"?%s\"?[^;]*;", keyspace, table));
```
yifan-c, line 169: The pattern aimed for processing the describe table output, which should not contain the "IF NOT EXISTS" portion. 
But, the change is a good addition and improves robustness.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/streaming/BufferingInputStream.java
```diff
@@ -100,6 +100,21 @@ public BufferingInputStream(CassandraFileSource<T> source, BufferingInputStreamS
...
+        this(source, stats);
+        this.rangeStart = position;
+        this.bytesRead = position == 0 ? 0 : (position + 1);
+        this.bytesWritten.set(this.bytesRead);
+    }
+
+    public BufferingInputStream<T> reBuffer(long position)
+    {
+        return new BufferingInputStream<>(source, stats, position);
+    }
```
yifan-c: yep. It is needed for random file access for BTI

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/streaming/BufferingInputStream.java
```diff
@@ -237,8 +252,8 @@ public void onRead(StreamBuffer buffer)
         {
             return;
         }
-        bytesWritten.addAndGet(length);
         queue.add(buffer);
+        bytesWritten.addAndGet(length);
```
yifan-c: Just curious, why relocating the line?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/streaming/BufferingInputStream.java
```diff
@@ -346,24 +361,28 @@ else if (count <= bytesBuffered())
      * @param buffer the ByteBuffer
      * @throws EOFException if attempts to read beyond the end of the file
      * @throws IOException  for failure during I/O
+     * @return number of bytes read
      */
-    public void read(ByteBuffer buffer) throws IOException
+    public int read(ByteBuffer buffer) throws IOException
     {
+        int read = 0; // Cassandra 4.x vs 5.x
```
yifan-c, line 370: Not sure if I understand the comment. It is only a temporary marker when developing.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/streaming/BufferingInputStream.java
```diff
@@ -346,24 +361,28 @@ else if (count <= bytesBuffered())
...
         for (int remainingLength = buffer.remaining(); 0 < remainingLength; remainingLength = buffer.remaining())
         {
             if (checkState() < 0)
             {
-                throw new EOFException();
+                break;
```
yifan-c: Should it still signal EOF when it has reached EOF?

cassandra-analytics-integration-framework/build.gradle
```diff
@@ -38,7 +38,9 @@ publishing {
     }
 }
 
-ext.dtestJar = System.getenv("DTEST_JAR") ?: "dtest-4.1.4.jar" // latest supported Cassandra build is 4.1
+// ext.dtestJar = System.getenv("DTEST_JAR") ?: "dtest-4.1.4.jar" // latest supported Cassandra build is 4.1
+// TODO(c4c5): Make configurable?
```
yifan-c: It is already configurable via the environment variable, right?

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/CompressionMetadata.java
```diff
@@ -0,0 +1,134 @@
...
+        int chunkLength = inData.readInt();
+        int minCompressRatio = 2147483647;
+        if (hasCompressedLength)
+        {
+            minCompressRatio = inData.readInt();
+        }
+
+        CompressionParams params = new CompressionParams(compressorName, chunkLength, minCompressRatio, options);
+        // TODO(c4c5): CRC check change gone?
+        // params.setCrcCheckChance(AbstractCompressionMetadata.CRC_CHECK_CHANCE);
```
yifan-c: See https://issues.apache.org/jira/browse/CASSANDRA-18872
crc check chance is moved to table params

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/BtiIndexReader.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.cassandra.spark.reader.common.IIndexReader;
+import org.apache.cassandra.spark.sparksql.filters.SparkRangeFilter;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class BtiIndexReader implements IIndexReader
+{
+    private TokenRange ssTableRange = null;
+
+    public BtiIndexReader(@NotNull SSTable ssTable,
```
yifan-c, line 48: According to [the spec for partition index and row index](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/io/sstable/format/bti/BtiFormat.md#partition-index), the nodes page is fixed to 4096. You may want to hard-code the chuck size to it when opening the buffering input stream.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/SSTable.java
```diff
@@ -63,7 +63,13 @@ public InputStream openSummaryStream()
     @Nullable
     public InputStream openPrimaryIndexStream()
     {
-        return openInputStream(FileType.INDEX);
+        return openInputStream(isBigFormat() ? FileType.INDEX : FileType.PARTITIONS_INDEX); // Cassandra 4.x vs 5.x
+    }
+
+    @Nullable
+    public InputStream openSecondaryIndexStream()
```
yifan-c: How about just call it `openBtiRowIndexStream` to avoid any confusion with secondary index feature in Cassandra?

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/BtiIndexReader.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.cassandra.spark.data.IncompleteSSTableException;
+import org.apache.cassandra.spark.data.SSTable;
+import org.apache.cassandra.spark.reader.common.IIndexReader;
+import org.apache.cassandra.spark.sparksql.filters.SparkRangeFilter;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class BtiIndexReader implements IIndexReader
+{
+    private TokenRange ssTableRange = null;
```
yifan-c, line 46: I think it is always `null`. We want to initialize the range value according to partitions.db, which record the first and last token of the sstable.

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/SSTableReader.java
```diff
@@ -0,0 +1,898 @@
...
+            {
+                // BTI format
+                try (InputStream primaryIndex = ssTable.openPrimaryIndexStream())
+                {
+                    File file = new File(ssTable.getDataFileName());
+                    long size = ssTable.length(FileType.PARTITIONS_INDEX);
+                    try (ReadOnlyInputStreamFileChannel fileChannel = new ReadOnlyInputStreamFileChannel((BufferingInputStream<?>) primaryIndex, size);
+                         ChannelProxy proxy = new ChannelProxy(file, fileChannel);
+                         FileHandle fileHandle = new FileHandle.Builder(file).complete(f -> proxy);
+                         PartitionIndex partitionIndex = PartitionIndex.load(fileHandle, metadata.partitioner, false);
```
yifan-c, line 508: Disabling preloading makes sense to avoid downloading data over the wire. Maybe add a comment for posterity.

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/SSTableReader.java
```diff
@@ -0,0 +1,898 @@
...
+                            startOffset = 0L;
+                        }
+                        else if (startOffsetCandidate < 0)
+                        {
+                            // we got the offset from data file
+                            startOffset = ~startOffsetCandidate;
+                        }
+                    }
+                }
+            }
```
yifan-c: I am thinking of this change, for the following reasons:
1. Unify the handling when `startOffset == null` for both BIG and BTI.
2. In BTI branch -- since only Murmur3 and Random partitioners are supported, we can create the corresponding token directly, w/o ser/deser via string value.
3. In BTI branch -- the filtering token range is open-closed. Its start token is excluded. Therefore, `ceiling`/immediate greater branch of the max keybound of the excluded token should be the key included by the token range. 
4. Still BTI branch -- added the error handling for the invalid start offset found. 
```suggestion
           if (indexSummary != null)
            {
                // BIG format
                // If start is null we failed to find an overlapping token in the Index.db file,
                // this is unlikely as we already pre-filter the SSTable based on the start-end token range.
                // But in this situation we read the entire Data.db file to be safe, even if it hits performance.
                startOffset = IndexDbUtils.findDataDbOffset(indexSummary, range, metadata.partitioner, ssTable, stats);
            }
            else
            {
                // BTI format
                try (InputStream primaryIndex = ssTable.openPrimaryIndexStream())
                {
                    File file = new File(ssTable.getDataFileName());
                    long size = ssTable.length(FileType.PARTITIONS_INDEX);
                    try (ReadOnlyInputStreamFileChannel fileChannel = new ReadOnlyInputStreamFileChannel((BufferingInputStream<?>) primaryIndex, size);
                         ChannelProxy proxy = new ChannelProxy(file, fileChannel);
                         FileHandle fileHandle = new FileHandle.Builder(file).complete(f -> proxy);
                         PartitionIndex partitionIndex = PartitionIndex.load(fileHandle, metadata.partitioner, false);
                         PartitionIndex.Reader reader = partitionIndex.openReader())
                    {
                        Token startToken = null;
                        if (metadata.partitioner instanceof Murmur3Partitioner)
                        {
                            startToken = new Murmur3Partitioner.LongToken(range.lowerEndpoint().longValue());
                        }
                        else
                        {
                            startToken = new RandomPartitioner.BigIntegerToken(range.lowerEndpoint());
                        }
                        Token.KeyBound maxKeyBound = startToken.maxKeyBound();
                        startOffset = reader.ceiling(maxKeyBound, (position, assumeNoMatch, partitionPosition) -> position);
                        if (startOffset != null)
                        {
                            // we got the offset from data file
                            startOffset = ~startOffset;
                            if (startOffset < 0)
                            {
                                LOGGER.error("Found invalid start offset in Data.db. Read from the begining of the file.");
                                startOffset = null;
                            }
                        }

                    }
                }
            }

            if (startOffset == null)
            {
                LOGGER.error("Failed to find Data.db start offset, performance will be degraded sstable='{}'", ssTable);
            }
```

cassandra-five-zero-types/src/main/java/org/apache/cassandra/spark/reader/ComplexTypeBuffer.java
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * Pack the cell ByteBuffers into a single ByteBuffer using Cassandra's packing algorithm.
+     * It is similar to {@link #build()}, but encoding the data differently.
+     *
+     * @return a single ByteBuffer with all cell ByteBuffers encoded.
+     */
+    public ByteBuffer pack()
+    {
+        // See CollectionSerializer.deserialize for why using the protocol v3 variant is the right thing to do.
+        // TODO(c4c5): Do we need to specify ProtocolVersion.V3 as last parameter?
```
yifan-c: The ProtocolVersion can be removed safely. See https://issues.apache.org/jira/browse/CASSANDRA-18114
Can you remove the comment at line#109 (that comes from the four-zero subproject)? To avoid any confusion.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/streaming/BufferingInputStream.java
lukasz-antoniak: Correct, just a temporal marker. Wanted to indicate that we are returning now number of bytes read.

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/BtiIndexReader.java
lukasz-antoniak: Just to be sure, do you mean chunk size [here](https://github.com/lukasz-antoniak/cassandra-analytics/blob/v5/cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/streaming/BufferingInputStream.java#L194)?

cassandra-analytics-integration-framework/build.gradle
lukasz-antoniak: Comment was more about adding a CI integration run for 5.0.3 dtest. Doing it now.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/FileSystemSource.java
lukasz-antoniak: I have introduced a parameter to auto close the file, which is true for BIG and false for BTI.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/FileSystemSSTable.java
lukasz-antoniak: I have introduced the logic based on `isBtiFormat()` result. If `useBufferingInputStream || isBtiFormat()` is true, we enforce usage of `BufferingInputStream`.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/streaming/BufferingInputStream.java
lukasz-antoniak: I cannot remember now and cannot reproduce any potential issue. Reverted the change.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/streaming/BufferingInputStream.java
lukasz-antoniak: I have changed the logic and preserved EOF error.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/SSTable.java
lukasz-antoniak: Can we call it `openRowIndexStream`? We did not add `big` or `bti` to open methods.

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/CompressionMetadata.java
```diff
@@ -0,0 +1,138 @@
...
+        // TODO(c4c5): CRC check change gone?
+        // return parameters.getCrcCheckChance();
+        return 1.0;
+    }
+
+    // TODO(c4c5): Find better way to create Cassandra CompressionMetadata.
+    public org.apache.cassandra.io.compress.CompressionMetadata toInternal(File file, long compressedFileLength)
+    {
+        // TODO(c4c5): Duplicated memory allocation for Memory and BigLongArray.
+        Memory chunkOffsetsMem = Memory.allocate(chunkCount * 8L);
```
lukasz-antoniak: Should we implement here class that extends `Memory`, but uses `chunkOffsets` underneath?

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/BtiIndexReader.java
lukasz-antoniak: I have added the logic to lookup first / last token.

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/BtiIndexReader.java
yifan-c: Yes. `chunkBufferSize`. Returns 4096 if reading the partition index component.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/FileSystemSource.java
```diff
@@ -48,12 +48,14 @@ public class FileSystemSource<T extends CassandraFile> implements CassandraFileS
     private final RandomAccessFile file;
     private final FileType fileType;
     private final long length;
+    private final boolean autoClose;
 
-    public FileSystemSource(T cassandraFile, FileType fileType, Path path) throws IOException
+    public FileSystemSource(T cassandraFile, FileType fileType, Path path, boolean autoClose) throws IOException
```
yifan-c, line 58: please add a comment for the new parameter `autoClose`. There are no javadoc for the other parameters, feel free to ignore those. But would appreciate if you add comment for them too.

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/io/sstable/format/bti/BtiReaderUtils.java
```diff
@@ -0,0 +1,299 @@
...
+                                                                         compression) : null;
+             FileHandle partitionFileHandle = createFileHandle(file,
+                                                               ssTable.openPrimaryIndexStream(),
+                                                               ssTable.length(FileType.PARTITIONS_INDEX),
+                                                               null);
+             FileHandle rowFileHandle = loadRowsIndex ? createFileHandle(file,
+                                                                         ssTable.openRowIndexStream(),
+                                                                         ssTable.length(FileType.ROWS_INDEX),
+                                                                         null) : null;
+             PartitionIndex partitionIndex = PartitionIndex.load(partitionFileHandle, partitioner, false))
```
yifan-c, line 260: `partitionFileHandle` could be `null`, since `null` is a permitted return value from `createFileHandle`.

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/io/sstable/format/bti/BtiReaderUtils.java
```diff
@@ -0,0 +1,299 @@
...
+    {
+        withPartitionIndex(ssTable, descriptor, metadata.partitioner, metadata.params.crcCheckChance, loadDataFile, loadRowsIndex, consumer);
+    }
+
+    private static void withPartitionIndex(@NotNull SSTable ssTable,
+                                           @NotNull Descriptor descriptor,
+                                           @NotNull IPartitioner partitioner,
+                                           double crcCheckChance,
+                                           boolean loadDataFile,
+                                           boolean loadRowsIndex,
```
yifan-c, line 242: Seems quite complex. It would be more clear to me with dedicated method and expressive method names, instead of using boolean to control the behavior. 
But it is just a nit, feel free to ignore.

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/io/sstable/format/bti/BtiReaderUtils.java
```diff
@@ -0,0 +1,299 @@
...
+            consumer.accept(dataFileHandle, partitionFileHandle, rowFileHandle, partitionIndex);
+        }
+    }
+
+    private static FileHandle createFileHandle(File file, InputStream stream, long size, CompressionMetadata compression)
+    {
+        if (stream == null)
+        {
+            return null;
+        }
```
yifan-c: Thinking about it a bit more. What is the benefit of returning `null`? There is no handling for `null` value in the call-sites. 
Maybe throw `IOException` instead?

yifan-c (commented):
forgot to submit the comments :p

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/CompressionMetadata.java
lukasz-antoniak: I have moved around handing of CRC check chance parameter in v5 bridge.

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/CompressionMetadata.java
lukasz-antoniak: Implemented a read only wrapper for `Memory` class.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/FileSystemSource.java
lukasz-antoniak: Yes, done.

lukasz-antoniak:
Closing this PR. Work will be continued in https://github.com/apache/cassandra-analytics/pull/139.

PR #112, opened 2025-05-23
CASSANALYTICS-55 Replace NotImplementedException with UnsupportedOperationException in SparkType

smiklosovic:


smiklosovic:
@jberragan please!

frankgh (approved):
+1 JDK API is always preferred

PR #115, opened 2025-06-03
CASSANALYTICS-61: Allow getting cassandra role in Spark options for use in Sidecar requests for RBAC

sarankk:


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
```diff
@@ -50,6 +50,10 @@ public enum WriterOptions implements WriterOption
     KEYSTORE_PATH,
     KEYSTORE_BASE64_ENCODED,
     KEYSTORE_TYPE,
+    /**
+     * Option that specifies cassandra role used for role based access control in Sidecar.
```
yifan-c, line 54: 👍 on documenting the option. Can you add a bit more details? e.g. the role has to match an exact role that is stored in Cassandra corresponding to the identity, etc.

yifan-c (commented):
lgtm. One minor comment on documentation.

frankgh (approved):
+1 looks good to me

PR #118, opened 2025-06-13
CASSANALYTICS-36: Bulk Reader should dynamically size the Spark job b…

frankgh:
…ased on estimated table size

Patch by Francisco Guerrero; reviewed by TBD for CASSANALYTICS-36


cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkReaderTest.java
```diff
@@ -47,6 +48,25 @@ class BulkReaderTest extends SharedClusterSparkIntegrationTestBase
     QualifiedName table2 = uniqueTestTableFullName(TEST_KEYSPACE);
     QualifiedName tableForNullStaticColumn = uniqueTestTableFullName(TEST_KEYSPACE);
 
+    @Test
+    void testDynamicSizingOption()
+    {
+        Dataset<Row> data = bulkReaderDataFrame(tableForNullStaticColumn).option("SIZING", "dynamic").load();
```
JeetKunDoug: using `tableForNullStaticColumn` here seems implies that this test somehow has something to do with the null static column table, which isn't really the case other than the `row.getString(0)` below - can we switch this to just use one of the other tables?

frankgh: Yeah, good point. Let me make the change

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/DynamicSizing.java
```diff
@@ -0,0 +1,121 @@
...
+                         int availableCores)
+    {
+        this.tableSizeProvider = tableSizeProvider;
+        this.consistencyLevel = consistencyLevel;
+        this.replicationFactor = replicationFactor;
+        this.keyspace = keyspace;
+        this.table = table;
+        this.dc = datacenter;
+        this.maxPartitionSize = maxPartitionSize;
+        this.numCores = availableCores;
```
JeetKunDoug: NIT on my own patch: We should rename the field `availableCores` as well.

analytics-sidecar-client-common/src/main/java/org/apache/cassandra/sidecar/common/data/RestoreJobStatus.java
```diff
@@ -45,6 +45,7 @@ public enum RestoreJobStatus
     /**
      * @deprecated replaced by {@link #ABORTED}
      */
+    @Deprecated
```
yifan-c: unrelated change. Please revert

analytics-sidecar-vertx-client-shaded/build.gradle
```diff
@@ -47,7 +47,7 @@ configurations {
 
 dependencies {
     shadow(group: 'org.slf4j', name: 'slf4j-api', version: "${project.slf4jApiVersion}")
-    api(project(':analytics-sidecar-vertx-client'))
+    implementation(project(':analytics-sidecar-vertx-client'))
```
yifan-c: Unrelated; please revert. If it is a problem, we should fix it in another patch.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/DynamicSizing.java
```diff
@@ -0,0 +1,121 @@
...
+        this.availableCores = availableCores;
+    }
+
+    /**
+     * Returns the effective number of cores to be used during the spark execution.
+     * The value is calculated by getting the table size * the number of replicas
+     * we will use to read the data and then dividing it by the maximum partition
+     * size in GB. For example, assume we have a table with 7.25 GB of data, and
+     * assume a maximum partition size of 2.5 GB. Also, assume that a consistency
+     * level of 2. The number of cores is calculated by the following formula:
```
yifan-c: I do not think it is "consistency level of 2", but should be 3. minReplicas is the quorum of 3.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/DynamicSizing.java
```diff
@@ -0,0 +1,121 @@
...
+     *
+     * <pre>
+     *                                7.25 GB * 2
+     *     effectiveNumberOfCores = --------------- = 5.8 ~&gt; 6 cores
+     *                                  2.5 GB
+     * </pre>
+     *
+     * <p>This method is guaranteed to return at least 1 core and at most {@code availableCores}
+     *
+     * @return the effective number of cores to be used during the spark execution
```
yifan-c, line 105: nit: making the guarantee on return value as part of the `@return` document?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/SidecarTableSizeProvider.java
```diff
@@ -0,0 +1,151 @@
...
+        if (successCount == 0)
+        {
+            throw new RuntimeException(String.format("Unable to determine the table size for table '%s.%s'. 0/%d instances available.",
+                                                     keyspace, table, instancesSize));
+        }
+        else if (successCount < instancesSize)
+        {
+            // normalize table size to the number of instances
+            LOGGER.debug("{}/{} instances were used to determine the table size {} for table {}.{}",
+                         successCount, instancesSize, tableSizeSum, keyspace, table);
```
yifan-c: How about logging at `info` level? It is logged only once per job and the information is seemingly useful for debugging.

JeetKunDoug (approved):

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/DynamicSizing.java
frankgh: not sure I follow this comment

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/DynamicSizing.java
yifan-c: Correction: I meant to say, "the consistency level is quorum/local_quorum in this example, and RF is 3". It leads to the minReplicas of 2.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/DynamicSizing.java
frankgh: Clarified offline, but decided not to take any action as it is a NIT

PR #120, opened 2025-06-17
CASSANALYTICS-32 expose number of rows which violated constraints to JobStats

smiklosovic:


smiklosovic:
This patch lacks the property in settings which would say whether we want to just skip rows which violated constraints or we want to fail the job.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -137,8 +141,24 @@ public void addRow(BigInteger token, Map<String, Object> boundValues) throws IOE
...
-        cqlSSTableWriter.addRow(boundValues);
-        rowCount += 1;
+        try
+        {
+            cqlSSTableWriter.addRow(boundValues);
+            rowCount += 1;
+        }
+        catch (Throwable t)
+        {
+            if (t.getCause() != null && t.getCause().getClass().getName().equals("org.apache.cassandra.cql3.constraints.ConstraintViolationException"))
```
smiklosovic: This is questionable but ... it is here to get the idea. What I am not sure about is that if we moved it "closer" to the source, cqlSSTableWriter is basically a bridge to real implementation. So if we moved it to a bridge, we would need to copy it all over again for every bridge there is / will be.

For now we have only Cassandra 4 bridge and I can not put it there anyway because the concept of "constraints" was not known back in Cassandra 4, it will be there for the first time in Cassandra 5.1 / 6.0 we do not have any bridge for yet.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/JobInfo.java
```diff
@@ -70,6 +70,8 @@ default String getId()
 
     boolean skipExtendedVerify();
 
+    boolean skipRowsViolatingConstraints();
```
frankgh, line 78: We should add javadocs in all public interfaces. This project lacks a log of javadocs in many places, but I would like new contributions to start adding better documentation

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
frankgh: I was going to suggest using `ThrowableUtils` , but it looks like the Analytics version of this class is less featured than the Sidecar version of `ThrowableUtils` .

frankgh: Tiny NIT:
```suggestion
            if (t.getCause() != null && "org.apache.cassandra.cql3.constraints.ConstraintViolationException".equals(t.getCause().getClass().getName()))
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
```diff
@@ -45,6 +45,7 @@ public enum WriterOptions implements WriterOption
     COMMIT_THREADS_PER_INSTANCE,
     COMMIT_BATCH_SIZE,
     SKIP_EXTENDED_VERIFY,
+    SKIP_ROWS_VIOLATING_CONSTRAINTS,
```
frankgh, line 48: Can you maybe add a comment here that this feature is only available in Cassandra 5.0 version? Or is it 6?

frankgh (commented):
Looks good. I left a couple of comments. I was going to suggest adding an integration test, but Analytics currently only supports 4.0, so we won't be able to test it yet

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
smiklosovic: @frankgh thank you for the review but I am not sure this approach is still viable based on the discussion in the related ticket. Did I interpret that comment correctly when I think that what I did here is itself not enough to consider CASSANALYTICS-32 to be resolved as we need to expose failed rows _somehow_, but this I am trying to do here is nice to have / the first step towards that direction which is valuable already to have as such?

https://issues.apache.org/jira/browse/CASSANALYTICS-32?focusedCommentId=17980384&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17980384

frankgh: Yeah, I saw the comment from Yifan after I reviewed the PR. I think it's a good first approach, you need to explicitly enable this feature in order to write data that does not violate the constraints.

PR #121, opened 2025-06-18
[DRAFT] CASSANALYTICS-62 Support for using IP Address as Sidecar/Instance identifier

anderoo:
Currently a draft PR as I need to add more testing and descriptions.

Introduces a new flag for both the reader and writer configuration which,
when set to true, will use the IP address as the identifier and contact
address for Sidecars / Instances. Defaults to false which will continue to
use FQDN instead.

TODO:
- [ ] Tests
- [ ] Ensure IP address is used as identifier when determining ring changes (RecordWriter)


anderoo:
Closing this in favour of [#CASSIDECAR](https://github.com/apache/cassandra-sidecar/pull/227). Happy to reopen if there's a use-case outside of setting this on the sidecar.

PR #123, opened 2025-06-23
CASSANALYTICS-69 Set KeyStore to be optional

anderoo:
Validation currently fails if KeyStore is not present, it would be useful to only fail when KeyStore is invalid or expired.

### DIFF
--- a/cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/validation/KeyStoreValidationTests.java
+++ b/cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/validation/KeyStoreValidationTests.java
@@ -38,7 +38,7 @@ public class KeyStoreValidationTests
         KeyStoreValidation validation = new KeyStoreValidation(secrets);
 
         Throwable throwable = validation.perform();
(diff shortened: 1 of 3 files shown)


sarankk (commented):
Looks good to me.

sarankk (approved):

yifan-c (commented):
Please update changes.txt

I think it makes sense to relax the condition. The prior thought was to enforce mTLS.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/validation/KeyStoreValidationTests.java
```diff
@@ -38,7 +38,7 @@ public void testUnconfiguredKeyStore()
         KeyStoreValidation validation = new KeyStoreValidation(secrets);
 
         Throwable throwable = validation.perform();
-        assertNull(throwable);  // KeyStore is optional
+        assertThat(throwable).isNull(); // KeyStore is optional
```
yifan-c, line 41: 👍

frankgh (approved):
+1

PR #124, opened 2025-06-23
CASSANALYTICS-70: Add spark and scala suffix to artifact name

bbotella:
CASSANALYTICS-70

### DIFF
--- a/analytics-sidecar-vertx-client-shaded/build.gradle
+++ b/analytics-sidecar-vertx-client-shaded/build.gradle
@@ -98,12 +98,18 @@ ext {
     mavenPassword = propertyWithDefault("maven.password", "")
 }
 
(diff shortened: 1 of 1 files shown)


analytics-sidecar-vertx-client-shaded/build.gradle
```diff
@@ -98,12 +98,18 @@ ext {
     mavenPassword = propertyWithDefault("maven.password", "")
 }
 
+archivesBaseName = "analytics-sidecar-vertx-client-all"
```
frankgh, line 105: One minor NIT, let's add a comment here detailing why we don't directly use the available `archivesBaseName` variable

frankgh (approved):
+1

PR #126, opened 2025-06-25
CASSANALYTICS-73: Simplify SidecarTableSizeProvider

frankgh:
Patch by Francisco Guerrero; reviewed by TBD for CASSANALYTICS-73

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/common/SizingFactory.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/common/SizingFactory.java
@@ -22,7 +22,6 @@ package org.apache.cassandra.spark.common;
 import java.util.concurrent.CompletableFuture;
 
 import o.a.c.sidecar.client.shaded.common.response.RingResponse;
-import org.apache.cassandra.clients.Sidecar;
 import org.apache.cassandra.sidecar.client.SidecarClient;
 import org.apache.cassandra.spark.data.ClientConfig;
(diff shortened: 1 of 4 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/common/SizingFactory.java
```diff
@@ -91,9 +90,9 @@ else if (options.sizing() == null || options.sizing().isEmpty() || SIZING_DEFAUL
     }
 
     protected static TableSizeProvider getTableSizeProvider(SidecarClient sidecarClient,
```
liucao-dd, line 92: I'm working on a S3CassandraDataLayer which extends PartitionedDataLayer, but does not rely on sidecar and use the object storage uploaded SSTables.

This means in many code path i don't have a sidecar client. I would argue that taking in a TableSizeProvider as input directly to `Sizing.create` makes more sense in terms of abstraction? I might be biased but that will make providing a  tableSizeProvider easier

frankgh: This makes sense. I think there are some opportunities to decouple Sidecar from the implementation and we should probably aim for that.

liucao-dd: Separately - thoughts on making `maxPartitionSize` accept float (instead of only integer) such that we can allow 0.5GB or 256MB partition size? in my testing 1GB was too big for spark to handle per partition

PR #127, opened 2025-06-27
CASSANALYTICS-74: Add signature to analytics-sidecar-vertx-client-all artifact

bbotella:
Patch for CASSANALYTICS-74


frankgh (approved):
+1

PR #129, opened 2025-07-02
CASSANALYTICS-75: Allow writing to local datacenter only for coordinated write

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSANALYTICS-75


frankgh (approved):
+1 Looks good to me

PR #130, opened 2025-07-03
CASSANALYTICS-76: Add missing artifacts to build

bbotella:
Patch by Bernardo Botella; reviewed by Doug Rohrer for CASSANALYTICS-76


JeetKunDoug (commented):
:shipit: Thanks for the patch!

PR #137, opened 2025-08-27
Filter indexes from snapshots

bbotella:
CASSANALYTICS-86


cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/data/SecondaryIndexFilterTest.java
```diff
@@ -0,0 +1,218 @@
...
+class SecondaryIndexFilterTest extends SharedClusterSparkIntegrationTestBase
+{
+    static final QualifiedName TABLE_WITH_INDEXES = new QualifiedName(TEST_KEYSPACE, "test_secondary_index_filter");
+    static final List<String> DATASET = Arrays.asList("alice", "bob", "charlie", "diana", "eve", "frank", "grace", "henry");
+
+    @Test
+    void testSecondaryIndexFilesAreFilteredFromBulkReader()
+    {
+        // Insert data and flush to ensure SSTables are created (both regular and index files)
+        populateTableAndFlush(TABLE_WITH_INDEXES);
```
frankgh: I think we should populate during schema initialization.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/data/SecondaryIndexFilterTest.java
```diff
@@ -0,0 +1,218 @@
...
+
+        // Verify that secondary index files exist in the filesystem
+        List<Path> indexFiles = findSecondaryIndexFiles();
```
frankgh: instead of collecting all index files, it should be sufficient to return `true` if `anyMatch`

frankgh (commented):
good find, I have a few minor comments

frankgh (approved):
+1 looks good to me

PR #138, opened 2025-09-05
CASSANALYTICS-87: Fix LEAK DETECTED errors during bulk read

yifan-c:
Patch by Yifan Cai, James Berragan; Reviewed by TBD for CASSANALYTICS-87


cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/SSTableCache.java
```diff
@@ -107,6 +108,28 @@ private <T> Cache<SSTable, T> buildCache(int size, int expireAfterMins)
...
+                                   {
+                                       LOGGER.debug("Evicting auto-closable of type: {}", typeLiteral);
+                                       ((AutoCloseable) val).close();
+                                   }
+                                   catch (Exception e)
+                                   {
+                                       LOGGER.error("Exception closing cached instance of {}", typeLiteral, e);
+                                   }
+                               }
+                           })
```
yifan-c, line 139: If commenting it out, the new test fails with "LEAK DETECTED" errors

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/SummaryDbUtils.java
```diff
@@ -70,6 +70,12 @@ public DecoratedKey last()
         {
             return lastKey;
         }
+
+        @Override // The method is expected to be called when evicting the object from sstable cache; do not call it explicitly.
+        public void close() throws Exception
+        {
+            indexSummary.close();
```
jberragan: Be careful with this, I remember trying this once before but it resulted in seg. faults (`SIGSEGV`).

jberragan: I think it is ref. counted in the `org.apache.cassandra.io.util.Memory` class

yifan-c: Ack. I will double check. Thank you!

cassandra-five-zero-bridge/src/test/java/org/apache/cassandra/spark/reader/SSTableCacheTests.java
```diff
@@ -191,4 +201,35 @@ public void testCache()
...
+
+            SummaryDbUtils.Summary summary = SSTableCache.INSTANCE.keysFromSummary(metadata, sstable);
+            assertThat(summary).isNotNull();
+            assertThat(SSTableCache.INSTANCE.containsSummary(sstable)).isTrue();
+            SSTableCache.INSTANCE.invalidate(sstable);
+            summary = null;
+            assertThat(SSTableCache.INSTANCE.containsSummary(sstable)).isFalse();
+            // trigger GC and wait a bit before asserting LEAK DETECTED is not logged.
+            System.gc();
+            Thread.sleep(1000);
```
lukasz-antoniak: NIT: We could use [Awaitility](https://mvnrepository.com/artifact/org.awaitility/awaitility), if we consider more places that need waiting for asynchronous condition. Kafka developers do not pull the dependency (even though only for test), but implemented a simple function to do the same [here](https://github.com/apache/kafka/blob/trunk/core/src/test/scala/unit/kafka/utils/TestUtils.scala#L712).

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/SummaryDbUtils.java
yifan-c: It should be fixed in https://github.com/apache/cassandra-analytics/pull/138/commits/61add5b66b8d29e387ab60248f89f4d747aaa0f0

cassandra-five-zero-bridge/src/test/java/org/apache/cassandra/spark/reader/SSTableCacheTests.java
yifan-c: Changed to `Uninterruptibles`. I was under the impression that we removed `guava` from the repo. But looks like we still have them as test dependency.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/IndexSummaryComponent.java
```diff
@@ -0,0 +1,102 @@
...
+
+import org.apache.cassandra.db.DecoratedKey;
+import org.apache.cassandra.dht.IPartitioner;
+import org.apache.cassandra.io.sstable.IndexSummary;
+import org.apache.cassandra.utils.ByteBufferUtil;
+import org.jetbrains.annotations.Nullable;
+
+public class IndexSummaryComponent implements AutoCloseable
+{
+    private final IndexSummary indexSummary;
```
yifan-c, line 34: Extracted the `IndexSummaryComponent` out from SummaryDbUtils, since `IndexSummary` is version specific.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/IndexSummaryComponent.java
```diff
@@ -0,0 +1,102 @@
...
+     *
+     * @param summaryStream    input stream for Summary.db file
+     * @param partitioner      token partitioner
+     * @param minIndexInterval min index interval
+     * @param maxIndexInterval max index interval
+     * @return Summary object
+     * @throws IOException io exception
+     */
+    @Nullable
+    static IndexSummaryComponent readSummary(InputStream summaryStream,
```
yifan-c, line 49: The read method implementation is version specific too.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/IndexSummaryComponent.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * Get a shared copy of the IndexSummary, whose reference count is incremented.
+     * It is important to close the shared copy to decrement the reference count.
+     * @return a shared copy of the IndexSummary object
+     */
+    public IndexSummary summarySharedCopy()
+    {
+        return indexSummary.sharedCopy();
```
yifan-c, line 84: get a shared copy, so that if `IndexSummaryComponent` is closed, the underlying `indexSummary` is still not freed (since reference count > 0)

yifan-c:
slf4j-test is too slow. So I have to remove the test in https://github.com/apache/cassandra-analytics/pull/138/commits/4190d40539e944c551b936ebbc4be8eb7a66c6f7

yifan-c:
hmm.. just realized that I never merge it!

PR #140, opened 2025-09-17
CASSANALYTICS-90: Add table property to CqlTable

bbotella:
CASSANALYTICS-90

### DIFF
--- a/cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlTable.java
+++ b/cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlTable.java
@@ -46,18 +46,12 @@ public class CqlTable implements Serializable
 
     public enum TableProperty
     {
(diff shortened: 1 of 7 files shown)


cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlTable.java
```diff
@@ -43,6 +44,34 @@ public class CqlTable implements Serializable
 {
     private static final long serialVersionUID = 1018995207366817661L;
 
+    public enum TableProperty
+    {
+        CDC("cdc"),
```
frankgh: I think we should match this to `org.apache.cassandra.spark.utils.CqlUtils#TABLE_PROPERTY_OVERRIDE_ALLOWLIST` because we are bounded to the table property override list. Also, we should really test this to make sure this works when we are parsing the CQL statement and converting that to a CqlTable, like we do in `CassandraDataLayer`


```
        String createStmt = CqlUtils.extractTableSchema(fullSchema, keyspace, table);
        ....
        cqlTable = bridge().buildSchema(createStmt, keyspace, replicationFactor, partitioner, udts, null, indexCount, false);
```

PR #142, opened 2025-09-23
Add CDC Watermarkers interface and implementations

bbotella:
CASSANALYTICS-92


jberragan:
We don't need these anymore, they were replaced with the [CdcState](https://github.com/apache/cassandra-analytics/blob/trunk/cassandra-analytics-common/src/main/java/org/apache/cassandra/cdc/state/CdcState.java) object.

PR #143, opened 2025-09-26
CASSANALYTICS-79: Use tokenRangeReplicas endpoint for determining read replicas

anderoo:
Initial implementation for using tokenRangeReplicas endpoint rather than
calculating token ranges and replicas within the library. There's a fair amount
of code which can be removed afterwards, but wanted to keep the PR small.


anderoo:
Closing in favour of #149

PR #146, opened 2025-10-21
CASSANALYTICS-89: Create dedicated data class for broadcast variable during bulk write

yifan-c:
Extract BulkWriterConfig as an immutable, serializable data class for Spark broadcast variables, replacing direct broadcasting of BulkWriterContext. This addresses the issue where BulkWriterContext, containing mutable state and non-serializable dependencies, was being broadcast to executors.
Follows Spark best practices for broadcast variables (immutable data only) and provides a clear separation between configuration data and stateful context objects.

Patch by Yifan Cai; reviewed by TBD for CASSANALYTICS-89


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -57,7 +57,6 @@
 public class BulkSparkConf implements Serializable
 {
     private static final long serialVersionUID = -5060973521517656241L;
-    private static final Logger LOGGER = LoggerFactory.getLogger(BulkSparkConf.class);
```
yifan-c, line 60: removed the logger to prevent reference walking from root logger by spark size estimator.

CHANGES.txt
```diff
@@ -1,5 +1,6 @@
 0.2.0
 -----
+ * Refactor BulkWriterContext broadcasting to use immutable config class (CASSANALYTICS-TBD)
```
frankgh:
```suggestion
 * Refactor BulkWriterContext broadcasting to use immutable config class (CASSANALYTICS-89)
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/util/SbwKryoRegistrator.java
```diff
@@ -65,7 +67,7 @@ public void registerClasses(@NotNull Kryo kryo)
         kryo.register(StorageCredentials.class, new StorageCredentials.Serializer());
     }
 
-    public static void addJavaSerializableClass(@NotNull Class<? extends Serializable> javaSerializableClass)
+    public static void addJavaSerializableClass(@NotNull Class<?> javaSerializableClass)
```
frankgh, line 78: Can we change the method name here to reflect the actual behavior. Maybe also add a javadoc noting that the classes added here will be registered to Kryo.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -158,7 +157,7 @@ public class BulkSparkConf implements Serializable
     protected final String coordinatedWriteConfJson;
     private transient CoordinatedWriteConf coordinatedWriteConf; // it is transient; deserialized from coordinatedWriteConfJson in executors
 
-    public BulkSparkConf(SparkConf conf, Map<String, String> options)
+    public BulkSparkConf(SparkConf conf, Map<String, String> options, @Nullable Logger logger)
```
frankgh, line 166: NIT, I think preserving the existing constructor will avoid making a lot of changes in many calling sites.

```suggestion
    public BulkSparkConf(SparkConf conf, Map<String, String> options)
    {
        this(conf, options, null);
    }

    public BulkSparkConf(SparkConf conf, Map<String, String> options, @Nullable Logger logger)
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
frankgh: We should document this in code, so there is historical context and we avoid people re-adding the logger in the future.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CoordinatedImportCoordinatorTest.java
```diff
@@ -292,10 +294,10 @@ private RestoreJobProgressResponsePayload failedJobProgress()
                                                 .build();
     }
 
-    // loop at most 10 times until the condition is evaluated to true
+    // loop at most 50 times until the condition is evaluated to true
```
frankgh, line 297: is 50 desirable here?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriterConfig.java
```diff
@@ -0,0 +1,116 @@
...
+ * It contains serializable implementations of cluster information ({@link SerializableClusterInfo} or
+ * {@link SerializableClusterInfoGroup}) that have zero transient fields for safe serialization.
+ * <p>
+ * On the driver, {@link BulkWriterContext} instances use driver-only implementations like
+ * {@link CassandraClusterInfo}. Before broadcasting, these are converted to serializable forms.
+ * On executors, {@link BulkWriterContext} instances are reconstructed from this config using
+ * {@link BulkWriterContext#from(BulkWriterConfig, boolean)}, which uses the serializable cluster
+ * information directly without converting back to driver-only types.
+ */
+public final class BulkWriterConfig implements Serializable
```
frankgh, line 48: My concern here is that some of these fields need to be recomputed dynamically from the executor and we are not honoring the when we are trying to fetch non-cached values in some of the methods of this class.

frankgh (changes requested):
There are some issues with he serialized configurations when the values are dynamic depending on when they get executed.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/util/SbwKryoRegistrator.java
yifan-c: Added documentation

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkWriterConfig.java
yifan-c: addressed

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -84,6 +84,18 @@ public class RecordWriter
...
+    /**
+     * Constructor that accepts a BulkWriterConfig and constructs the context on the executor.
+     * This is used when the config is broadcast to executors.
+     *
+     * @param config      immutable configuration broadcast from driver
+     * @param columnNames column names array
+     */
+    public RecordWriter(BulkWriterConfig config, String[] columnNames)
+    {
+        this(BulkWriterContext.from(config), columnNames);
```
frankgh: We only create a record writer from the executor, the other constructor is not necessary.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/Tokenizer.java
```diff
@@ -39,13 +38,12 @@ interface SerializableFunction<T, R> extends Function<T, R>, Serializable
...
+    public Tokenizer(BroadcastableTableSchema broadcastableTableSchema, boolean isMurmur3Partitioner)
     {
-        TableSchema tableSchema = writerContext.schema().getTableSchema();
-        this.tokenUtils = new TokenUtils(tableSchema.partitionKeyColumns,
-                                         tableSchema.partitionKeyColumnTypes,
-                                         writerContext.cluster().getPartitioner() == Partitioner.Murmur3Partitioner);
-        this.keyColumnProvider = tableSchema::getKeyColumns;
+        this.tokenUtils = new TokenUtils(broadcastableTableSchema.getPartitionKeyColumns(),
+                                         broadcastableTableSchema.getPartitionKeyColumnTypes(),
+                                         isMurmur3Partitioner);
```
frankgh, line 45: it seems a little silly that we pass a boolean flag here for determining whether the partitioner is murmur3 or not. Can't we just directly pass the partitioner to TokenUtils instead?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
```diff
@@ -59,17 +61,29 @@
...
     // immutable
     private final List<ClusterInfo> clusterInfos;
-    private transient volatile Map<String, ClusterInfo> clusterInfoById;
-    private transient volatile TokenRangeMapping<RingInstance> consolidatedTokenRangeMapping;
+    private final String clusterId;
+    private volatile Map<String, ClusterInfo> clusterInfoById;
+    private volatile TokenRangeMapping<RingInstance> consolidatedTokenRangeMapping;
+    // Pre-computed values from BroadcastableClusterInfoGroup (only set when reconstructed on executors)
+    private volatile Partitioner cachedPartitioner;
+    private volatile String cachedLowestCassandraVersion;
```
frankgh: I don't think `cachedPartitioner` or `cachedLowestCassandraVersion` need to be volatile

frankgh (commented):
Looking good, I have some final minor comments

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
yifan-c: true. Deleted.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/Tokenizer.java
yifan-c: Passing the boolean because that is wanted by `TokenUtils` constructor. I would keep it as-is. The change would be unrelated to this refactor. 

```java
    public TokenUtils(List<String> partitionKeyColumns,
                      List<ColumnType<?>> partitionKeyColumnTypes,
                      boolean isMurmur3Partitioner)
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
yifan-c: they do not. Removed the modifier.

frankgh (approved):
+1

PR #147, opened 2025-10-21
CASSANALYTICS-94: Expose SidecarCdc

bbotella:
CASSANALYTICS-94: Expose SidecarCdc client so it is consumable from Sidecar


cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdc.java
```diff
@@ -43,6 +50,31 @@ protected SidecarCdc(@NotNull SidecarCdcBuilder builder)
         initSchema();
     }
 
+    public static SidecarCdcBuilder builder(@NotNull String jobId,
```
jyothsnakonisa, line 53: I do not see this being used anywhere, I am assuming that we need this downstream in sidecar. Can you please remove it if this is not used.

jyothsnakonisa (approved):
Can you please move all the classes in to a new package called `client`, it will be easier to remove all classes in the future when sidecar and analytics repos merge.

Also, since you are adding these classes to avoid dependency conflicts in the dtests, can you please run the dtests and validate that the dependency conflicts doesn't exist,

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdc.java
bbotella: Yes, this will be used in Sidecar.

bbotella:
`Also, since you are adding these classes to avoid dependency conflicts in the dtests` -> this is not really the case. The interfaces are being added to avoid exposing shaded classes to the CDC builders from sidecar. Even if we end up merging the repositories, I think there is still value in having specific interfaces to be implemented to keep a nice separation between project modules. Not a hill I will die on tho, and if the preference is strong towards adding a new package on `org.apache.cassandra.cdc.sidecar.client`, I can definitely move them.

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcClient.java
```diff
@@ -150,4 +167,242 @@ public String hostname()
...
+                                       DEFAULT_MAX_BUFFER_SIZE,
+                                       4 * 1024 * 1024L,
+                                       DEFAULT_MAX_POOL_SIZE,
+                                       DEFAULT_TIMEOUT_SECONDS,
+                                       DEFAULT_CASSANDRA_ROLE,
+                                       DEFAULT_MAX_BUFFER_OVERRIDE,
+                                       chunkOverride);
+        }
+
+        public static ClientConfig create(Map<String, String> options)
```
frankgh: this method doesn't really make sense here. I don't think options for this client are passed via a Map? Is this used anywhere?

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdc.java
frankgh: should we add a unit test then ?

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcClient.java
```diff
@@ -150,4 +167,225 @@ public String hostname()
             }
         };
     }
+
+    public static final class ClientConfig
+    {
+        public static final String SIDECAR_PORT = "sidecar_port";
+        public static final String MAX_RETRIES_KEY = "maxRetries";
+        public static final String DEFAULT_MILLIS_TO_SLEEP_KEY = "defaultMillisToSleep";
+        public static final String MAX_MILLIS_TO_SLEEP_KEY = "maxMillisToSleep";
```
frankgh: no longer needed
```suggestion
```

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcClient.java
```diff
@@ -150,4 +167,225 @@ public String hostname()
...
+        public static final String MAX_BUFFER_SIZE_BYTES_KEY = "maxBufferSizeBytes";
+        public static final String CHUNK_BUFFER_SIZE_BYTES_KEY = "chunkBufferSizeBytes";
+        public static final String MAX_POOL_SIZE_KEY = "maxPoolSize";
+        public static final String TIMEOUT_SECONDS_KEY = "timeoutSeconds";
+        public static final String CASSANDRA_ROLE_KEY = "cassandra_role";
```
frankgh: No longer needed
```suggestion
```

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcClient.java
```diff
@@ -150,4 +167,225 @@ public String hostname()
...
+            return ClientConfig.create(userProvidedPort, DEFAULT_MAX_RETRIES, DEFAULT_MILLIS_TO_SLEEP);
+        }
+
+        public static ClientConfig create(int userProvidedPort, int maxRetries, long millisToSleep)
+        {
+            Map<FileType, Long> chunkOverride = new HashMap<>();
+            chunkOverride.put(FileType.COMMITLOG, 4 * 1024 * 1024L);  // 4MB chunks
+
+            Map<FileType, Long> maxBufferOverride = new HashMap<>();
+            maxBufferOverride.put(FileType.COMMITLOG, 8 * 1024 * 1024L);
```
frankgh: this is unused, is that on purpose, or are we missing something here?

frankgh (approved):
+1

PR #148, opened 2025-10-21
CASSANALYTICS-95: Remove extra buffer flips

bbotella:
CASSANALYTICS-95

### DIFF
--- a/cassandra-four-zero-bridge/src/test/java/org/apache/cassandra/io/util/CdcRandomAccessReaderTest.java
+++ b/cassandra-four-zero-bridge/src/test/java/org/apache/cassandra/io/util/CdcRandomAccessReaderTest.java
@@ -45,7 +45,6 @@ public class CdcRandomAccessReaderTest
 {
     private TestCommitLog testCommitLog;
     private TestCassandraFileSource testSource;
-    private TestCdcStats testStats;
 
     private CdcRandomAccessReader reader;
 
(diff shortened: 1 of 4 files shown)


cassandra-four-zero-bridge/src/test/java/org/apache/cassandra/io/util/CdcRandomAccessReaderTest.java
```diff
@@ -0,0 +1,455 @@
...
+        assertThat(reader).isNotNull();
+        assertThat(reader.getPath()).isEqualTo("/test/path/commitlog");
+
+        // Verify no premature calls before rebuffer is used
+        assertThat(testSource.requestCalls.size()).isEqualTo(0);
+    }
+
+    // Test stub classes
+
+    private static class TestCommitLog implements CommitLog
```
jyothsnakonisa, line 246: Can you check if you can use `LocalCommitLog`

https://github.com/apache/cassandra-analytics/blob/trunk/cassandra-analytics-cdc/src/test/java/org/apache/cassandra/cdc/LocalCommitLog.java

cassandra-four-zero-bridge/src/test/java/org/apache/cassandra/io/util/CdcRandomAccessReaderTest.java
```diff
@@ -0,0 +1,455 @@
...
+            RequestCall(long start, long end, StreamConsumer consumer)
+            {
+                this.start = start;
+                this.end = end;
+                this.consumer = consumer;
+            }
+        }
+    }
+
+    private static class TestCdcStats implements ICdcStats
```
jyothsnakonisa: Can we use `ICdcStats.STUB` instead?

jyothsnakonisa (approved):
Looks good to me, minor comments.

cassandra-four-zero-bridge/src/test/java/org/apache/cassandra/io/util/CdcRandomAccessReaderTest.java
bbotella: I don't think that is possible. cassandra-analytics-cdc depends on `fourzeroBridge(project(path: ':cassandra-four-zero-bridge'))`, which is where this class is being implemented. Re-using it would introduce a circular dependency.

frankgh (approved):
+1

PR #149, opened 2025-10-23
CASSANALYTICS-79: Make Token ranges calculation rack aware for spark.data.partitioner.CassandraRing

sklaha:
#### Changes in the PR:

* An integration test (BulkReaderTokenRangeReplicasTest.java) has been added to demonstrate that the replicas for token range calculation were incorrect in CassandraRing with NetworkAwareReplication strategy in a multi-DC, multi-rack scenario.
* The sidecar token-range-replicas API has been utilized to retrieve replicas for token ranges, eliminating the need for token range calculation within the analytics-library.
* Unit tests have been added to further validate the changes.


cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/partitioner/MurmurHash.java
```diff
@@ -156,4 +156,89 @@ public static long[] hash(ByteBuffer key, int offset, int length, long seed)
 
         return new long[]{h1, h2};
     }
+
+    protected static long invRShiftXor(long value, int shift)
```
yifan-c, line 160: Why are those new methods added?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/partitioner/CassandraInstance.java
```diff
@@ -100,6 +103,20 @@ public String toString()
...
+        this.node = in.readUTF();
+        this.dataCenter = in.readUTF();
+    }
+
+    private void writeObject(ObjectOutputStream out) throws IOException
+    {
+        out.writeUTF(token());
+        out.writeUTF(nodeName());
+        out.writeUTF(dataCenter());
+    }
```
yifan-c, line 118: Why adding the custom serialization code? The string fields should be serializable by default, and the readObject & writeObject methods are unnecessary.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/partitioner/CassandraInstance.java
```diff
@@ -34,9 +37,9 @@ public class CassandraInstance implements TokenOwner, Serializable
     public static final CassandraInstance.Serializer SERIALIZER = new CassandraInstance.Serializer();
 
     private static final long serialVersionUID = 6767636627576239773L;
-    private final String token;
-    private final String node;
-    private final String dataCenter;
+    private String token;
+    private String node;
+    private String dataCenter;
```
yifan-c, line 42: Once readObject & writeObject are removed, the `final` modifiers can be restored.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/partitioner/CassandraRing.java
```diff
@@ -74,6 +74,7 @@ public class CassandraRing implements Serializable
     private ReplicationFactor replicationFactor;
     private List<CassandraInstance> instances;
 
+    private transient RangeMap<BigInteger, List<CassandraInstance>> replicasForRanges;
     private transient RangeMap<BigInteger, List<CassandraInstance>> replicas;
```
yifan-c, line 78: It is confusing to have 2 maps and they actually point to the same object if I am not mistaken. 
More importantly, it leads to different `init` code paths. The patch is to fix the broken token calculation, but why leaving it half-fixed? The CDC code patch still calls the original constructor, which means CDC is still broken.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkReaderTokenRangeReplicasTest.java
```diff
@@ -0,0 +1,248 @@
...
+
+        // update the value internally at node 4
+        updateInternal(4, key, VALUE2);
+        // validate the values across the nodes:
+        // node 4 should have VALUE2
+        // node 2, node 6 shouldn't have the key
+        // all other nodes should have VALUE1
+        expectedValuesInNodes.put(4, VALUE2);
+        //updateInternal(1, key, VALUE2);
+        //expectedValuesInNodes.put(1, VALUE2);
```
yifan-c, line 153: Are those 2 lines wanted?

PR #150, opened 2025-10-29
CASSANALYTICS-98: Fix SSTable descriptor mismatch preventing newly produced SSTables from being uploaded

yifan-c:
Patch by Yifan Cai; reviewed by TBD for CASSANALYTICS-98

### DIFF
--- a/cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/common/SSTablesTest.java
+++ b/cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/common/SSTablesTest.java
@@ -165,6 +165,7 @@ public class SSTablesTest
 
         // These must be equal for the filter in SortedSSTableWriter.prepareSStablesToSend to work
         assertThat(descriptorFromPath).isEqualTo(descriptorFromProduction);
-        assertThat(descriptorFromPath.baseFilename).isEqualTo(descriptorFromProduction.baseFilename);
+        assertThat(descriptorFromPath.baseFilename).isEqualTo(descriptorFromProduction.baseFilename)
+                                                   .doesNotEndWith("-");
     }
 }
--- a/cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/common/SSTablesTest.java
+++ b/cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/common/SSTablesTest.java
@@ -1,0 +1,170 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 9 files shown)


cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/common/SSTablesTest.java
```diff
@@ -0,0 +1,170 @@
...
+        //
+        // When SortedSSTableWriter.prepareSStablesToSend creates a descriptor from a file path,
+        // it should also return "oa-1-big" (no trailing dash) so they match.
+
+        Path dataFile = Paths.get("/path/to/sstable/oa-1-big-Data.db");
+        SSTableDescriptor descriptorFromPath = SSTables.getSSTableDescriptor(dataFile);
+
+        // Simulate what SSTableWriterImplementation.onSSTablesProduced does
+        SSTableDescriptor descriptorFromProduction = new SSTableDescriptor("oa-1-big");
+
```
JeetKunDoug, line 165: Can we assert here something that clearly demonstrates what the _wrong_ answer was previously? Maybe:
```
    assertThat(descriptorFromPath).doesNotEndwith("-");
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
```diff
@@ -85,12 +85,17 @@ protected void onSSTablesProduced(Set<SSTableDescriptor> sstables)
...
                 // 4. remove the sstables once sent
                 Map<Path, Digest> fileDigests = sstableWriter.prepareSStablesToSend(writerContext, sstables);
-                recordStreamedFiles(fileDigests.keySet());
-                fileDigests.keySet()
-                           .stream()
-                           .filter(p -> p.getFileName().toString().endsWith(FileType.DATA.getFileSuffix()))
-                           .forEach(this::sendSStableToReplicas);
-                LOGGER.info("[{}]: Sent SSTables. sstables={}", sessionID, sstableWriter.sstableCount());
+                // retain only the SSTable data components
+                Set<Path> newSSTables = fileDigests.keySet()
```
JeetKunDoug: Why are we collecting these back into a Set? Given `fileDigests` is already a Map<Path, Digest>, we shouldn't need to collect these again, we could just use the "AtomicInteger in the foreach" trick to count them right (given that's the only thing we're using the set for?)

```java                
                AtomicInteger sstableCount = new AtomicInteger();
                fileDigests.keySet()
                           .stream()
                           .filter(p -> p.getFileName().toString().endsWith(FileType.DATA.getFileSuffix()))
                           .forEach(p -> {
                               sstableCount.incrementAndGet();
                               sendSStableToReplicas(p);
                           });
                LOGGER.info("[{}]: Sent newly produced SSTables. sstables={}", sessionID, sstableCount.get());```

JeetKunDoug (commented):
Looks good so far. A few small comments/questions for review.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
yifan-c: yep. it works better. 
I am actually using `org.apache.cassandra.util.IntWrapper` as the counter, as there is no concurrency requirement.

JeetKunDoug (approved):
Looking good - thanks for the quick updates.

PR #154, opened 2025-11-10
CASSANALYTICS-103 update Guice to 7.0.0

smiklosovic:


frankgh (approved):
+1 thanks for the fix

bbotella (approved):
+1

PR #155, opened 2025-11-13
CASSANALYTICS-102: Add TimeRangeFilter to filter out SSTables outside given time window

sarankk:


cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/filters/TimeRangeFilter.java
```diff
@@ -0,0 +1,138 @@
...
+
+import com.google.common.collect.Range;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * {@link TimeRangeFilter} to filter out based on timestamp.
+ * Uses Google Guava's Range internally for storing time range.
+ */
+public final class TimeRangeFilter implements Serializable
```
yifan-c: maybe consider removing the `final` modifier for extension.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/filters/TimeRangeFilter.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Determines if given start and end timestamp match the filter. SSTable is included if it overlaps with
+     * filter time range.
+     *
+     * @param givenStart the SSTable min timestamp
+     * @param givenEnd the SSTable max timestamp
+     * @return true if the SSTable should be included, false if it should be omitted.
+     */
+    public boolean filter(long givenStart, long givenEnd)
```
yifan-c: Can you just rename the method to `overlaps`? Given that it is really to check whether 2 ranges overlap. 
The "given" in the parameter names is redundant, please remove it and actually add "inclusive" and "micros" to the name, if the idea is to improve clarity.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/filters/TimeRangeFilter.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * Creates a {@link TimeRangeFilter} with only start bound.
+     *
+     * @param startTimestampMicros the start timestamp in microseconds (inclusive)
+     * @return {@link TimeRangeFilter} with only start timestamp
+     */
+    @NotNull
+    public static TimeRangeFilter startingAt(long startTimestampMicros)
```
yifan-c: 👍  on add the time unit here. 

Can you update the other time references that the time unit is in micros? I believe this filter implementation only expect time in micros. Please consider document it in the class, and even rename the class name if it can be more explicit.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/filters/TimeRangeFilter.java
```diff
@@ -0,0 +1,138 @@
...
+     * @param givenEnd the SSTable max timestamp
+     * @return true if the SSTable should be included, false if it should be omitted.
+     */
+    public boolean filter(long givenStart, long givenEnd)
+    {
+        // Create range for the given SSTable timestamps, always closed
+        Range<Long> sstableTimeRange = Range.closed(givenStart, givenEnd);
+
+        // Check if ranges are connected (overlap or adjacent)
+        return timeRange.isConnected(sstableTimeRange) && !timeRange.intersection(sstableTimeRange).isEmpty();
```
yifan-c: I think you can come up with a comparison by avoiding creating a new range instance. Could you update the implementation here? `intersection()` method creates a new range each time. 

Given both ranges for comparison are closed, calling `isConnected` should be able to tell the result. No?

cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/sparksql/filters/TimeRangeFilterTest.java
```diff
@@ -0,0 +1,267 @@
...
+
+        TimeRangeFilter sstableOverlapWithStartExclusive = TimeRangeFilter.create(99L, 200L);
+        // SSTable: [50, 150], Filter: (100, 200]
+        assertThat(sstableOverlapWithStartExclusive.filter(50L, 150L)).isTrue();
+
+        // start inclusive
+
+        TimeRangeFilter sstableTouchingEndExclusive = TimeRangeFilter.create(100L, 199L);
+        // SSTable: [200, 300], Filter: [100, 200) - no overlap
+        assertThat(sstableTouchingEndExclusive.filter(200L, 300L)).isFalse();
```
yifan-c: This does not really create the range [100, 200). It creates [100, 199]. The test case here (and the other ones testing exclusiveness) are all redundant and even misleading. Please have a clear definition that `TimeRangeFilter` creates closed range of timestamps in micros only.

cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/sparksql/filters/TimeRangeFilterTest.java
```diff
@@ -0,0 +1,267 @@
...
+    void testToString()
+    {
+        TimeRangeFilter boundedRangeFilter = TimeRangeFilter.create(100L, 200L);
+        assertThat(boundedRangeFilter.toString()).isEqualTo("TimeRangeFilter[100‥200]");
+
+        TimeRangeFilter startAtInclusiveFilter = TimeRangeFilter.startingAt(100L);
+        assertThat(startAtInclusiveFilter.toString()).isEqualTo("TimeRangeFilter[100‥+∞)");
+
+        TimeRangeFilter endAtInclusiveFilter = TimeRangeFilter.endingAt(200L);
+        assertThat(endAtInclusiveFilter.toString()).isEqualTo("TimeRangeFilter(-∞‥200]");
```
yifan-c: negative value looks confusing for time range. I would rather use 0 as the start.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/filters/TimeRangeFilter.java
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Creates a {@link TimeRangeFilter} with only end bound.
+     *
+     * @param endTimestampMicros the end timestamp in microseconds (inclusive)
+     * @return {@link TimeRangeFilter} with only end timestamp
+     */
+    @NotNull
+    public static TimeRangeFilter endingAt(long endTimestampMicros)
+    {
+        Range<Long> range = Range.atMost(endTimestampMicros);
```
yifan-c: Please consider having `Range.closed(0, endTimestampMicros)` instead

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -264,6 +264,7 @@ public void validateSSTables(@NotNull BulkWriterContext writerContext, @NotNull
                                                       Collections.emptyList() /* requestedFeatures */,
                                                       false /* useSSTableInputStream */,
                                                       null /* statsClass */,
+                                                      null,
```
yifan-c: add comment to be consistent with the other parameters.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/ClientConfig.java
```diff
@@ -81,6 +83,14 @@ public class ClientConfig
     public static final String ENABLE_EXPANSION_SHRINK_CHECK_KEY = "enableExpansionShrinkCheck";
     public static final String SIDECAR_PORT = "sidecar_port";
     public static final String QUOTE_IDENTIFIERS = "quote_identifiers";
+    /**
+     * {@code sstable_start_timestamp} and {@code sstable_end_timestamp} are by default inclusive. Both start and end
+     * timestamps are represented in microseconds. To have exclusive bounds, add +1 offset for exclusive start bound
+     * and add -1 for exclusive end bound.
```
yifan-c: The comment is confusing to read. Both the range ends are closed/inclusive _always_.

I would just have this.
```suggestion
     * {@code sstable_start_timestamp_micros} and {@code sstable_end_timestamp_micros} define a time range filter
     * for SSTable selection. Both timestamps are represented in microseconds and both bounds are always inclusive
     * (closed range).
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/DataLayer.java
```diff
@@ -198,6 +199,11 @@ public SparkRangeFilter sparkRangeFilter(int partitionId)
         return null;
     }
 
+    public List<TimeRangeFilter> sstableTimeRangeFilters()
```
yifan-c: Please add javadoc. In addition, define the meaning of the empty list.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/LocalDataLayer.java
```diff
@@ -249,6 +258,7 @@ private LocalDataLayer(@NotNull CassandraVersion version,
                            @NotNull List<SchemaFeature> requestedFeatures,
                            boolean useBufferingInputStream,
                            @Nullable String statsClass,
+                           @Nullable TimeRangeFilter sstableTimeRangeFilter,
```
yifan-c: I noticed that the patch uses `List.of()` for sstableTimeRangeFilter when not defined. But the `Nullable` here implies that null value can also be used. It is inconsistent. Please pick one; otherwise, it will bite us later. 
I would suggest to change the annotation to `@NotNull` and ensure callsites do not pass null value.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/utils/FilterUtils.java
```diff
@@ -111,4 +115,29 @@ public static List<List<String>> cartesianProduct(List<List<String>> orderedValu
...
     }
+
+    /**
+     * Parses {@link TimeRangeFilter} from spark options.
+     */
+    public static TimeRangeFilter parseSSTableTimeRangeFilter(Map<String, String> options)
+    {
+        if (!options.containsKey(SSTABLE_START_TIMESTAMP_MICROS) && !options.containsKey(SSTABLE_END_TIMESTAMP_MICROS))
+        {
+            return null;
```
yifan-c: This is _only_ a nit:

```suggestion
            return TimeRangeFilter.empty(); // create a constant: Range.openClosed(0L, 0L) as empty range in TimeRangeFilter
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/utils/FilterUtils.java
```diff
@@ -111,4 +115,29 @@ public static List<List<String>> cartesianProduct(List<List<String>> orderedValu
...
+        }
+
+        long startTimestamp = MapUtils.getLong(options, SSTABLE_START_TIMESTAMP_MICROS, Long.MIN_VALUE);
+        if (!options.containsKey(SSTABLE_END_TIMESTAMP_MICROS))
+        {
+            return TimeRangeFilter.startingAt(startTimestamp);
+        }
```
yifan-c: I do not think it is necessary. Both cases (the end time option exists or not) are already covered in `TimeRangeFilter.create`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/utils/FilterUtils.java
```diff
@@ -111,4 +115,29 @@ public static List<List<String>> cartesianProduct(List<List<String>> orderedValu
...
+
+        long endTimestamp = MapUtils.getLong(options, SSTABLE_END_TIMESTAMP_MICROS, Long.MAX_VALUE);
+        if (!options.containsKey(SSTABLE_START_TIMESTAMP_MICROS))
+        {
+            return TimeRangeFilter.endingAt(endTimestamp);
+        }
```
yifan-c: Not necessary for the same reason.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/data/LocalDataLayerTests.java
```diff
@@ -88,4 +97,98 @@ public void testEquality(CassandraBridge bridge)
...
+        options.put("keyspace", "test_keyspace");
+        options.put("createstmt", SchemaTests.SCHEMA);
+        options.put("dirs", "/tmp/data1,/tmp/data2");
+        options.put("sstable_start_timestamp_micros", "1000");
+        options.put("sstable_end_timestamp_micros", "2000");
+
+        LocalDataLayer dataLayer = LocalDataLayer.from(options);
+
+        List<TimeRangeFilter> filters = dataLayer.sstableTimeRangeFilters();
+        assertThat(filters).hasSize(1);
```
yifan-c: When there will be a list of `TimeRangeFilter`? And how to configure the list?

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/SSTableReaderTests.java
```diff
@@ -143,4 +146,161 @@ private void testTtlUsingConstantReferenceTimeHelper(CassandraBridge bridgeForTe
...
+            schema.writeSSTable(dir, bridgeInTest, partitioner, writer -> {
+                for (int i = 0; i < 10; i++)
+                {
+                    writer.write(i, i * 100);
+                }
+            });
+
+            // Wait to ensure timestamp advances
+            Uninterruptibles.sleepUninterruptibly(100, TimeUnit.MILLISECONDS);
+            long currentTimestampMicros = TimeUnit.MILLISECONDS.toMicros(System.currentTimeMillis()); // Convert to microseconds
```
yifan-c: Can you look into using the `TimeProvider` to mock the time, instead of pausing? 
Check out `org.apache.cassandra.spark.SSTableReaderTests#testTtlUsingConstantReferenceTimeHelper` for the usage example.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkReaderFilteringIntegrationTest.java
```diff
@@ -0,0 +1,221 @@
...
+            String query = String.format("INSERT INTO %s (id, data, timestamp) VALUES ('%s', 'data_%s', %d) USING TIMESTAMP %d",
+                                         twcsTable, id, "data" + id, MIDDLE_TIMESTAMP_MICROS, timestamp);
+            coordinator.execute(query, ConsistencyLevel.ALL);
+        }
+
+        // Flush to create second SSTable
+        instance.nodetool("flush", TEST_KEYSPACE, twcsTable.table());
+
+        // wait for nodetool flush
+        Uninterruptibles.sleepUninterruptibly(30, TimeUnit.SECONDS);
```
yifan-c: This test suite is slow. Each test spends 1 minute on sleeping. 
Why would nodetool flush take so long?

yifan-c (commented):
I do not see the validation that check the target table is using TWCS when the time range filter is configured. Should you add the check?

cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/sparksql/filters/TimeRangeFilterTest.java
sarankk: Removed all exclusive test cases and redundant tests.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/filters/TimeRangeFilter.java
sarankk: Yes with closed ranges, intersection is no longer needed, removed the call.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/data/LocalDataLayerTests.java
sarankk: Currently we do not have support for reading list in spark options. We can add that later when required.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkReaderFilteringIntegrationTest.java
sarankk: I checked flush call, it is synchronous, will remove wait times.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/SSTableReaderTests.java
sarankk: Wait was unnecessary, removed it.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/sparksql/filters/SSTableTimeRangeFilter.java
```diff
@@ -0,0 +1,127 @@
...
+public class SSTableTimeRangeFilter implements Serializable
+{
+    public static final SSTableTimeRangeFilter EMPTY = new SSTableTimeRangeFilter(Range.closed(0L, 0L))
+    {
+        @Override
+        public boolean overlaps(long startMicros, long endMicros)
+        {
+            return true; // accepts all SSTables
+        }
+    };
```
yifan-c: When I originally suggested, the empty filter was with range `Range.openClosed(0L, 0L)`. The filter does _not_ overlap with any sstables, hence it is named `EMPTY`. It is to replace the usage of `null`. 
The code here creates a filter that overlaps with all SSTables. Now I realize that this is probably the behavior you want. Could you rename the filter? It is not really `EMPTY`, but `ALL`. The code can be simplified to 
```suggestion
    public static final SSTableTimeRangeFilter ALL = new SSTableTimeRangeFilter(Range.all());
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/DataLayer.java
```diff
@@ -198,6 +199,17 @@ public SparkRangeFilter sparkRangeFilter(int partitionId)
         return null;
     }
 
+    /**
+     * Returns a list of {@link SSTableTimeRangeFilter} to filter out SSTables based on min and max timestamp.
+     *
+     * @return list of {@link SSTableTimeRangeFilter} for filtering SSTables based on max and min timestamps. When
+     * the list is empty, all SSTables are included for bulk reading.
+     */
+    public List<SSTableTimeRangeFilter> sstableTimeRangeFilters()
```
yifan-c: Please update the return type. Just return the filter, instead of a list of filters.

yifan-c (commented):
One last comment

PR #156, opened 2025-11-14
CASSANALYTICS-105: Fix gradle generated artifacts

bbotella:
CASSANALYTICS-105


JeetKunDoug (approved):
Looking good now. We should probably try to add the verification instructions into the repo somewhere as a follow-up.

Verified I can build/sign/checksum the deliverables, and the signatures and shasums all match.

PR #157, opened 2025-11-14
CASSANALYTICS-106: Setup CI Pipeline with GitHub Actions

lukasz-antoniak:
Fixes [CASSANALYTICS-106](https://issues.apache.org/jira/browse/CASSANALYTICS-106).


yifan-c:
I would say we just merge the PR. It is pure added CI function. Thoughts? @lukasz-antoniak @frankgh @JeetKunDoug

frankgh:
+1 thanks for this @lukasz-antoniak

lukasz-antoniak:
Here is an example how results look like: https://github.com/lukasz-antoniak/cassandra-analytics/actions/runs/22672262229?pr=12

PR #162, opened 2025-12-12
CASSANALYTICS-107: Fix race condition in DirectStreamSession#onSSTablesProduced and SortedSStableWriter#close

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSANALYTICS-107

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
@@ -63,6 +63,21 @@ import org.jetbrains.annotations.Nullable;
  * entire partition, i.e. repartitionAndSortWithinPartitions. By doing so, it eliminates the nice property of the
  * output sstable being globally sorted and non-overlapping.
  * Unless you can think of a better use case, we should stick with this SortedSSTableWriter
+ * <br>
+ * <p>Threading Model:</p>
+ * This class has limited thread-safety guarantees:
+ * <ul>
+ *   <li>{@link #addRow(BigInteger, Map)} and {@link #close(BulkWriterContext)} MUST be called from the same thread
+ *       (typically the RecordWriter thread). These methods are NOT synchronized and must not be called concurrently.</li>
+ *   <li>{@link #prepareSStablesToSend(BulkWriterContext, Set)} MAY be called concurrently from background threads
+ *       (via {@link StreamSession}'s executor service) and is synchronized to protect shared state.</li>
+ *   <li>{@link #close(BulkWriterContext)} is synchronized to prevent races with concurrent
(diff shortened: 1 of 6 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -63,6 +63,21 @@
  * entire partition, i.e. repartitionAndSortWithinPartitions. By doing so, it eliminates the nice property of the
  * output sstable being globally sorted and non-overlapping.
  * Unless you can think of a better use case, we should stick with this SortedSSTableWriter
+ * <br>
+ * <p>Threading Model:</p>
+ * This class has limited thread-safety guarantees:
+ * <ul>
+ *   <li>{@link #addRow(BigInteger, Map)} and {@link #close(BulkWriterContext)} MUST be called from the same thread
```
lukasz-antoniak, line 70: `addRow` method cannot be easily made thread safe because down the line `CQLSStableWriter` is not. Shall we annotate `SSTableWriterImplementation` as `@NotThreadSafe`?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -209,24 +275,27 @@ public void close(BulkWriterContext writerContext) throws IOException
         }
         isClosed = true;
         cqlSSTableWriter.close();
```
lukasz-antoniak, line 275: Unrelated to this change, but shall we do any cleanup if this operation fails with IO error (or when validation of SSTable fails)?

yifan-c: good catch!

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/StreamSession.java
```diff
@@ -139,7 +141,16 @@ public Future<StreamResult> finalizeStreamAsync() throws IOException
...
+        try
+        {
+            sstableWriter.close(writerContext);
+        }
+        catch (IOException ioe)
+        {
+            // clean up the sstables and rethrow on I/O errors when closing
+            cleanupSSTables(LOGGER);
+            throw ioe;
+        }
```
yifan-c, line 153: @lukasz-antoniak , the clean up is added here. When close is called and failed, all `prepareSStablesToSend` scheduled in the future will be no-op and it should be safe to remove all sstables eagerly.

lukasz-antoniak (approved):
Nice, looks good to me.

PR #169, opened 2026-02-10
CASSANALYTICS-24 and CASSANALYTICS-25 SSTable version based bridge determination

skoppu22:
Jira tasks: CASSANALYTICS-24 and CASSANALYTICS-25

This PR adds sstable version based bridge determination feature to C* analytics. This feature allows C* analytics to choose bridge (which C* version jar to load) based on sstable versions existing on the cluster, instead of strictly depending on running C* versions. This is needed especially to support C* 5.0, which supports multiple compatibility modes.

CI run: https://app.circleci.com/pipelines/github/skoppu22/cassandra-analytics/37/workflows/da025a29-00c0-41f2-95e8-63c2b7f1f276


cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/CassandraVersion.java
```diff
@@ -35,22 +37,49 @@
...
 
-    CassandraVersion(int number, String name, String jarBaseName, String... sstableFormats)
+
+    CassandraVersion(int number, String name, String jarBaseName, String[] sstableFormats, String[] nativeSStableVersions)
     {
         this.number = number;
         this.name = name;
         this.jarBaseName = jarBaseName;
         this.sstableFormats = new HashSet<>(Arrays.asList(sstableFormats));
+        this.nativeSStableVersions = Collections.unmodifiableList(Arrays.asList(nativeSStableVersions));
```
sarankk: Nit
```suggestion
        this.nativeSStableVersions = List.of(nativeSStableVersions);
```

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/CassandraVersion.java
```diff
@@ -35,22 +37,49 @@
...
 public enum CassandraVersion
 {
-    THREEZERO(30, "3.0", "three-zero", "big"),
-    FOURZERO(40, "4.0", "four-zero", "big"),
-    FOURONE(41, "4.1", "four-zero", "big"),
-    FIVEZERO(50, "5.0", "five-zero", "big", "bti");
+    THREEZERO(30, "3.0", "three-zero", new String[]{"big"},
+              new String[]{
+                  // Cassandra 3.x native sstable versions
+                  // order is important, used to determine the latest version
```
sarankk: Array based ordering will be hard to debug in case of issues and hard to maintain as well.

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/SSTableVersionAnalyzer.java
```diff
@@ -0,0 +1,233 @@
...
+import java.util.Comparator;
+import java.util.Optional;
+import java.util.Set;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Analyzes SSTable versions on a cluster to determine the appropriate
+ * Cassandra bridge to load for bulk write operations.
```
sarankk: We are using version analyzer for both read and write
```suggestion
 * Cassandra bridge to load for bulk read/write operations.
```

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/SSTableVersionAnalyzer.java
```diff
@@ -0,0 +1,233 @@
...
+            if (!v1Opt.isPresent())
+            {
+                throw new IllegalArgumentException(
+                    String.format("Unknown SSTable version: %s. Cannot determine Cassandra version.", v1));
+            }
+
+            if (!v2Opt.isPresent())
+            {
+                throw new IllegalArgumentException(
+                    String.format("Unknown SSTable version: %s. Cannot determine Cassandra version.", v2));
```
sarankk: Instead of throwing even if one format is incorrect, we could log these as errors and throw in the end if no max version is found.

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/CassandraVersion.java
sarankk: After reading `SSTableVersionAnalyzer` we could do away without the internal ordering between the sstable versions. Since we are still interested in the associated cassandra version. We could treat the sstable versions within the Cassandra version on same level. Wdyt?

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/CassandraVersion.java
```diff
@@ -68,36 +97,126 @@ public String jarBaseName()
...
+     */
+    @VisibleForTesting
+    int getPreviousMajorVersion()
+    {
+        // Get major version: 50 -> 5, 41 -> 4, 40 -> 4, 30 -> 3
+        int majorVersion = this.number / 10;
+
+        // Calculate previous major version: (majorVersion - 1) * 10
+        // E.g., 5 -> 40, 4 -> 30, 3 -> 20
+        return (majorVersion - 1) * 10;
```
sarankk: Nit: How about we keep an ordered list of `CassandraVersion` enums and return the previous one, instead of computing the version.

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/SSTableVersionAnalyzer.java
```diff
@@ -0,0 +1,233 @@
...
+     * @throws IllegalArgumentException if format is invalid, or cluster doesn't support requested format,
+     *                                  or SSTable versions are empty/unknown
+     */
+    public static CassandraVersion determineBridgeVersionForWrite(Set<String> sstableVersionsOnCluster,
+                                                                  String requestedFormat,
+                                                                  String cassandraVersion,
+                                                                  boolean isSSTableVersionBasedBridgeDisabled)
+    {
+        // Check for fallback mode
+        if (isSSTableVersionBasedBridgeDisabled)
```
sarankk: Nit: this conf check is duplicated both in read and write path, could we extract it out?

cassandra-analytics-common/src/test/java/org/apache/cassandra/bridge/CassandraVersionTest.java
```diff
@@ -0,0 +1,231 @@
...
+
+        result = CassandraVersion.fromSSTableVersion("bti-da");
+        assertThat(result).isPresent();
+        assertThat(result.get()).isEqualTo(CassandraVersion.FIVEZERO);
+    }
+
+    @Test
+    void testFromSSTableVersionThreeZeroNativeVersions()
+    {
+        Optional<CassandraVersion> result = CassandraVersion.fromSSTableVersion("big-ma");
```
sarankk: Could we make this parameterized test and test other SSTable versions in 3.0 too?

cassandra-analytics-common/src/test/java/org/apache/cassandra/bridge/CassandraVersionTest.java
```diff
@@ -0,0 +1,231 @@
...
+        // C* 4.1 did not introduce new native SSTable versions
+        List<String> nativeVersions = CassandraVersion.FOURONE.getNativeSStableVersions();
+        assertThat(nativeVersions).isEmpty();
+    }
+
+    @Test
+    void testGetSSTableVersionIndexValidVersion()
+    {
+        int index = CassandraVersion.FOURZERO.getSSTableVersionIndex("big-na");
+        assertThat(index).isEqualTo(0);
```
sarankk: We could avoid the index checks if we treat the SSTable versions within a Cassandra version on the same level.

cassandra-analytics-common/src/test/java/org/apache/cassandra/bridge/SSTableVersionAnalyzerTest.java
```diff
@@ -0,0 +1,219 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/**
+ * Unit tests for SSTableVersionAnalyzer
+ */
+public class SSTableVersionAnalyzerTest
+{
+    @Test
+    void testDetermineBridgeVersionForWriteFallbackDisabledSingleVersion()
```
sarankk: Nit: Could we make these tests parameterized, passing in expected result as well?

cassandra-analytics-core/src/main/java/org/apache/cassandra/bridge/SSTableVersionAnalyzer.java
```diff
@@ -0,0 +1,233 @@
...
+     *
+     * @param sstableVersionsOnCluster Set of SSTable versions to validate
+     * @throws IllegalStateException if versions are null or empty
+     */
+    private static void ensureSSTableVersionsNotEmpty(Set<String> sstableVersionsOnCluster)
+    {
+        if (sstableVersionsOnCluster == null || sstableVersionsOnCluster.isEmpty())
+        {
+            throw new IllegalStateException(String.format(
+                "Unable to retrieve SSTable versions from cluster. " +
```
sarankk: Why not support fallback here and return Cassandra version based bridge?

cassandra-analytics-sidecar-client/src/main/java/org/apache/cassandra/clients/Sidecar.java
```diff
@@ -149,6 +153,79 @@ public static List<CompletableFuture<NodeSettings>> allNodeSettings(SidecarClien
...
+     * Retrieve gossip info from all nodes on the cluster
+     * @param client Sidecar client
+     * @param instances all Sidecar instances
+     * @return completable futures with GossipInfoResponse
+     */
+    public static List<CompletableFuture<GossipInfoResponse>> gossipInfoFromAllNodes(SidecarClient client,
+                                                                                     Set<SidecarInstance> instances)
+    {
+        return instances.stream()
+                .map(instance -> client
```
sarankk: Nit: alignment

cassandra-analytics-sidecar-client/src/main/java/org/apache/cassandra/clients/Sidecar.java
```diff
@@ -149,6 +153,79 @@ public static List<CompletableFuture<NodeSettings>> allNodeSettings(SidecarClien
                         .collect(Collectors.toList());
     }
 
+    /**
+     * Retrieve gossip info from all nodes on the cluster
+     * @param client Sidecar client
+     * @param instances all Sidecar instances
+     * @return completable futures with GossipInfoResponse
+     */
+    public static List<CompletableFuture<GossipInfoResponse>> gossipInfoFromAllNodes(SidecarClient client,
```
sarankk, line 164: Not for addressing in this PR, but would be good to have a cluster wide gossip call available in Sidecar.

sarankk (commented):
Thanks Shailaja, took a first pass. Reviewed half the files.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
```diff
@@ -56,9 +58,21 @@ protected CassandraBulkWriterContext(@NotNull BulkWriterConfig config)
...
+    @Override
+    protected Set<String> getSSTableVersionsOnCluster(@NotNull BulkSparkConf conf)
+    {
+        return CassandraClusterInfo.getSSTableVersionsOnCluster(conf, null);
+    }
+
+    @Override
+    protected ClusterInfo buildClusterInfo(CassandraVersion bridgeVersion)
+    {
+        return new CassandraClusterInfo(bulkSparkConf(), bridgeVersion);
```
sarankk: We seem to `buildCassandraContext` thrice among these 3 method, each cassandra context initializes its own sidecar client. We should avoid this and combine them.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
```diff
@@ -395,4 +366,61 @@ Map<String, ClusterInfo> clusterInfoByIdUnsafe()
...
+    {
+        CoordinatedWriteConf coordinatedWriteConf = conf.coordinatedWriteConf();
+        Preconditions.checkArgument(coordinatedWriteConf != null,
+                                    "CoordinatedWriteConf is required for multi-cluster operations");
+
+        Set<String> aggregatedSSTableVersions = new HashSet<>();
+        for (String clusterId : coordinatedWriteConf.clusters().keySet())
+        {
+            Set<String> sstableVersions = CassandraClusterInfo.getSSTableVersionsOnCluster(conf, clusterId);
+            aggregatedSSTableVersions.addAll(sstableVersions);
```
sarankk: the check we do comparing lowest and highest Cassandra version in `getLowestCassandraVersion` is missing when we get bridge based on SSTable versions.

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/CassandraVersion.java
skoppu22: Yes, regardless of which sstable version found in a given Cassandra version, we choose that cassandra version as long as there is no sstable version from higher cassandra version. So the order of sstable versions within a cassandra version doesn't matter. Updated the sorting logic not to depend on array index.

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/CassandraVersion.java
skoppu22: C* 5.0 supports can read C* 4.0, 4.1, 4.2 ... 
C* 4.x can read C* 3.0, 3.1, ..
We cannot add all minor versions and won't be able to keep up with C* release cycle if we do so. Also this PR goal is to make C* analytics independent of C* version numbers. Hence dynamically calculating this makes analytics independent of C* releases.

cassandra-analytics-common/src/test/java/org/apache/cassandra/bridge/CassandraVersionTest.java
skoppu22: Removed this test

cassandra-analytics-core/src/main/java/org/apache/cassandra/bridge/SSTableVersionAnalyzer.java
skoppu22: Here we are asking user to run the job using fallback mechanism if they thinks that's appropriate. We do not use fallback mechanism when spark.cassandra_analytics.bridge.disable_sstable_version_based is false. There must be a reason why sstable version based bridge selection is failing. User can evaluate the cluster state and try with fallback mechanism only if they think that's fine for their use case or cluster situation.

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/SSTableVersionAnalyzer.java
skoppu22: We should error and abort job if encountered unrecognized sstable versions. We can log error and ask the user to run the job using fallback mechanism if the user thinks it is safe to run the job. I added error to log this.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
skoppu22: We do that above in getLowestCassandraVersion function, line 393 sorts versions, line 396 picks the lowest and line 397 picks the highest.
Whereas this function is getSSTableVersionsOnCluster, to get all sstable versions on the cluster. Then bridge determination logic sorts them, picks highest version and determines bridge version accordingly.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraBulkWriterContext.java
skoppu22: Looking into the possibility, seems difficult as we need to build context once and store somewhere to reuse it.

skoppu22: Fixed this. There is a chicken and egg problem here. Context is needed to determine the bridge, and context contains determined bridge version. Hence creating preliminary context initially (with null bridge version) which is used for bridge determination, then we create final context with the bridge value. Also made changes to reuse nodesettings.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
```diff
@@ -84,35 +85,37 @@ public class CassandraClusterInfoGroup implements ClusterInfo, MultiClusterSuppo
...
 
     /**
      * Creates {@link CassandraClusterInfoGroup} with the list of {@link ClusterInfo} from {@link BulkSparkConf} and validation
      * The validation ensures non-empty list of {@link ClusterInfo}, where all objects have non-empty and unique clusterId
      * @param conf bulk write conf
+     * @param bridgeVersion bridge version (nullable for preliminary construction)
      * @return new {@link CassandraClusterInfoGroup} instance
      */
-    public static CassandraClusterInfoGroup fromBulkSparkConf(BulkSparkConf conf)
+    public static CassandraClusterInfoGroup fromBulkSparkConf(BulkSparkConf conf, CassandraVersion bridgeVersion)
```
lukasz-antoniak: NIT, `bridgeVersion` seems always passed `null`.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
```diff
@@ -272,7 +278,11 @@ private int initBulkReader(@NotNull ClientConfig options) throws ExecutionExcept
         NodeSettings nodeSettings = sidecar.nodeSettings().get();
         String cassandraVersion = getEffectiveCassandraVersionForRead(clusterConfig, nodeSettings);
         Partitioner partitioner = Partitioner.from(nodeSettings.partitioner());
-        bridge = CassandraBridgeFactory.get(cassandraVersion);
+
+        // Initialize SSTable versions and bridge version
+        CassandraVersion bridgeVersion = initializeSSTableVersionsAndBridgeVersion(cassandraVersion);
+        bridge = CassandraBridgeFactory.get(bridgeVersion);
```
lukasz-antoniak, line 284: Should we add here the log analogical to write-side, e.g. `"Selected bridge version: {}, SSTable versions: {}`?

In the `determineBridgeVersionForRead` we have below log which is not present in `determineBridgeVersionForWrite`.
```
LOGGER.debug("Determined bridge version {} for read based on SSTable versions on cluster: {}", bridgeVersion.versionName(), sstableVersionsOnCluster);
```
I would suggest to make both logs about determined sstable version "symmetric".

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
```diff
@@ -324,6 +334,76 @@ private int initBulkReader(@NotNull ClientConfig options) throws ExecutionExcept
...
+        // validation and serialization - needs no null handling. On executors an empty set is the
+        // signal that the feature was disabled on the driver.
+        if (isSSTableVersionBasedBridgeDisabled)
+        {
+            //Use a HashSet, because Kryo serializer reads back via kryo.readObject(in, HashSet.class)
+            this.sstableVersionsOnCluster = new HashSet<>();
+        }
+        else
+        {
+            this.sstableVersionsOnCluster = retrieveSSTableVersionsFromCluster();
```
lukasz-antoniak: `HashSet` is not guaranteed here, compare comment above.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkRoundtripSSTableVersionBridgeTestBase.java
```diff
@@ -0,0 +1,268 @@
...
+    {
+        // Set before CassandraVersion class initialization (which reads this property once, statically) so the bulk
+        // writer generates SSTables in the configured format. This runs at the start of cluster provisioning, before
+        // any analytics bridge usage, and relies on forkEvery = 1 (the integration-test default) for a fresh JVM per
+        // test class.
+        System.setProperty("cassandra.analytics.bridges.sstable_format", sstableFormat());
+        if ("bti".equals(sstableFormat()))
+        {
+            // BTI (bti-da) is a Cassandra 5.0+ format; skip on older versions.
+            assumeTrue(testVersion.version().startsWith("5."),
```
lukasz-antoniak: NIT: Just to support Cassandra 6, maybe we can do:
```
Semver version = new Semver(testVersion.version(), Semver.SemverType.LOOSE);
assumeTrue(version.isGreaterThanOrEqualTo(new Semver("5.0", Semver.SemverType.LOOSE)),
           "BTI format (bti-da) requires Cassandra 5.0+, but test version is " + testVersion.version());
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -117,6 +118,9 @@ public class BulkSparkConf implements Serializable
     // Cassandra version of target cluster. Configuration parameter is exposed to be able to correctly initialize static
     // components, before cluster version is discovered via Sidecar.
     public static final String CASSANDRA_VERSION                       = SETTING_PREFIX + "cassandra.version";
+    // Disable SSTable version-based bridge determination. When true, falls back to using cassandra.version for bridge selection.
+    // This provides a safety fallback mechanism if SSTable version detection fails or encounters issues.
+    public static final String DISABLE_SSTABLE_VERSION_BASED_BRIDGE   = SETTING_PREFIX + "bridge.disable_sstable_version_based";
```
yifan-c: 1. add one more space before `=` to align with the other lines. 
2. The exact same config is defined again in SSTableVersionAnalyzer. It is error-prone. The spark config key should be defined in one single place, and it should be declared in `BulkSparkConf`. 
3. `SSTableVersionAnalyzer` should be relocated to `anayltics-core` subproject. The reason is that the `common` subproject should be spark-agnostic. `SSTableVersionAnalyzer` currently reference to spark config key. Plus, the original version determination code is in `analytics-core` too.

cassandra-analytics-core/src/main/java/org/apache/cassandra/bridge/SSTableVersionAnalyzer.java
yifan-c, line 1: `SSTableVersionAnalyzer` should live in `cassandra-analytics-core`, not `common`. Its only callers (`AbstractBulkWriterContext`, `CassandraDataLayer`) are in core, and the version-determination logic it replaces (`CassandraClusterInfo.getVersionFromSidecar`) has always been in core. 

Placing it in common also forces the `DISABLE_SSTABLE_VERSION_BASED_BRIDGE` Spark config key into the common module, which should remain Spark-agnostic. Move `SSTableVersionAnalyzer` (and its test) to core, define the constant solely in `BulkSparkConf`, and reference `BulkSparkConf.DISABLE_SSTABLE_VERSION_BASED_BRIDGE` from the analyzer's error messages.

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/SSTableVersionAnalyzer.java
```diff
@@ -0,0 +1,247 @@
...
+                                                                  String requestedFormat,
+                                                                  String cassandraVersion,
+                                                                  boolean isSSTableVersionBasedBridgeDisabled)
+    {
+        // Check for fallback mode
+        Optional<CassandraVersion> fallback = resolveFallbackVersion(cassandraVersion, isSSTableVersionBasedBridgeDisabled);
+        if (fallback.isPresent())
+        {
+            return fallback.get();
+        }
```
yifan-c: The method `resolveFallbackVersion` is misleading. "Fallback" implies the method is triggered when SSTable version-based selection has failed, but it actually represents the original pre-patch behavior — bridge selection driven by the Cassandra release version — which the user opts into explicitly via configuration. Nothing has failed when this path is taken.

I'd suggest rename the method to `resolveLegacyVersionBasedBridge` and the variable to `version`.

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/SSTableVersionAnalyzer.java
```diff
@@ -0,0 +1,247 @@
...
+     * @throws IllegalStateException if highest version is unknown
+     */
+    private static CassandraVersion findHighestCassandraVersion(Set<String> sstableVersionsOnCluster)
+    {
+        String highestSSTableVersion = findHighestSSTableVersion(sstableVersionsOnCluster);
+        return CassandraVersion.fromSSTableVersion(highestSSTableVersion)
+                               .orElseThrow(() -> new IllegalStateException(
+                               String.format("Unknown SSTable version: %s. Cannot determine bridge version. " +
+                                             "SSTable versions on cluster: %s. " +
+                                             "To retry the job using a fallback Cassandra version, " +
```
yifan-c: The "fallback" wording is confusing. Refer to my previous comment for the reasoning. 
There is no fallback that happens automatically; therefore it is not a fallback. 
If my understanding is correct, it is to ask user to retry with the original cassandra.version-based bridge selection. Please update the error message to reflect the fact.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkRoundtripSSTableVersionBridgeTestBase.java
skoppu22: Done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
skoppu22: Done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
skoppu22: Write had log message in the caller. Now removed that and added comment in determineBridgeVersionForWrite. Also made log messages in determineBridgeVersionForRead and determineBridgeVersionForWrite as info as it gets printed only once in driver and helps in debugging.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
skoppu22: Removed bridgeVersion param

skoppu22:
Latest circle CI https://app.circleci.com/pipelines/github/skoppu22/cassandra-analytics/150/workflows/c112514d-49f8-44f3-a605-cb6a36ca239f

lukasz-antoniak:
Using this branch I was able to bundle DS Cassandra bridges that process `bti-cc` sstable version (but register themselves as 4.x / 5.x version), which was not possible before.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -447,22 +413,35 @@ private TokenRangeMapping<RingInstance> getTokenRangeReplicasFromSidecar()
                                         metadata -> new RingInstance(metadata, clusterId()));
     }
 
-    public String getVersionFromFeature()
```
yifan-c, line 450: The behavior is changed unexpectedly with the method deletion.
`getVersionFromFeature` is used in `getLowestCassandraVersion` to allow version override via feature flag. The override behavior should be kept, i.e. a feature flag can override the determined cassandra version from the sstable analyzer. 

This patch should only introduce the new sstable based cassandra version determination logic.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -390,33 +383,6 @@ public TokenRangeMapping<RingInstance> getTokenRangeMapping(boolean cached)
...
-        synchronized (this)
-        {
-            if (cassandraVersion == null)
-            {
-                String versionFromFeature = getVersionFromFeature();
-                if (versionFromFeature != null)
-                {
-                    // Forcing writer to use a particular version
-                    cassandraVersion = versionFromFeature;
-                }
```
yifan-c, line 411: The version override function is removed accidentally.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/AbstractBulkWriterContext.java
```diff
@@ -98,10 +101,32 @@ protected AbstractBulkWriterContext(@NotNull BulkSparkConf conf,
...
+        // Get SSTable versions from cluster only if SSTable version-based selection is enabled
+        // If disabled, skip retrieval to allow job to proceed even when SSTable version detection fails
+        if (!conf.isSSTableVersionBasedBridgeDisabled())
+        {
+            sstableVersionsOnCluster = getSSTableVersionsOnCluster(conf);
+        }
+
+        // Determine bridge version
+        this.bridgeVersion = SSTableVersionAnalyzer.determineBridgeVersionForWrite(
+            sstableVersionsOnCluster,
```
yifan-c: `sstableVersionsOnCluster` can be `null` when `isSSTableVersionBasedBridgeDisabled == true`. But `SSTableVersionAnalyzer.determineBridgeVersionForWrite` disallows `null` value. 
In fact, you may not need the boolean parameter `isSSTableVersionBasedBridgeDisabled` in the method `SSTableVersionAnalyzer.determineBridgeVersionForWrite`. When `sstableVersionsOnCluster` is null, it already indicate the feature is disabled. 

Can you update the corresponding code paths to simplify the implementation?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/AbstractBulkWriterContext.java
```diff
@@ -98,10 +101,32 @@ protected AbstractBulkWriterContext(@NotNull BulkSparkConf conf,
...
+            CassandraVersion.configuredSSTableFormat(),
+            lowestCassandraVersion,
+            conf.isSSTableVersionBasedBridgeDisabled()
+        );
+
+        // Validate that Kryo registrator exists for this bridge version
+        KryoRegister.validateKryoRegistratorExists(this.bridgeVersion, lowestCassandraVersion);
```
yifan-c: I believe KryoRegistrator is not really the concern; instead, the real concern is to ensure the bridgeVersion determined is recognized. 

To make the API clean, I'd rather suggest the change to ensure that `SSTableVersionAnalyzer` always returns a recognized bridgeVersion or throw when unable to determine.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
```diff
@@ -398,9 +369,62 @@ private RuntimeException toRuntimeException(ClusterInfo clusterInfo, Throwable c
...
+
+        CassandraVersionFeatures first = versions.get(0);
+        CassandraVersionFeatures last = versions.get(versions.size() - 1);
+        Preconditions.checkState(first.getMajorVersion() == last.getMajorVersion(),
+                                 "Cluster versions are not compatible. lowest=%s and highest=%s",
+                                 first.getRawVersionString(), last.getRawVersionString());
+
+        return first.getRawVersionString();
+    }
+
```
yifan-c: Why is the original implementation `getLowestCassandraVersion` in this class changed?

I do not see an obvious benefit; but the unrelated change is making code review hard. Please revert the change, unless there is a pressing reason.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
```diff
@@ -398,9 +369,62 @@ private RuntimeException toRuntimeException(ClusterInfo clusterInfo, Throwable c
...
+    public Set<String> getSSTableVersionsOnCluster()
+    {
+        Set<String> aggregatedSSTableVersions = new HashSet<>();
+        for (ClusterInfo ci : clusterInfos)
+        {
+            CassandraClusterInfo cci = (CassandraClusterInfo) ci;
+            aggregatedSSTableVersions.addAll(cci.getSSTableVersionsOnCluster());
+        }
+        return aggregatedSSTableVersions;
     }
```
yifan-c, line 409: Follow the pattern in the original `getLowestCassandraVersion`, i.e. return cached if available, otherwise get first, finally aggregate when there are multiple.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
```diff
@@ -211,38 +214,6 @@ public TokenRangeMapping<RingInstance> getTokenRangeMapping(boolean cached)
...
-        List<CassandraVersionFeatures> versions = aggregated.values()
-                                                            .stream()
-                                                            .map(CassandraVersionFeatures::cassandraVersionFeaturesFromCassandraVersion)
-                                                            .sorted()
-                                                            .collect(Collectors.toList());
-        CassandraVersionFeatures first = versions.get(0);
-        CassandraVersionFeatures last = versions.get(versions.size() - 1);
-        Preconditions.checkState(first.getMajorVersion() == last.getMajorVersion(),
-                                 "Cluster versions are not compatible. lowest=%s and highest=%s",
-                                 first.getRawVersionString(), last.getRawVersionString());
```
yifan-c, line 241: This check should be lifted when sstable-based bridge determination is enabled. Mismatching major version no longer mean incompatibility necessarily.

yifan-c (changes requested):
Submit since there are a handful of comments already. Note that it is not a full review.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
skoppu22: On the cluster, if one node is running 5.0 and another node running 4.x, the bulk writer selects highest supported format 5.0 as bridge and writes BTI or BIG-oa sstables. 4.x node will fail to read them. This will impact live node operations reading that table as well. 

The assumption behind this new feature is 
- either all nodes in a cluster are in 4.x version
- or all nodes in a cluster in 5.0 version and SCM is cassandra_4 or upgrading or none or combination of these.
In both cases, lowest and highest major version should match. Hence we need to have this check in bridge selection logic for write as well, which I missed adding.

Other option is, we need to select lowest supported sstable format for write, so all nodes understands the sstables being written. With this we need to allow lowest and highest major versions to differ as long as lowest sstable version can be read on highest cassandra version node. i.e, we need to change bridge determination logic to lowest sstable version for bulk write and highest sstable version for bulk read.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/AbstractBulkWriterContext.java
skoppu22: - determineBridgeVersionForWrite first checks for the isSSTableVersionBasedBridgeDisabled, if true uses fallback logic without looking at sstableVersionsOnCluster value. 
- if isSSTableVersionBasedBridgeDisabled is false, then we must have retrieved sstableVersionsOnCluster and there should be atleast one sstable on the cluster, hence we shouldn't have it null or empty.
- We have feature flag available here, better to use it instead of hacking on sstableVersionsOnCluster being null which can be error case as well by any chance.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/AbstractBulkWriterContext.java
skoppu22: We added this call KryoRegister.validateKryoRegistratorExists when we were adding only one cassandra.version registrator class to KRYO_SERIALIZERS, which may not be the same as determined bridgerversion. Now I have modified KryoRegister to add all implemented registrator classes to KRYO_SERIALIZERS, so this check is not needed. We can leave it as it is just to verify upfront or we can remove it.

SSTableVersionAnalyzer already ensures only recognized bridgeversion is returned.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
skoppu22: Not accidentally, I thought this may not be needed after bridge determination feature. Will restore it.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
skoppu22: With bridge determination feature getLowestCassandraVersion is only used for logging in the driver. Executors doesn't need it. If I remember correctly (opened PR more than a month ago, might have forgotten a bit), I might have modified getLowestCassandraVersion as one time call as not needed on executors. Let me try to recollect and restore it.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
skoppu22: Sorry, now I realize I overlooked this one line `// Forcing writer to use a particular version`
I was just removing getLowestCassandraVersion to make it just one time call from the driver as executors doesn't need it with bridge determination feature. Restoring getVersionFromFeature

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
skoppu22: Now lowestCassandraVersion is driver only, not needed on executors. Hence rewritten getLowestCassandraVersion for one time call.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
skoppu22: getSSTableVersionsOnCluster is called only on driver. Executors get determined bridge version. So there is no cached value to return.

cassandra-analytics-core/src/main/java/org/apache/cassandra/bridge/SSTableVersionAnalyzer.java
skoppu22: Done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
yifan-c: Please restore it

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/AbstractBulkWriterContext.java
yifan-c: Please remove the unnecessary code. In general, less lines of code the better.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
skoppu22: Done

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/SSTableVersionAnalyzer.java
skoppu22: Done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
yifan-c: It is not true. Cassandra version should be determined on task start (on executors) and fail the task early.

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/SSTableVersionAnalyzer.java
skoppu22: Done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
skoppu22: I have restored getVersionFromFeature. But old style getLowestCassandraVersion with cache value is not needed as it now gets invoked only on driver

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
skoppu22: Restored getVersionFromFeature call from getLowestCassandraVersion. Note that only legacy fallback uses it, bridge determination feature doesn't use getVersionFromFeature

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/AbstractBulkWriterContext.java
skoppu22: Removed

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
yifan-c: The version determination logic in the `SSTableVersionAnalyzer` in wrong. Writer should pick the lowest compatible bridge; reader should pick the highest compatible bridge. Please check out the fix in https://github.com/yifan-c/cassandra-analytics/commits/simplify-sstable-bridge/

skoppu22: - Create getBridgeVersion which first picks version override if available, otherwise next picks bridge based on sstables version (if feature enabled), otherwise picks bridge legacy style
- Modified determineBridgeVersionForWrite to pick lowest version as bridge, this picks lowest compatible version for a cluster
- CassandraClusterInfoGroup's getBridgeVersion picks lowest compatible version across clusters and verifies highest version can read sstables written by selected bridge version
- Modified CassandraVersion enum to have lowestCompatibleVersionNumber instead of calculating dynamically
- Modified bulk reader's validateSStableVersions (runs on executors) to ensure bridge version can support sstables seen (previously was just checking sstable versions seen are same as driver seen), this allows different sstable versions to be written by the time executors run and works as long as bridge can support them

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
skoppu22: Previously, only driver calculated bridge version and broadcasted to executors. Now the code is reverted to existing style

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
skoppu22: Now the code changed to pick getVersionFromFeature as priority no matter bridge feature enabled or disabled.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
skoppu22: Existing style restored

cassandra-analytics-sidecar-client/src/main/java/org/apache/cassandra/clients/Sidecar.java
```diff
@@ -149,6 +153,79 @@ public static List<CompletableFuture<NodeSettings>> allNodeSettings(SidecarClien
...
+                                                             long maxRetryDelayMillis,
+                                                             int maxRetries)
+    {
+        LOGGER.debug("Retrieving SSTable versions from cluster via gossip...");
+
+        List<CompletableFuture<GossipInfoResponse>> gossipInfoFutures =
+            gossipInfoFromAllNodes(client, instances);
+
+        // Calculate total timeout
+        final long totalTimeout = maxRetryDelayMillis * maxRetries * gossipInfoFutures.size();
```
yifan-c: The total timeout indicates the sequential execution, but the requests are indeed executed in parallel. I think you want to remove `gossipInfoFutures.size()` in the calculation.

cassandra-analytics-sidecar-client/src/main/java/org/apache/cassandra/clients/Sidecar.java
```diff
@@ -149,6 +153,79 @@ public static List<CompletableFuture<NodeSettings>> allNodeSettings(SidecarClien
...
+     * @param instances all Sidecar instances in the cluster
+     * @param maxRetryDelayMillis maximum delay in milliseconds between retries
+     * @param maxRetries maximum number of retry attempts
+     * @return a set of SSTable versions across all nodes in the cluster
+     * @throws RuntimeException if unable to retrieve gossip info from any nodes
+     */
+    public static Set<String> getSSTableVersionsFromCluster(SidecarClient client,
+                                                             Set<SidecarInstance> instances,
+                                                             long maxRetryDelayMillis,
+                                                             int maxRetries)
```
yifan-c: This and many other changes in the patch do not follow the codestyle of the project. 
You can run `./gradlew idea` to re-generate the project, and run `crtl + option + i` (or in the menu `Code -> Auto-Indent Lines`) in each changed files.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/MockBulkWriterContext.java
```diff
@@ -378,9 +379,17 @@ public void checkBulkWriterIsEnabledOrThrow()
...
+        return CassandraVersion.fromVersion(cassandraVersion).orElse(CassandraVersion.FIVEZERO);
+    }
+
+    public Set<String> getSSTableVersionsOnCluster()
+    {
+        // Return SSTable versions based on Cassandra version for testing
+        CassandraVersion version = CassandraVersion.fromVersion(cassandraVersion)
+            .orElse(CassandraVersion.FIVEZERO);
+        return new HashSet<>(version.getNativeSStableVersions());
     }
```
yifan-c: not used

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/CassandraVersion.java
```diff
@@ -78,13 +109,69 @@ public String jarBaseName()
...
+    {
+        return sstableFormats;
+    }
+
+    /**
+     * Get the list of native SSTable version strings for this Cassandra version.
+     *
+     * @return List of native SSTable version strings
+     */
+    public List<String> getNativeSStableVersions()
```
yifan-c: `SStable` -> `SSTable`

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/CassandraVersion.java
```diff
@@ -78,13 +109,69 @@ public String jarBaseName()
...
+
+    /**
+     * Get the set of SSTable version strings that this Cassandra version can read.
+     * This includes the native versions of every Cassandra version from {@link #lowestCompatibleVersionNumber}
+     * (inclusive) up to this version (inclusive). For example, Cassandra 5.0 can read 4.0, 4.1 and 5.0
+     * SSTables (lowestCompatible=40) but NOT 3.x.
+     *
+     * @return Set of full SSTable version strings that can be read
+     */
+    public Set<String> getSupportedSStableVersionsForRead()
```
yifan-c: `SStable` -> `SSTable`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
```diff
@@ -844,6 +934,67 @@ private static String readNullable(ObjectInputStream in) throws IOException
...
+    /**
+     * Validates that all SSTables being read have versions that were observed in gossip info.
+     * This catches cases where SSTables have unexpected versions that weren't seen during driver initialization.
+     * This validation runs on executors.
+     *
+     * @param sstables list of SSTables to validate
+     * @throws UnsupportedOperationException if any SSTable has a version not in expected sstable versions
+     */
+    @VisibleForTesting
+    void validateSStableVersions(List<SSTable> sstables)
```
yifan-c: `SStable` -> `SSTable`

cassandra-analytics-core/src/main/java/org/apache/cassandra/bridge/SSTableVersionAnalyzer.java
```diff
@@ -0,0 +1,199 @@
...
+    public static CassandraVersion determineBridgeVersionForWrite(Set<String> sstableVersionsOnCluster, String requestedFormat)
+    {
+        requireSSTableVersionsPresent(sstableVersionsOnCluster);
+
+        // Every observed version must be mutually compatible (readable by the highest version present).
+        ensureMutuallyCompatibleVersions(sstableVersionsOnCluster, highestCompatibleVersion(sstableVersionsOnCluster));
+
+        CassandraVersion bridgeVersion = cassandraVersionsFromSSTableVersions(sstableVersionsOnCluster)
+                                         .min(Comparator.comparingInt(CassandraVersion::versionNumber))
+                                         .orElseThrow(() -> new IllegalStateException("Unable to determine the lowest SSTable version"));
```
yifan-c: In this method, `sstableVersionsOnCluster` is converted to `CassandraVersion` twice.

lukasz-antoniak:
What do you think about adding integration test [like this one](https://github.com/lukasz-antoniak/cassandra-analytics/pull/20/changes/37b5c387c07bc830fff099cb2fb39e7863d15823) to simulate write with 2 different sstable version and reader able to select proper bridge version?

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/KryoRegisterTest.java
```diff
@@ -0,0 +1,95 @@
...
+        // setup() must register a registrator for every implemented (bundled) bridge version,
+        // independent of spark.cassandra_analytics.cassandra.version, so serialization works for
+        // whichever bridge the SSTable-version analyzer selects at runtime.
+        assertThat(expected).isNotEmpty();
+        List<String> registrators = Arrays.asList(conf.get("spark.kryo.registrator").split(","));
+        assertThat(registrators).containsAll(expected);
+    }
+
+    @Test
+    void testSetupDoesNotDependOnCassandraVersionConfig()
```
lukasz-antoniak: I think that property `spark.cassandra_analytics.cassandra.version` should be remove as it will not be used any more. We can remove all occurrences from code and comments.

cassandra-analytics-sidecar-client/src/main/java/org/apache/cassandra/clients/Sidecar.java
lukasz-antoniak: Is the total timeout calculated correctly? In both methods - `getAllNodeSettings()` and `getSSTableVersionsFromCluster()` - we are calculating timeout for completion of parallel requests as `maxRetryDelayMillis * maxRetries * futuresSize`. Request timeout is not taken into account. Should not it be `maxRetryDelayMillis * maxRetries * requestTimeout`?

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/KryoRegisterTest.java
skoppu22: Removed

skoppu22:
> What do you think about adding integration test [like this one](https://github.com/lukasz-antoniak/cassandra-analytics/pull/20/changes/37b5c387c07bc830fff099cb2fb39e7863d15823) to simulate write with 2 different sstable version and reader able to select proper bridge version?

Good one, added

cassandra-analytics-core/src/main/java/org/apache/cassandra/bridge/SSTableVersionAnalyzer.java
skoppu22: Fixed this

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/CassandraDataLayer.java
skoppu22: Done

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/CassandraVersion.java
skoppu22: Done

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/CassandraVersion.java
skoppu22: Done

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/MockBulkWriterContext.java
skoppu22: Removed

cassandra-analytics-sidecar-client/src/main/java/org/apache/cassandra/clients/Sidecar.java
skoppu22: Done

cassandra-analytics-sidecar-client/src/main/java/org/apache/cassandra/clients/Sidecar.java
skoppu22: This is copied from getAllNodeSettings, which calculates worst case timeout, which is only upper bound. I tried removing gossipInfoFutures.size() from multiplication, then in-jvm tests failed because unable to retrieve sstable versions before timeout. Perhaps in-jvm specific settings could be problem, but don't want to risk the same problem on real cluster.

yifan-c: @lukasz-antoniak had a good point. The calculation is missing the request timeout. As a minor correction, the worst scenario is `(maxRetryDelayMillis + requestTimeout) * maxRetries`. It is the worst scenario because it calculates with the max retry delay; meanwhile, exponential retry is used in the production system. 

Serial execution is not expected. The http client executes the requests in parallel. The original calculation is bug. It is likely just shaded by the overly large value from the additional multiplication.

cassandra-analytics-common/src/test/java/org/apache/cassandra/bridge/CassandraVersionTest.java
```diff
@@ -0,0 +1,163 @@
...
+    void testConfiguredSSTableFormatDefault()
+    {
+        // Assuming no system property set, should return "big"
+        String format = CassandraVersion.configuredSSTableFormat();
+        assertThat(format).isNotNull();
+        assertThat(format).isIn("big", "bti"); // Could be either depending on env
+    }
+
+    @Test
+    void testSStableFormatsFourZero()
```
lukasz-antoniak: NIT: `SSTable`

cassandra-analytics-common/src/test/java/org/apache/cassandra/bridge/CassandraVersionTest.java
```diff
@@ -0,0 +1,163 @@
...
+    {
+        Set<String> formats = CassandraVersion.FOURZERO.sstableFormats();
+        assertThat(formats).containsExactly("big");
+    }
+
+    @Test
+    void testSStableFormatsFiveZero()
```
lukasz-antoniak: NIT: `SSTable`

cassandra-analytics-core-example/src/main/java/org/apache/cassandra/spark/example/DirectWriteAndReadJob.java
```diff
@@ -36,7 +36,6 @@ public static void main(String[] args)
     {
         System.setProperty("SKIP_STARTUP_VALIDATIONS", "true");
         // Configuration for Cassandra 5.x (and optionally BTI sstable format):
```
lukasz-antoniak: NIT: Maybe change to `Optionally select BTI sstable format:`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -114,9 +114,9 @@ public class BulkSparkConf implements Serializable
     //       which will throw a configuration exception for each setting with that prefix it does not recognize
     public static final String SETTING_PREFIX = "spark.cassandra_analytics.";
 
-    // Cassandra version of target cluster. Configuration parameter is exposed to be able to correctly initialize static
-    // components, before cluster version is discovered via Sidecar.
-    public static final String CASSANDRA_VERSION                       = SETTING_PREFIX + "cassandra.version";
+    // Disable SSTable version-based bridge determination. When true, falls back to legacy mode.
```
lukasz-antoniak: NIT: Maybe just to be clear:
`Disable SSTable version-based bridge determination. When true, falls back to legacy mode, which selects bridge based on Cassandra version returned by the Sidecar.`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -515,9 +557,28 @@ public String getLowestVersion(List<NodeSettings> allNodeSettings)
         return ns.releaseVersion();
     }
 
+    /**
+     * Retrieves SSTable versions using the existing cassandraContext.
```
lukasz-antoniak: NIT: `CassandraContext`

cassandra-analytics-core/src/test/java/org/apache/cassandra/bridge/SSTableVersionAnalyzerTest.java
```diff
@@ -0,0 +1,178 @@
...
+    // --- determineBridgeVersionForWrite: picks the LOWEST mutually-compatible version ---
+
+    static Stream<Arguments> writeLowestVersionCases()
+    {
+        return Stream.of(
+        Arguments.of(Collections.singleton("big-oa"), "big", CassandraVersion.FIVEZERO),
+        Arguments.of(new HashSet<>(Arrays.asList("big-na", "big-nb")), "big", CassandraVersion.FOURZERO),
+        // mixed 4.0 + 5.0 sstables: lowest that every node can import is 4.0
+        Arguments.of(new HashSet<>(Arrays.asList("big-na", "big-oa")), "big", CassandraVersion.FOURZERO),
+        Arguments.of(Collections.singleton("bti-da"), "bti", CassandraVersion.FIVEZERO)
```
lukasz-antoniak, line 50: Just wondering, can there be a possibility where we see sstable versions on cluster as "big-na", because we have just upgraded to C* 5.x and at the same time "bti" format is configured.
```
Arguments.of(Collections.singleton("big-na"), "bti", CassandraVersion.FIVEZERO)
```

cassandra-analytics-core/src/test/java/org/apache/cassandra/bridge/SSTableVersionAnalyzerTest.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    // --- determineBridgeVersionForRead: picks the HIGHEST mutually-compatible version ---
+
+    static Stream<Arguments> readHighestVersionCases()
+    {
+        return Stream.of(
+        Arguments.of(Collections.singleton("big-oa"), CassandraVersion.FIVEZERO),
+        Arguments.of(new HashSet<>(Arrays.asList("big-na", "big-nb")), CassandraVersion.FOURZERO),
+        Arguments.of(new HashSet<>(Arrays.asList("big-na", "big-oa")), CassandraVersion.FIVEZERO)
```
lukasz-antoniak, line 123: NIT: Maybe we could add: `Arguments.of(new HashSet<>(Arrays.asList("big-na", "bti-da")), CassandraVersion.FIVEZERO)`

cassandra-analytics-sidecar-client/src/main/java/org/apache/cassandra/clients/Sidecar.java
skoppu22: Done, not updated getAllNodeSettings in this PR, as we wanted to have only feature specific changes in this PR

cassandra-analytics-core/src/test/java/org/apache/cassandra/bridge/SSTableVersionAnalyzerTest.java
skoppu22: This is possible I believe. In that case, bulk writer bridge selected would be 4.0 and that bridge cannot write in BTI format, so the user sees an error suggesting to switch back to legacy mode. Added test for the same.

cassandra-analytics-core/src/test/java/org/apache/cassandra/bridge/SSTableVersionAnalyzerTest.java
skoppu22: Done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
skoppu22: Done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
skoppu22: Done

cassandra-analytics-core-example/src/main/java/org/apache/cassandra/spark/example/DirectWriteAndReadJob.java
skoppu22: Done

cassandra-analytics-common/src/test/java/org/apache/cassandra/bridge/CassandraVersionTest.java
skoppu22: Done

cassandra-analytics-common/src/test/java/org/apache/cassandra/bridge/CassandraVersionTest.java
skoppu22: Done

cassandra-analytics-sidecar-client/src/main/java/org/apache/cassandra/clients/Sidecar.java
lukasz-antoniak: Created [CASSANALYTICS-179](https://issues.apache.org/jira/browse/CASSANALYTICS-179), I will fix it tomorrow.

skoppu22: Thank you

cassandra-analytics-core/src/test/java/org/apache/cassandra/bridge/SSTableVersionAnalyzerTest.java
yifan-c: I doubt if the scenario will happen in practice. 

If it happens, should it decide to write 4.0 SSTables, instead of exiting? 4.0 sstables can be read by 5.0 server using BTI format.

skoppu22: Bridge determined will be 4.0. But we will not be able to honor the user requested sstable format 'BTI' in that case, hence we error. The error printed will be as below. Perhaps we need to reword this to avoid 'Cluster does not support'
 `"Cluster does not support requested SSTable format BTI. Bridge version determined is 4.0, which only supports formats: BIG"`

yifan-c: Sounds good. I actually misread the test parameters as mixed sstables of na and BIT. Now I stay corrected. The revised error message looks clear and good.

skoppu22: Updated error message.

yifan-c (approved):

yifan-c:
@lukasz-antoniak, please let me know if you are +1 on merging. I will hold on merging until tomorrow (July 9)

lukasz-antoniak (approved):

skoppu22:
This code merged manually, so closing this PR

skoppu22:
This code merged manually, so closing this PR

PR #172, opened 2026-02-14
CASSANALYTICS-120: Quote identifiers option must be set to true if ttl has mixed case column name

sarankk:
If users set TTL column name or timestamp column name to be mixed case, they have to set quote identifiers option to true for Cassandra to resolve the column name correctly. Sometimes users miss setting quote identifiers option and perRowTTL setting incorrectly overrides their existing TTL with null values when bulk write job runs. This patch fixes this, by checking for mixed case column names and setting quote identifiers option when required.

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
@@ -53,8 +53,6 @@ import org.apache.spark.SparkConf;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import static java.lang.Character.isUpperCase;
-
 @SuppressWarnings("WeakerAccess")
 public class BulkSparkConf implements Serializable
 {
@@ -209,7 +207,6 @@ public class BulkSparkConf implements Serializable
         this.ttl = MapUtils.getOrDefault(options, WriterOptions.TTL.name(), null);
         this.timestamp = MapUtils.getOrDefault(options, WriterOptions.TIMESTAMP.name(), null);
         this.quoteIdentifiers = MapUtils.getBoolean(options, WriterOptions.QUOTE_IDENTIFIERS.name(), false, "quote identifiers");
-        validateUserAddedColumns(this.ttl, this.timestamp);
         int storageClientConcurrency = MapUtils.getInt(options, WriterOptions.STORAGE_CLIENT_CONCURRENCY.name(),
                                                        DEFAULT_STORAGE_CLIENT_CONCURRENCY, "storage client concurrency");
         long storageClientKeepAliveSeconds = MapUtils.getLong(options, WriterOptions.STORAGE_CLIENT_THREAD_KEEP_ALIVE_SECONDS.name(),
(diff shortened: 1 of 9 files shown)


sarankk:
Thanks @frankgh for the review, good spot. Addressed your comment. Instead of automatically setting quote identifiers option to true, added a validation instead to throw if users use mixed case column names without quote identifiers. With current behavior, when perRow TTL option is set with mixed case TTL column name and without quote identifiers, the job succeeds but silently rewrites existing TTLs of the rows to null. Hence throwing to fix this.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -364,6 +367,43 @@ protected void validateEnvironment() throws RuntimeException
...
+    }
+
+    private String columnValidationErrMsg(String option)
+    {
+        return String.format("Prefer setting lowercase chars for %s option. If uppercase is required, set %s to " +
+                             "true. Note %s option will be applied for all column names", option,
+                             WriterOptions.QUOTE_IDENTIFIERS.name(), WriterOptions.QUOTE_IDENTIFIERS.name());
+    }
+
+    private boolean doesNotHaveUpperCaseChars(String name)
```
skoppu22: Not sure is there a existing Java utility function to o this check. Java experts may comment here.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -364,6 +367,43 @@ protected void validateEnvironment() throws RuntimeException
...
+        Preconditions.checkArgument(doesNotHaveUpperCaseChars(ttlColName),
+                                    columnValidationErrMsg(WriterOptions.TTL.name()));
+        Preconditions.checkArgument(doesNotHaveUpperCaseChars(timestampColName),
+                                    columnValidationErrMsg(WriterOptions.TIMESTAMP.name()));
+    }
+
+    private String columnValidationErrMsg(String option)
+    {
+        return String.format("Prefer setting lowercase chars for %s option. If uppercase is required, set %s to " +
+                             "true. Note %s option will be applied for all column names", option,
```
skoppu22: How about "Note %s option is applicable for all column names" ?

skoppu22:
+1 Looks good

frankgh (commented):
The approach now seems better than before. When we don't quote we validate that the TTL and Timestamp columns only have lowercase. However, this doesn't cover for the case of special characters. I suggest leveraging the `bridge.maybeQuoteIdentifier(String)` functionality to make a determination whether the input is valid or not

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
```diff
@@ -307,4 +311,44 @@ private static List<Integer> getKeyFieldPositions(StructType dfSchema,
                           .map(dfFieldNames::indexOf)
                           .collect(Collectors.toList());
     }
+
+    private static void validateUserAddedColumns(String lowestCassandraVersion, boolean quoteIdentifiers,
+                                                 TTLOption ttlOption, TimestampOption timestampOption)
+    {
+        CassandraBridge bridge = CassandraBridgeFactory.get(lowestCassandraVersion);
+        if (!quoteIdentifiers)
+        {
```
frankgh: NIT avoid getting an instance of the bridge when it's not needed
```suggestion
        if (!quoteIdentifiers)
        {
            CassandraBridge bridge = CassandraBridgeFactory.get(lowestCassandraVersion);
```

frankgh (commented):
Looks good. I have one minor suggestion, and we should be good

frankgh (approved):
+1

PR #179, opened 2026-03-10
CASSANALYTICS-127: Fixing CdcTests.testMockedCdc broken due to incorrect position update in BufferingCommitLogReader

jyothsnakonisa:


cassandra-analytics-cdc/src/test/java/org/apache/cassandra/cdc/LocalCommitLog.java
```diff
@@ -67,7 +70,21 @@ public long length()
...
+    private List<String> readIdxFile()
+    {
+        Path idxPath = path.resolveSibling(name.replace(".log", "_cdc.idx"));
+        try
+        {
+            return Files.exists(idxPath) ? Files.readAllLines(idxPath) : List.of();
+        }
+        catch (IOException e)
+        {
+            return List.of();
```
lukasz-antoniak: I would propagate a runtime exception, so that it is not silently ignored in case of failure?

lukasz-antoniak (approved):
+1. Just squash and change commit message.

cassandra-analytics-cdc/src/test/java/org/apache/cassandra/cdc/LocalCommitLog.java
```diff
@@ -67,7 +70,21 @@ public long length()
...
     {
         // for CDC Cassandra writes COMPLETED in the final line of the CommitLog-7-*_cdc.idx index file.
-        return maxOffset() >= 67108818;
+        List<String> lines = readIdxFile();
+        return lines.size() >= 2 && lines.get(1).trim().equalsIgnoreCase("COMPLETED");
+    }
+
+    private List<String> readIdxFile()
+    {
+        Path idxPath = path.resolveSibling(name.replace(".log", "_cdc.idx"));
```
jmckenzie-dev, line 80: Recommend adding a brief comment here that each log file has a corresponding `_cdc.idx` file that goes along with it; ironically I was the one that wrote the patch that made that change and I couldn't remember off the top of my head the relationship between the two. :)

A follow up JIRA for C* proper or the sidecar to provide interfaces that give this functionality around CDC (get the file that goes with commitlog X, tell me if it's COMPLETED, etc) seems like a useful primitive abstraction for us to extract. Maybe that's what `LocalCommitLog.java` and `CommitLog.java` are intended to be in the analytics library here? The names aren't great IMO and the overlap with what's in the core C* database make that more confusing to me.

Since none of these classes have top level javadocs explaining their purpose nor their relationship to the root classes and concepts in the core C* database, it leaves a lot up to the interpretation of other maintainers that come to this code cold.

So for this patch: brief comment here, and I'll actually take the note for the follow up JIRA for the above if that resonates with you.

jmckenzie-dev (approved):
Good to merge w/that one little comment add.

cassandra-analytics-cdc/src/test/java/org/apache/cassandra/cdc/LocalCommitLog.java
jyothsnakonisa: Yes the CommitLog.java file is the class that provides that interface and it has documentation around index file, completed state and offset. But you are right, the naming of the class could have been better

PR #181, opened 2026-03-11
CASSANALYTICS-128: Add flag to allow bulk write to indexed tables

jmckenzie-dev:
Patch by Josh McKenzie; reviewed by TBD for CASSANALYTICS-128

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
@@ -127,6 +127,7 @@ public class BulkSparkConf implements Serializable
     public final int commitThreadsPerInstance;
     public final double importCoordinatorTimeoutMultiplier;
     public boolean quoteIdentifiers;
+    public final boolean skipSecondaryIndexCheck;
     protected final String keystorePassword;
     protected final String keystorePath;
     protected final String keystoreBase64Encoded;
@@ -207,6 +208,7 @@ public class BulkSparkConf implements Serializable
         this.ttl = MapUtils.getOrDefault(options, WriterOptions.TTL.name(), null);
         this.timestamp = MapUtils.getOrDefault(options, WriterOptions.TIMESTAMP.name(), null);
         this.quoteIdentifiers = MapUtils.getBoolean(options, WriterOptions.QUOTE_IDENTIFIERS.name(), false, "quote identifiers");
+        this.skipSecondaryIndexCheck = MapUtils.getBoolean(options, WriterOptions.SKIP_SECONDARY_INDEX_CHECK.name(), false, "skip secondary index check");
         int storageClientConcurrency = MapUtils.getInt(options, WriterOptions.STORAGE_CLIENT_CONCURRENCY.name(),
                                                        DEFAULT_STORAGE_CLIENT_CONCURRENCY, "storage client concurrency");
         long storageClientKeepAliveSeconds = MapUtils.getLong(options, WriterOptions.STORAGE_CLIENT_THREAD_KEEP_ALIVE_SECONDS.name(),
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
@@ -69,7 +69,8 @@ public class TableSchema
                        TTLOption ttlOption,
                        TimestampOption timestampOption,
                        String lowestCassandraVersion,
-                       boolean quoteIdentifiers)
+                       boolean quoteIdentifiers,
+                       boolean skipSecondaryIndexCheck)
     {
         this.writeMode = writeMode;
         this.ttlOption = ttlOption;
@@ -79,7 +80,21 @@ public class TableSchema
         this.quoteIdentifiers = quoteIdentifiers;
 
         validateDataFrameCompatibility(dfSchema, tableInfo);
-        validateNoSecondaryIndexes(tableInfo);
+        // If a table has indexes on it, some external process (application, DB, etc.) is responsible for rebuilding
+        // indexes on the table after the bulk write completes; cassandra does this as part of the SSTable import
+        // process today. 2i and SAI have different ergonomics here regarding if stale data is served during index build;
+        // ultimately we want the bulk writer to also write native SAI index files alongside sstables but until
+        // then, this is allowable and fine for users who Know What They're Doing.
+        if (!skipSecondaryIndexCheck)
+        {
+            validateNoSecondaryIndexes(tableInfo);
+        }
+        else if (tableInfo.hasSecondaryIndex())
+        {
+            LOGGER.warn("Bulk writing to tables with SecondaryIndexes will have an asynchronous index rebuild "
+                      + "take place automatically after writing. Reads against the index during this time "
+                      + "window will produce inconsistent or stale results until index rebuild is complete.");
+        }
         validateUserAddedColumns(lowestCassandraVersion, quoteIdentifiers, ttlOption, timestampOption);
 
         this.createStatement = getCreateStatement(tableInfo);
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
@@ -132,4 +132,12 @@ public enum WriterOptions implements WriterOption
      * - a failure otherwise
      */
     JOB_TIMEOUT_SECONDS,
+    /**
+     * Option to bypass the secondary index validation check during bulk write job setup.
+     * By default, bulk writes to tables with secondary indexes are rejected.
+     * Setting this option to {@code true} allows bulk writes to proceed on tables that have secondary indexes,
+     * with the understanding that the secondary indexes will NOT be updated by the bulk write and must be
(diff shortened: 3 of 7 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -127,6 +127,7 @@ public class BulkSparkConf implements Serializable
     public final int commitThreadsPerInstance;
     public final double importCoordinatorTimeoutMultiplier;
     public boolean quoteIdentifiers;
+    public boolean skipSecondaryIndexCheck;
```
sarankk:
```suggestion
    public final boolean skipSecondaryIndexCheck;
```

sarankk (commented):
Thanks Josh, Lgtm. 1 minor nit. Would it be useful to test writer with this flag in integration test too?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
```diff
@@ -132,4 +132,12 @@ public enum WriterOptions implements WriterOption
...
      */
     JOB_TIMEOUT_SECONDS,
+    /**
+     * Option to bypass the secondary index validation check during bulk write job setup.
+     * By default, bulk writes to tables with secondary indexes are rejected.
+     * Setting this option to {@code true} allows bulk writes to proceed on tables that have secondary indexes,
+     * with the understanding that the secondary indexes will NOT be updated by the bulk write and must be
+     * rebuilt separately after the job completes.
+     */
+    SKIP_SECONDARY_INDEX_CHECK,
```
yifan-c, line 142: How about defining it in the spark conf, instead of write options? 

The rationale is that the toggle is to for advanced use case and not directly related to the write behavior. There is another existing spark conf, `org.apache.cassandra.spark.bulkwriter.BulkSparkConf#SKIP_CLEAN` that skips cleaning up SSTable when job fails. 

Admittedly, there are some existing inconsistency of where to have conf and where to have write options in the project.

jmckenzie-dev: Would putting this in `BulkSparkConf` force it session-wide? i.e. a user would lose the ability to reason about and easily configure this setting on a per-table / per operation basis vs. the public exposure of it via `WriterOptions`?

My intuition right now is that this is something that probably should have been enabled by default all this time w/a configurable guardrail to turn it off, so while I'm sympathetic to the idea of taking a small step from "don't allow it" to "allow it but make it hard to use", the risk of this being easily accessible for users is that they'll bulk write to a table that will then have a long running index building operation happen in the background. Which, other than "load on node" and "application might read a partial index if you don't have automation that clearly delineates when a bulk insert and reindex finish from application accessing it", doesn't represent a structural or correctness risk to the data.

jmckenzie-dev:
> Would it be useful to test writer with this flag in integration test too?

Eh, I didn't because ultimately it's a validation error inside the `TableSchema` construction so basically, if we can get past that point then we should be good to go.

The rest of the integration path would effectively be testing whether bulk import on cassandra with a table that has legacy SecondaryIndexes on it works correctly. Which, to be completely honest, I'm not super confident of (dug around a bit and `ImportTest.java` doesn't seem to have coverage). BUT - that's a core C* test coverage problem and not an analytics / sidecar problem.

So maybe a follow up JIRA to add test coverage in C* proper for legacy 2i on imports would be reasonable?

jmckenzie-dev:
CI: https://app.circleci.com/pipelines/github/jmckenzie-dev/cassandra-analytics?branch=allow_write_2i

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -207,6 +208,7 @@ public BulkSparkConf(SparkConf conf, Map<String, String> options, @Nullable Logg
         this.ttl = MapUtils.getOrDefault(options, WriterOptions.TTL.name(), null);
         this.timestamp = MapUtils.getOrDefault(options, WriterOptions.TIMESTAMP.name(), null);
         this.quoteIdentifiers = MapUtils.getBoolean(options, WriterOptions.QUOTE_IDENTIFIERS.name(), false, "quote identifiers");
+        this.skipSecondaryIndexCheck = MapUtils.getBoolean(options, WriterOptions.SKIP_SECONDARY_INDEX_CHECK.name(), false, "skip secondary index check");
```
jyothsnakonisa, line 211: Should the default value be true?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
```diff
@@ -79,7 +80,15 @@ public TableSchema(StructType dfSchema,
         this.quoteIdentifiers = quoteIdentifiers;
 
         validateDataFrameCompatibility(dfSchema, tableInfo);
-        validateNoSecondaryIndexes(tableInfo);
+        // If a table has indexes on it, some external process (application, DB, etc.) is responsible for rebuilding
+        // indexes on the table after the bulk write completes; cassandra does this as part of the SSTable import
+        // process today. 2i and SAI have different ergonomics here regarding if stale data is served during index build;
+        // ultimately we want the bulk writer to also write native SAI index files alongside sstables but until
+        // then, this is allowable and fine for users who Know What They're Doing.
+        if (!skipSecondaryIndexCheck)
```
jyothsnakonisa, line 88: Can you please add a test if possible that check this behavior, basically the if block

jyothsnakonisa (approved):
looks good to me

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
```diff
@@ -132,4 +132,12 @@ public enum WriterOptions implements WriterOption
      * - a failure otherwise
      */
     JOB_TIMEOUT_SECONDS,
+    /**
+     * Option to bypass the secondary index validation check during bulk write job setup.
+     * By default, bulk writes to tables with secondary indexes are rejected.
+     * Setting this option to {@code true} allows bulk writes to proceed on tables that have secondary indexes,
+     * with the understanding that the secondary indexes will NOT be updated by the bulk write and must be
+     * rebuilt separately after the job completes.
```
skoppu22, line 140: Does C* detect and rebuild them, or does the user need to run a command to trigger the rebuild?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
skoppu22: I believe default value shouldn't be true, because user need to understand that rebuild need to happen after the job.
@jmckenzie-dev can we print a note/warning to the user mentioning rebuild needs to happen after the job, so they are aware of penalty of using this?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
jmckenzie-dev: C* automatically rebuilds them on import. Part of the SSTableImporter logic paths.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
jmckenzie-dev: Is this not explicitly tested with the following?
```
    @ParameterizedTest
    @MethodSource("org.apache.cassandra.bridge.VersionRunner#supportedVersions")
    public void testSecondaryIndexAllowedWithSkipCheck(String cassandraVersion)
    {
        TableSchema schema = getValidSchemaBuilder(cassandraVersion)
                .withHasSecondaryIndex()
                .withSkipSecondaryIndexCheck()
                .build();
        assertThat(schema).isNotNull();
    }
```
As `TableSchemaTestCommon.MockTableSchemaBuilder#build` calls the `TableSchema` constructor with 2i enabled + the flag to allow, and `testSecondaryIndexIsUnsupprted` tests the other case.

Or am I misunderstanding the request?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
jmckenzie-dev: Good call; added:
```
        if (!skipSecondaryIndexCheck)
        {
            validateNoSecondaryIndexes(tableInfo);
        }
        else if (tableInfo.hasSecondaryIndex())
        {
            LOGGER.warn("Bulk writing to tables with SecondaryIndexes will have an asynchronous index rebuild "
                      + "take place automatically after writing. Reads against the index during this time "
                      + "window will produce inconsistent or stale results until index rebuild is complete.");
        }
```

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/WriterOptions.java
yifan-c: Josh explained to me offline too. I think it makes sense to have it as part of writer option. Especially in the scenario when the spark job has multiple write sessions, writer option allows finer control to toggle the flag on specific tables / write sessions.

skoppu22:
LGTM

PR #182, opened 2026-03-12
CASSANALYTICS-129 : Fix ReadStatusTracker to distinguish clean completion from error termination in BufferingCommitLogReader

jyothsnakonisa:


cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/db/commitlog/BufferingCommitLogReader.java
```diff
@@ -316,6 +316,14 @@ private void readCommitLogSegment() throws IOException
                     break;
                 }
             }
+
+            // If the segment is flagged complete in the index but no end-of-segment marker was
```
jmckenzie-dev, line 320: So my understanding of this is that we have both:
a. a path that can terminate gracefully with the `LEGACY_END_OF_SEGMENT_MARKER`, and
b. a path that terminates gracefully when our idx file says we're completed and we trust it

We have an issue where above in the code we have if we're < 4 bytes from the end of a segment on a legacy CommitLog - this isn't an error case at all but if we find it, we flag `statusTracker` to error:
```
                // We rely on reading serialized size == 0 (LEGACY_END_OF_SEGMENT_MARKER) to identify the end
                // of a segment, which happens naturally due to the 0 padding of the empty segment on creation.
                // However, it's possible with 2.1 era commitlogs that the last mutation ended less than 4 bytes
                // from the end of the file, which means that we'll be unable to read an a full int and instead
                // read an EOF here
                if (end - reader.getFilePointer() < 4)
                {
                    logger.trace("Not enough bytes left for another mutation in this CommitLog section, continuing");
                    statusTracker.requestTermination();
                    return;
                }
```

So in this case, we'll `requestTermination` on the statusTracker which sets:
```
        public void requestTermination()
        {
            error = true;
        }
```
That'll then run afoul of the logic you have here, since the `statusTracker.shouldContinue()` will return false and thus we not set position to log.maxOffset. Overloading the `statusTracker.requestTermination` to mean both "things are fine" and "things are failed" is making this incorrect I think.

If we augment `ReadStatusTracker` to have both a `markAsError()` method that sets `error = true;` and a separate `markCleanCompletion()` that sets another boolean, we'll cleanly work around this situation and can check the combination of:
`if (log.completed() && statusTracker.noErrors() && statusTracker.isCompleted())` (or whatever better names you'd pick), then I _think_ we'd be solid here and be able to handle both the legacy CDC path and the new index based.

jmckenzie-dev (approved):
LGTM with the changes - thanks!

PR #183, opened 2026-03-17
CASSANALYTICS-122 : Use long for absolute times and support C* 5.0 extended localDeletionTime

skoppu22:
- Integers cannot hold absolute times beyond year 2038
- C* 5.0 fixed this, and supporting extended localDeletionTime. C* analytics need to support this especially for CDC events.
- Changed `deletedAt` in Avro schema to long 

CI run here : https://app.circleci.com/pipelines/github/skoppu22/cassandra-analytics/75/workflows/f6d5f03b-8373-4549-8daf-5f776f78ff96


.circleci/config.yml
jyothsnakonisa, line 1: Do you wanna revert these changes or are these intentional?

cassandra-four-zero-bridge/src/test/java/org/apache/cassandra/spark/data/CqlTypeY2038Test.java
```diff
@@ -0,0 +1,95 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/**
+ * Tests verifying Y2038 boundary behavior for {@link CqlType#tombstone} and {@link CqlType#expiring}
+ * in Cassandra 4.0.
+ */
+class CqlTypeY2038Test
+{
+    private static final CassandraBridgeImplementation BRIDGE = new CassandraBridgeImplementation();
```
jyothsnakonisa: Unused CassandraBridgeImplementation object

cassandra-analytics-cdc/src/test/java/org/apache/cassandra/cdc/avro/TestSchemaStore.java
jyothsnakonisa, line 1: This class looks very similar to "LocalTableSchemaStore" can you check if you can use that instead of adding this new class. You might make small modifications in LocalTableSchemaStore if required

jyothsnakonisa (approved):
Thanks for working on this shailaja, looks good to me left few minor comments

.circleci/config.yml
skoppu22: Intentional, CI pipelines failing otherwise.

cassandra-analytics-cdc/src/test/java/org/apache/cassandra/cdc/avro/TestSchemaStore.java
skoppu22: LocalTableSchemaStore and new Avro transformer tests  are in different modules. LocalTableSchemaStore loads avsc files from class path which doesn't exists in this other modules. Hence using TestSchemaStore, which doesn't load any schema files, instead allows registering of schemas.

cassandra-four-zero-bridge/src/test/java/org/apache/cassandra/spark/data/CqlTypeY2038Test.java
skoppu22: Removing this

.circleci/config.yml
```diff
@@ -64,10 +64,13 @@ commands:
...
             INTEGRATION_MAX_PARALLEL_FORKS: 1
             INTEGRATION_MAX_HEAP_SIZE: "1500M"
+            CORE_MAX_PARALLEL_FORKS: 2
+            CORE_TEST_MAX_HEAP_SIZE: "2048m"
             CASSANDRA_USE_JDK11: <<parameters.use_jdk11>>
           command: |
+            export GRADLE_OPTS="-Xmx2g -Dorg.gradle.jvmargs=-Xmx2g"
             # Run compile/unit tests, skipping integration tests
-            ./gradlew --stacktrace clean assemble check -x cassandra-analytics-integration-tests:test -Dcassandra.analytics.bridges.sstable_format=<<parameters.sstable_format>>
+            ./gradlew --no-daemon --max-workers=2 --stacktrace clean assemble check -x cassandra-analytics-integration-tests:test -Dcassandra.analytics.bridges.sstable_format=<<parameters.sstable_format>>
```
yifan-c, line 73: Instead of making CI pipeline changes, can you try to tag the relevant tests with `@Tag("Sequential")`?


`--no-daemon` flag makes sense.

cassandra-analytics-cdc-codec/src/main/java/org/apache/cassandra/cdc/avro/CdcEventUtils.java
```diff
@@ -281,14 +281,14 @@ public static GenericData.Record getTTLAvro(CdcEvent event, Schema ttlSchema)
...
-    public static Map<String, Integer> getTTL(CdcEvent event)
+    public static Map<String, Long> getTTL(CdcEvent event)
     {
         CdcEvent.TimeToLive ttl = event.getTtl();
         if (ttl == null)
         {
             return null;
         }
-        return mapOf(AvroConstants.TTL_KEY, ttl.ttlInSec, AvroConstants.DELETED_AT_KEY, ttl.expirationTimeInSec);
+        return mapOf(AvroConstants.TTL_KEY, (long) ttl.ttlInSec, AvroConstants.DELETED_AT_KEY, ttl.expirationTimeInSec);
```
yifan-c, line 291: Why converting TTL to `long`, meanwhile the schema has it as `int`? 
The change is probably unnecessary.

cassandra-analytics-cdc/src/test/java/org/apache/cassandra/cdc/AvroGenericRecordTransformerTest.java
```diff
@@ -0,0 +1,502 @@
...
+        String namespace = cqlTable.keyspace() + "." + cqlTable.table();
+        schemaStore.registerSchema(namespace, avroSchema);
+
+        Function<KeyspaceTypeKey, CqlField.CqlType> typeLookup = key -> bridge.parseType(key.type);
+        return new AvroGenericRecordTransformer(schemaStore, typeLookup, "");
+    }
+
+    @ParameterizedTest
+    @MethodSource("org.apache.cassandra.cdc.test.TestVersionSupplier#testVersions")
+    public void testBasicInsertAvroEncoding(CassandraVersion version)
```
yifan-c, line 89: The file adds a few irrelevant test cases. Improving test coverage is in general a good thing. So I am fine with those test cases in this file.

cassandra-analytics-core/build.gradle
```diff
@@ -217,8 +217,9 @@ tasks.register('testSequential', Test) {
 test {
     systemProperty "cassandra.analytics.bridges.sstable_format", System.getProperty("cassandra.analytics.bridges.sstable_format", "big")
     minHeapSize = '1024m'
-    maxHeapSize = '3072m'
-    maxParallelForks = Math.max(Runtime.runtime.availableProcessors() * 2, 8)
+    maxHeapSize = System.getenv('CORE_TEST_MAX_HEAP_SIZE') ?: '3072m'
+    maxParallelForks = System.getenv('CORE_MAX_PARALLEL_FORKS')?.toInteger()
+                       ?: Math.max(Runtime.runtime.availableProcessors() * 2, 8)
```
yifan-c, line 222: Please try out the Sequential test tag first.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteComplexTypeTtlTest.java
```diff
@@ -0,0 +1,201 @@
...
+                                  .collect(Collectors.toList());
+        Dataset<Row> df = spark.createDataFrame(rows, schema);
+
+        bulkWriterDataFrameWriter(df, UDT_TTL_TABLE).option(WriterOptions.TTL.name(), TTLOption.constant(80))
+                                                    .save();
+
+        Dataset<Row> preExpiry = bulkReaderDataFrame(UDT_TTL_TABLE).load();
+        assertThat(preExpiry.collectAsList()).hasSize(ROW_COUNT);
+
+        Uninterruptibles.sleepUninterruptibly(80, TimeUnit.SECONDS);
```
yifan-c: Please find a different way and avoid sleep the test for 80 seconds. We cannot have the CI sleep for 80 seconds. 

If there is no other way, please delete all the tests in this file. They are irrelevant to the patch anyway.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/db/DbUtils.java
```diff
@@ -33,23 +33,27 @@ private DbUtils()
         throw new IllegalStateException(getClass() + " is static utility class and shall not be instantiated");
     }
 
-    public static DeletionTime deletionTime(long markedForDeleteAt, int localDeletionTime)
+    // C* 4.0 DeletionTime constructor requires int for localDeletionTime; checked cast will throw after Y2038
+    public static DeletionTime deletionTime(long markedForDeleteAt, long localDeletionTime)
```
yifan-c: The comment looks scary, but the methods (this and the others in this file) are indeed for test only. Let's add the annotation `@VisibleForTesting` to all methods. The test should ensure passing the valid value.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/AbstractStreamScanner.java
```diff
@@ -443,7 +444,8 @@ public void consume()
                     Cell<?> cell = cells.next();
                     // Re: isLive vs. isTombstone - isLive considers TTL so that if a cell is expiring soon,
                     // it is handled as tombstone
-                    if (cell.isLive(timeProvider.referenceEpochInSeconds()))
+                    // C* 4.0 Cell.isLive requires int for nowInSec; checked cast will throw after Y2038
+                    if (cell.isLive(Ints.checkedCast(timeProvider.referenceEpochInSeconds())))
```
yifan-c: Since the `referenceEpochInSeconds` value is static, can you check it just once on time provider initialization? Instead of cast and check in every loop here. The code here is the hot path.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/CompactionStreamScanner.java
```diff
@@ -117,15 +118,16 @@ protected void handleCellTombstoneInComplex(BigInteger token, Cell<?> cell)
...
         Keyspace keyspace = Keyspace.openWithoutSSTables(metadata.keyspace);
         ColumnFamilyStore cfStore = keyspace.getColumnFamilyStore(metadata.name);
         controller = new PurgingCompactionController(cfStore, CompactionParams.TombstoneOption.NONE);
         List<ISSTableScanner> scannerList = toCompact.stream()
                                                      .map(Scannable::scanner)
                                                      .collect(Collectors.toList());
         scanners = new AbstractCompactionStrategy.ScannerList(scannerList);
-        ci = new CompactionIterator(OperationType.COMPACTION, scanners.scanners, controller, nowInSec, taskId);
+        // C* 4.0 CompactionIterator requires int for nowInSec; checked cast will throw after Y2038
+        ci = new CompactionIterator(OperationType.COMPACTION, scanners.scanners, controller, Ints.checkedCast(nowInSec), taskId);
```
yifan-c, line 130: again, the checkedCast is probably unnecessary.

cassandra-four-zero-bridge/src/test/java/org/apache/cassandra/spark/data/CqlTypeY2038Test.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.cassandra.spark.reader.SchemaBuilder;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/**
+ * Tests verifying Y2038 boundary behavior for {@link CqlType#tombstone} and {@link CqlType#expiring}
+ * in Cassandra 4.0.
+ */
+class CqlTypeY2038Test
```
yifan-c: I do not think this test is valuable. Analytics/CDC does not _set_ TTL, it only _reads_ the TTL values from Cassandra. The TTL values are guaranteed to be valid already; otherwise the cql write fails already. Please delete this file.

cassandra-four-zero-types/src/main/java/org/apache/cassandra/spark/data/complex/AbstractCqlList.java
```diff
@@ -105,15 +105,15 @@ public void addCell(final org.apache.cassandra.db.rows.Row.Builder rowBuilder,
...
     {
         for (Object o : (List<?>) value)
         {
             if (ttl != NO_TTL)
             {
-                rowBuilder.addCell(BufferCell.expiring(cd, timestamp, ttl, now, type().serialize(o),
-                                                       CellPath.create(ByteBuffer.wrap(UUIDGen.getTimeUUIDBytes()))));
+                rowBuilder.addCell(CqlType.expiring(cd, timestamp, ttl, now, type().serialize(o),
+                                                    CellPath.create(ByteBuffer.wrap(UUIDGen.getTimeUUIDBytes()))));
             }
```
yifan-c, line 117: I think this change can be reverted once you drop `CqlTypeY2038Test` and revert the change in `CqlType`

cassandra-four-zero-types/src/main/java/org/apache/cassandra/spark/data/complex/CqlMap.java
```diff
@@ -120,15 +120,15 @@ public void addCell(final org.apache.cassandra.db.rows.Row.Builder rowBuilder,
...
                         Object value)
     {
         for (Map.Entry<?, ?> entry : ((Map<?, ?>) value).entrySet())
         {
             if (ttl != NO_TTL)
             {
-                rowBuilder.addCell(BufferCell.expiring(cd, timestamp, ttl, now, valueType().serialize(entry.getValue()),
-                                                       CellPath.create(keyType().serialize(entry.getKey()))));
+                rowBuilder.addCell(CqlType.expiring(cd, timestamp, ttl, now, valueType().serialize(entry.getValue()),
+                                                    CellPath.create(keyType().serialize(entry.getKey()))));
```
yifan-c, line 131: can be reverted too

cassandra-four-zero-types/src/main/java/org/apache/cassandra/spark/data/complex/CqlSet.java
```diff
@@ -104,15 +104,15 @@ public void addCell(final org.apache.cassandra.db.rows.Row.Builder rowBuilder,
...
                         Object value)
     {
         for (Object o : (Set<?>) value)
         {
             if (ttl != NO_TTL)
             {
-                rowBuilder.addCell(BufferCell.expiring(cd, timestamp, ttl, now, ByteBufferUtil.EMPTY_BYTE_BUFFER,
-                                                       CellPath.create(type().serialize(o))));
+                rowBuilder.addCell(CqlType.expiring(cd, timestamp, ttl, now, ByteBufferUtil.EMPTY_BYTE_BUFFER,
+                                                    CellPath.create(type().serialize(o))));
```
yifan-c, line 115: can revert

CHANGES.txt
```diff
@@ -1,5 +1,6 @@
 0.4.0
 -----
+ * Fix year 2038 problem using long for absolute times and support C* 5.0 extended localDeletionTime
```
yifan-c: The year 2038 problem sounds cool, but I feel it shades the actual change in this patch. 

I would revise the change log entry to "Support extended deletion time in CDC for Cassandra 5.0" and update the JIRA title too.

yifan-c (changes requested):

.circleci/config.yml
skoppu22: - We don't know which tests causing this to tag them. Gradle getting killed while executing pipelines starting with spark3-* which run unit tests (gradlew check). Pipelines starting with int-* are running fine. This problem exists in trunk, not created by this PR. Running CI with --info didn't give any clues as it was bloating logs and CI killing gradle
- And we will have the same problem of finding the culprit everytime CI is broken

cassandra-analytics-cdc-codec/src/main/java/org/apache/cassandra/cdc/avro/CdcEventUtils.java
skoppu22: Here we are creating map, one entry as (TTL_KEY, TTL value as int), second entry (DELETED_AT_KEY, expirationTimeInSec value as long), so we have to take higher of int and long. Otherwise we need to return Map<String, Number> so Number can represent both int and long.

cassandra-analytics-cdc/src/test/java/org/apache/cassandra/cdc/AvroGenericRecordTransformerTest.java
skoppu22: We have updated two avro schema files, and there are no existing tests going through avro transformers. Hence had to add them. testMaxSupportedTtlAvroEncoding test below in this file verifies deletedAt is long and supporting max TTL as expected on both C* 4.x and 5.0

cassandra-analytics-core/build.gradle
skoppu22: As I mentioned above, pipelines broken on trunk, we don't know which test caused that, gradlew check is getting killed on CI

cassandra-four-zero-types/src/main/java/org/apache/cassandra/spark/data/complex/CqlMap.java
skoppu22: CqlMap is implemented in C* 4.0 bridge and not extended (i.e, reused) in C* 5.0 bridge. While 4.0 bridge expects int for nowInSec, 5.0 bridge expects long for the same. I have modified this to use CqlType's method which is implemented separately for both bridges and uses appropriate type in corresponding bridges. Without this change, we will have to implement CqlMap again in 5.0 bridge and override this function to pass long value.

cassandra-four-zero-types/src/main/java/org/apache/cassandra/spark/data/complex/AbstractCqlList.java
skoppu22: BufferCell.expiring in 4.0 bridge expects int for nowInSec, 5.0 bridge expects long. CqlList implemented in C* 5.0 bridge doesn't override this method to pass long. Hence I have modified this to use CqlType's method which is implemented separately for both bridges and uses appropriate type in corresponding bridges. Without this change, we will have to override this method again in 5.0 bridge to pass long value.

cassandra-four-zero-types/src/main/java/org/apache/cassandra/spark/data/complex/CqlSet.java
skoppu22: Same as explained above for CqlMap

CHANGES.txt
skoppu22: OK

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/CompactionStreamScanner.java
skoppu22: checkedCast throws exception when a long cannot fit in int, and we need to know when that happens. Simple `(int) nowInSec` conversion will silently truncate the value. It will happen after Jan 19, 2038 if we don't remove C* 4.x support by then.

cassandra-four-zero-bridge/src/test/java/org/apache/cassandra/spark/data/CqlTypeY2038Test.java
skoppu22: We are just ensuring checkedCast code we added indeed complaining when time indeed exceeds the range where int cannot accommodate. But the test can be removed as we are sure checkedCast does the job.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteComplexTypeTtlTest.java
skoppu22: These tests are relevant. CqlSet, CqlMap etc are implemented in C* 4.0 bridge and extended (i.e, reused) in C* 5.0 bridge. While 4.0 bridge expects int, 5.0 bridge expects long. I have modified these files below to use CqlType's method which is implemented separately for both bridges. And there are no existing tests for collections TTL going though this modified code path.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/db/DbUtils.java
skoppu22: Why are these methods test only? I see DbUtils methods are being called from AbstractCdcBridgeImplementation.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/AbstractStreamScanner.java
skoppu22: OK

.circleci/config.yml
yifan-c: > This problem exists in trunk, not created by this PR

All commits are made only after CI is green. I do not think the problem pre-exists in trunk. Likely relevant to the new tests. You will have to dig into the details to determine which test should be tagged with `Sequential`. The general idea is to run the resource consuming tests sequentially. @lukasz-antoniak, might have some tips. 

> we will have the same problem of finding the culprit everytime CI is broken

I share the same concern. This is why I am conservative on modifying the CI configuration, given that the CI issue is probably relevant to the current patch.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteComplexTypeTtlTest.java
yifan-c: Just to be clear, I was not arguing the test is not valuable :p

The point is that we should not sleep for 80 seconds. Can you instead try to write data using a timestamp in the past?

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/db/DbUtils.java
yifan-c: Because the methods are used to create mutations for testing purpose only.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/CompactionStreamScanner.java
yifan-c: Right. My point is that the method is **unnecessary**, not **incorrect**. CDC consumes live valid data from Cassandra. So it is already guaranteed by the upstream system the data is valid.

cassandra-four-zero-bridge/src/test/java/org/apache/cassandra/spark/data/CqlTypeY2038Test.java
yifan-c: The reason for removal is mentioned in https://github.com/apache/cassandra-analytics/pull/183#discussion_r2958030914, and similarly in this comment too, https://github.com/apache/cassandra-analytics/pull/183#discussion_r2975959891

Please remove the test.

.circleci/config.yml
skoppu22: - I see same failures on trunk as well without any changes https://app.circleci.com/pipelines/github/skoppu22/cassandra-analytics?branch=trunk
I am seeing this consistency on trunk, not sure is Circle CI backend for Europe region has different setup by any chance. 

- Finding the test which went in trunk causing this will be a separate task and takes time. We won't be able to merge in this PR until we do that, as my trunk pipelines also never turning green without this fix

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/db/DbUtils.java
skoppu22: I see caller's caller is always test code. Marked them as @VisibleForTesting in both 4.0 and 5.0 bridge

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/CompactionStreamScanner.java
skoppu22: I understand if analytics 4.0 bridge is used with C* 4.0, then the value is always guaranteed to fit in int. But with sstable version dependent bridge determination feature (we have a open PR for this), there is a possibility of using 4.0 bridge with C* 5.0 if the cluster is running in 4_0 compatibility mode and doesn't have 5.0 sstables. Having checkedCast gives us freedom from worrying about such cases, not only now, later as well with upcoming C* releases. checkedCast has just one more check on top of a simple conversion, and CPUs can predict branching well, so there is no considerable performance impact of using checkedCast. Hence I like to keep checkedCast to feel comfortable with such special cases.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteComplexTypeTtlTest.java
skoppu22: I have merged all 4 tests into one and reduced the sleep to 5 seconds. That way entire test file sleeps for max 5 seconds.

cassandra-four-zero-bridge/src/test/java/org/apache/cassandra/spark/data/CqlTypeY2038Test.java
skoppu22: This one was removed in the previous commit

CHANGES.txt
```diff
@@ -1,6 +1,6 @@
 0.4.0
 -----
- * Setup CI Pipeline with GitHub Actions (CASSANALYTICS-106)
+ * Support extended deletion time in CDC for Cassandra 5.0
```
yifan-c, line 3: 👍 to the removal. GitHub Actions is not user facing

yifan-c (approved):

skoppu22:
Latest CI run https://app.circleci.com/pipelines/github/skoppu22/cassandra-analytics/82/workflows/cd78a753-181a-46e0-8ee1-c8889976ab07

.circleci/config.yml
jmckenzie-dev: I'm with @skoppu22 on this one; we should band-aid here to get things working again and open another JIRA to follow up and fix it the right way. Blocking all work on root causing a regression from another commit should either be solved by:
a. git revert on that commit, or
b. a band-aid workaround w/a follow up JIRA to unblock work

In my humble opinion. :)

yifan-c: I am good with having follow up work to address the root cause.

My reason was that all the prior commits were only merged with green CI. This is the policy we have been following. Admittedly, there could be reruns and some flakiness. If the build starts to fail consistently, it could mean something specific to this patch.

jmckenzie-dev: Did the 5.0 support w/parameterized testing go in w/out green CI?

skoppu22: @yifan-c definitely something went in without green CI or something changed in Circle CI side. As I shared link above, trunk pipelines consistently failing with the same problem.

skoppu22: Something went in between March 2nd and March 13 broke this. My trunk pipeline didn't had this problem on March 2nd, then consistently failing from March 13th.

PR #187, opened 2026-03-26
CASSANALYTICS-130 Support per-instance sidecar port resolution in CDC client

Klose6:
Currently when the sidecar tries to call another sidecar instance it assumes another sidecar will use the same port as itself, this change make the port as a input parameter so that the downstream caller can have the option to pass in the port if other sidecar is using a different port.

### DIFF
--- a/cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcBuilder.java
+++ b/cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcBuilder.java
@@ -19,8 +19,6 @@
 
 package org.apache.cassandra.cdc.sidecar;
 
-import java.util.function.Function;
-
 import com.google.common.base.Preconditions;
 
 import org.apache.cassandra.cdc.CdcBuilder;
@@ -63,12 +61,6 @@ public class SidecarCdcBuilder extends CdcBuilder
     public SidecarCdcBuilder withClusterConfigProvider(ClusterConfigProvider clusterConfigProvider)
     {
         this.clusterConfigProvider = clusterConfigProvider;
-        return this;
-    }
-
-    public SidecarCdcBuilder withPortResolver(@NotNull Function<String, Integer> portResolver)
-    {
-        this.sidecarCdcClient = sidecarCdcClient.withPortResolver(portResolver);
         return this;
     }
 
--- a/cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcBuilder.java
+++ b/cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcBuilder.java
@@ -19,10 +19,6 @@
 
 package org.apache.cassandra.cdc.sidecar;
 
-import java.io.IOException;
-import java.util.Map;
-import java.util.function.Function;
-
 import com.google.common.base.Preconditions;
 
 import org.apache.cassandra.cdc.CdcBuilder;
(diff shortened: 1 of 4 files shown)


cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcBuilder.java
```diff
@@ -43,6 +44,8 @@ public class SidecarCdcBuilder extends CdcBuilder
 {
     protected ClusterConfigProvider clusterConfigProvider;
     protected SidecarCdcClient sidecarCdcClient;
+    @NotNull
+    protected Function<String, Integer> portResolver;
```
jberragan: Could you keep a default implementation that returns `(ignored) ->  config.effectivePort()` so we keep the preexisting behavior. Then add a builder method e.g. `withPortResolver` if users wish to override with their own implementation.

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcBuilder.java
```diff
@@ -85,17 +89,38 @@ public class SidecarCdcBuilder extends CdcBuilder
...
+        this.sidecarCdcClient = new SidecarCdcClient(clientConfig, sidecarClient, cdcStats, portResolver);
         withCdcOptions(cdcOptions);
         withTokenRangeSupplier(tokenRangeSupplier);
     }
 
+    /**
+     * Builds a port resolver function from the {@link CdcSidecarInstancesProvider}. The resolver looks up
+     * the port for a given hostname from the provider, falling back to the configured effective port.
+     */
+    static Function<String, Integer> buildPortResolver(@NotNull CdcSidecarInstancesProvider provider,
```
jberragan: This is a bit `O(n)` can we change the interface to be `(CassandraInstance) -> Integer`, that way we can just return:

```
    (instance) -> instance.port();
```

jberragan (commented):
Thanks for the patch, I made a couple of suggestions.

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcBuilder.java
Klose6: sure, I added the preexisting behavior in the constructor and another new constructor for the new portResolver parameter, also I just found this latest [commit](https://github.com/apache/cassandra-analytics/commit/dfface1be7a0d70e745b6e0732e681bde237508c#diff-1e824b92860d2a734894ee972f24297f4f30cf450359333224a947b48c49c7ae) so I add the withPortResolver() in  SidecarCdcClient and SidecarCdcBuilder to support the port resolving. Thanks!

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcBuilder.java
Klose6: In this latest [commit](https://github.com/apache/cassandra-analytics/commit/dfface1be7a0d70e745b6e0732e681bde237508c#diff-1e824b92860d2a734894ee972f24297f4f30cf450359333224a947b48c49c7ae) it accepts the sidecarCdcClient in the SidecarCdc.java so I removed this and in cassandra-sidecar we can inject the sidecar port mapping into the sidecarCdcClient with its added withPortResolver() when calls the SidecarCdc.builder().

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcBuilder.java
Klose6: @jberragan I double check the latest commit and think we don't need the changes on SidecarCdcBuilder now because the SidecarCdcClient is a input parameter to SidecarCdc now, remove the withPortResolver() in SidecarCdcBuilder and only the changes in sidecarCdcClient, thanks!

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcClient.java
```diff
@@ -61,11 +63,23 @@ public class SidecarCdcClient implements AutoCloseable
...
+    {
+        this(clientConfig, instancesProvider, secretsProvider, cdcStats,
+             ignored -> clientConfig.effectivePort());
+    }
+
+    public SidecarCdcClient(ClientConfig clientConfig,
+                            CdcSidecarInstancesProvider instancesProvider,
+                            SecretsProvider secretsProvider,
+                            ICdcStats cdcStats,
+                            @NotNull Function<String, Integer> portResolver) throws IOException
```
jberragan: Is there a reason we can't use `CassandraInstance` here?

i.e.
```suggestion
                            @NotNull Function<CassandraInstance, Integer> portResolver) throws IOException
```

Klose6: sure, initially I think using the Function<String, Integer> is simpler and avoid the extra dep on cassandra-analytics-common which the CassandraInstance is located, but use CassandraInstance will make it consistent with the current toSidecarInstance(), updated.

jberragan (approved):
+1, thanks for the patch.

yifan-c:
This patch was committed too early. 

According to [Cassandra Project Governance](https://cwiki.apache.org/confluence/display/CASSANDRA/Cassandra+Project+Governance), there are 2 approvals required before committing. 

> Code modifications require two +1 committer votes (can be author + reviewer)

I will take a second review, and if changes required, I will revert the commit. 

@Klose6, please share CircleCI result link as well. It should have a green run before committing.

PR #188, opened 2026-03-30
CASSANALYTICS-142 : SidecarCdcClient should be passed as a constructor parameter for SidecarCdc to avoid thread/resource leaks

jyothsnakonisa:


cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcBuilder.java
```diff
@@ -61,21 +56,11 @@ public class SidecarCdcBuilder extends CdcBuilder
...
-                                                                                .collect(Collectors.toList())),
-                     clientConfig.toGenericSidecarConfig(), secretsProvider),
-        cdcStats
-        );
+        super(jobId, partitionId, eventConsumer, schemaSupplier);
+        this.clusterConfigProvider = clusterConfigProvider;
+        this.sidecarCdcClient = new SidecarCdcClient(clientConfig, sidecarInstancesProvider, secretsProvider, cdcStats);
+        withCdcOptions(cdcOptions);
+        withTokenRangeSupplier(tokenRangeSupplier);
     }
```
yifan-c: You should be able to reuse the other constructor.

```java
        this(jobId, partitionId, cdcOptions, clusterConfigProvider, eventConsumer, schemaSupplier, tokenRangeSupplier, 
             new SidecarCdcClient(clientConfig, sidecarInstancesProvider, secretsProvider, cdcStats), 
             cdcStats);
```

However, the real question is: why do we need the constructor that creates `SidecarCdcClient`? We want to share the global `SidecarCdcClient` instance.

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcClient.java
```diff
@@ -52,21 +56,42 @@
...
         this.config = config;
         this.sidecarClient = sidecarClient;
         this.stats = stats;
     }
 
+    @Override
+    public void close() throws Exception
+    {
+        sidecarClient.close();
+    }
```
yifan-c, line 103: Please add a detailed comment to explain who closes the shared client and when it is closed.

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdc.java
```diff
@@ -75,6 +75,27 @@ public static SidecarCdcBuilder builder(@NotNull String jobId,
                                      cdcStats);
     }
```
yifan-c: Do we still need this build? Please remove it if it is no longer needed. One reason for removal is that the builder path creates a new sidecar cdc client.

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdc.java
```diff
@@ -75,6 +75,27 @@ public static SidecarCdcBuilder builder(@NotNull String jobId,
...
 
+    public static SidecarCdcBuilder builder(@NotNull String jobId,
```
yifan-c, line 72: Please add the java doc of the public method. More importantly, please document the expectation on who should manage the lifecycle of `SidecarCdcClient` and make it clear that Cdc/SidecarCdc does not close the client.

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdcBuilder.java
jyothsnakonisa: We want the `SidecarCdcClient` object to be singleton created in the Sidecar process and it should be shared across all the `SidecarCdc` instances which are CDC iterators.

I have added this new constructor to take `sidecarCdcClient` object as parameter so that we can have singleton object and can pass it to all iterators to avoid creating one `sidecarCdcClient` instance per iterator.

If we have the `SidecarCdcClient` creation inside the constructor, then on token range change when new iterator is created a new `SidecarCdcClient` is created, which we want to avoid.

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdc.java
jyothsnakonisa: good catch, this is no longer needed and we should remove it as the builder path creates new SidecarCdcClient. 

I also think, we should remove the constructor which creates new SidecarCdcClient in the constructor.

yifan-c (approved):

PR #189, opened 2026-03-31
CASSANALYTICS-143 Fix various leaks in CDC caching and bridge implementation

jmckenzie-dev:
Patch by Josh McKenzie; reviewed by Shailaja Koppu for CASSANALYTICS-143

This patch makes a lot of the lifecycle implementation in tests explicit (temp fails, cleanup of resources, etc) that previously relied on the CI env Just Working. Some of it is about tidying up and making explicit our position on resource management w/regards to test class lifecycle, and some of it is about leaks preventing GC (schema graph, etc).

### DIFF
--- a/cassandra-analytics-cdc/src/main/java/org/apache/cassandra/bridge/CdcBridgeFactory.java
+++ b/cassandra-analytics-cdc/src/main/java/org/apache/cassandra/bridge/CdcBridgeFactory.java
@@ -23,9 +23,12 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import com.google.common.annotations.VisibleForTesting;
 
+import com.google.common.base.Preconditions;
+import org.apache.cassandra.cdc.TypeCache;
 import org.apache.cassandra.cdc.avro.CqlToAvroSchemaConverter;
 import org.apache.cassandra.spark.utils.Throwing;
 import org.jetbrains.annotations.NotNull;
@@ -33,9 +36,11 @@ import org.jetbrains.annotations.Nullable;
 
 public final class CdcBridgeFactory extends BaseCassandraBridgeFactory
 {
+    private static final AtomicBoolean SHUTDOWN_HOOK_REGISTERED = new AtomicBoolean(false);
+
     // maps Cassandra version-specific jar name (e.g. 'four-zero') to matching CassandraBridge and SparkSqlTypeConverter
     private static final Map<String, VersionSpecificBridge> CASSANDRA_BRIDGES =
-    new ConcurrentHashMap<>(CassandraVersion.values().length);
+        new ConcurrentHashMap<>(CassandraVersion.values().length);
 
     public static class VersionSpecificBridge
     {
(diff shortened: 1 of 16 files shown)


cassandra-analytics-cdc/src/main/java/org/apache/cassandra/bridge/CdcBridgeFactory.java
```diff
@@ -165,6 +168,51 @@ private static VersionSpecificBridge create(@NotNull String label)
...
+                }
+            }, CdcBridgeFactory.class.getSimpleName() + "-shutdown"));
+        }
+    }
+
+    /**
+     * Closes all cached bridge classloaders and clears the cache. Call during application shutdown
+     * to release classloader resources and delete temp JAR files.
+     *
+     * Do not use this method outside testing; rely on the shutdown hook logic to clean things up in prod.
```
skoppu22, line 208: Do we need to add this comment for CassandraBridgeFactory's close as well?

cassandra-analytics-core/src/test/java/org/apache/cassandra/bridge/CassandraBridgeFactoryShutdownHookTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Test
+    public void testShutdownHookRegisteredOnce()
+    {
+        CassandraBridgeFactory.maybeRegisterShutdownHook();
+        assertThat(CassandraBridgeFactory.isShutdownHookRegistered()).isTrue();
+
+        // Second call is idempotent — the flag stays true and no exception is thrown
+        CassandraBridgeFactory.maybeRegisterShutdownHook();
+        assertThat(CassandraBridgeFactory.isShutdownHookRegistered()).isTrue();
```
skoppu22, line 46: Would be helpful to have a test to verify close() is indeed clearing the cache.

cassandra-bridge/src/main/java/org/apache/cassandra/bridge/BaseCassandraBridgeFactory.java
skoppu22: Do we still need this or can we remove it? This one is creating uncached classloader which is never closed.

skoppu22:
LGTM, only minor comments

cassandra-analytics-cdc/src/test/java/org/apache/cassandra/cdc/test/TestCdcBridgeProvider.java
```diff
@@ -35,39 +36,72 @@
...
             }
-            catch (IOException e)
+            catch (IOException ignored)
             {
-                throw new IllegalStateException(e);
+                // best-effort cleanup
             }
         });
+        COMMIT_LOG_DIRS.clear();
+        CdcBridgeFactory.close();
```
skoppu22, line 100: Don't we need to clear TypeCache here?

cassandra-analytics-core/src/test/java/org/apache/cassandra/bridge/CassandraBridgeFactoryShutdownHookTest.java
jmckenzie-dev: Good call; added both a method for external tests to confirm bridges are closed and a test that populates, confirm live, then closes and confirm closed. We don't necessarily introspect to ensure that the _internal_ close runs actually succeeded (vs. the warn path on catch if they fail), but it at least indicates it went through that path and operated upon them.

jmckenzie-dev:
Rebased, doing a final run of CI, then will merge.

jmckenzie-dev:
CI looks mostly clean; failures are the intermittent ones I'm trying to fix here and in other PR's. Merging.

PR #195, opened 2026-04-08
CASSANALYTICS-148: Revert circle config to default to big format

jmckenzie-dev:
Patch by Josh McKenzie; reviewed by Doug Rohrer for CASSANALYTICS-148


JeetKunDoug (commented):
+1 - thanks for taking care of this.

PR #196, opened 2026-04-09
CASSANALYTICS-149 Change log level for logs in CassandraSchema to debug level to avoid log spamming

jyothsnakonisa:


jyothsnakonisa:
CircleCI: https://app.circleci.com/pipelines/gh/jyothsnakonisa/cassandra-analytics/76/workflows/764d972f-8f81-4b63-8dd9-3f417daff877

cassandra-four-zero-types/src/main/java/org/apache/cassandra/bridge/CassandraSchema.java
```diff
@@ -288,10 +288,14 @@ public static void updateCdcSchema(@NotNull Schema schema,
...
     {
-        LOGGER.info("Updating CDC schema tables='{}'",
+        if (LOGGER.isDebugEnabled())
+        {
+            LOGGER.debug("Updating CDC schema tables='{}'",
                     cdcTables.stream()
-                             .map(t -> String.format("%s.%s", t.keyspace(), t.table()))
-                             .collect(Collectors.joining(",")));
+                            .map(t -> String.format("%s.%s", t.keyspace(), t.table()))
+                            .collect(Collectors.joining(",")));
```
yifan-c: lines are not aligned.

yifan-c (approved):
have a nit. +1 in advance

cassandra-four-zero-types/src/main/java/org/apache/cassandra/bridge/CassandraSchema.java
```diff
@@ -288,10 +288,14 @@ public static void updateCdcSchema(@NotNull Schema schema,
                                        @NotNull Partitioner partitioner,
                                        @NotNull TableIdLookup tableIdLookup)
     {
-        LOGGER.info("Updating CDC schema tables='{}'",
+        if (LOGGER.isDebugEnabled())
```
skoppu22, line 291: Do we need explicit isDebugEnabled check? I believe LOGGER.debug work only when debug enabled right?

jyothsnakonisa: Yifan flagged that we do this following computation in the log. Even if debug logging is not enabled it gets computed if we don't have the check. This is an performance optimization

```
cdcTables.stream()
                 .map(t -> String.format("%s.%s", t.keyspace(), t.table()))
                 .collect(Collectors.joining(","))
```

PR #200, opened 2026-04-20
CASSANALYTICS-154: Make KafkaPublisher class generic to publish GenericRecords for different value serializers

jyothsnakonisa:


cassandra-analytics-cdc-codec/src/main/java/org/apache/cassandra/cdc/avro/AvroSerializer.java
```diff
@@ -133,12 +133,21 @@ public Deserializer(Schema cdcSchema, SchemaStore store)
         }
 
         public CdcEnvelope deserialize(String keyspace, String table, byte[] data)
+        {
+            return deserialize(keyspace, table, data, null);
```
jmckenzie-dev: I've traced it through, and I can't find where that 4th param is used down the chain. We're effectively passing a null value for a string that's a ticking time bomb for someone to NPE out on in the future, and afaict nobody is actually using this param right now.

Specifically, the 3 non-test implementations of `SchemaStore#getReader` are:
1. `CachingSchemaStore`: doesn't use that param
2. `LocalTableSchemaStore`: doesn't use that param
3. `LocalTableSchemaStore`: doesn't use the param

So they use the _namespace_, or their own locally cached data, but this nullable string looks like a ride-along vestigial param. Unless I'm missing something we should delete it entirely.

cassandra-analytics-cdc-codec/src/main/java/org/apache/cassandra/cdc/kafka/GenericRecordKafkaPublisher.java
```diff
@@ -0,0 +1,78 @@
...
+                                boolean failOnKafkaError,
+                                CdcLogMode logMode,
+                                KafkaStats kafkaStats,
+                                RecordProducer<GenericData.Record> recordProducer,
+                                EventHasher eventHasher) {
+        super(version, topicSupplier, producer, schemaStore,
+                maxRecordSizeBytes, failOnRecordTooLargeError, failOnKafkaError, logMode,
+                kafkaStats, recordProducer, eventHasher);
+        this.transformer = new AvroGenericRecordTransformer(schemaStore, typeLookup, schemaNamespacePrefix);
+        ;
```
jmckenzie-dev: ... Why hello there little floating `;`. :) We should remove this.

cassandra-analytics-cdc-codec/src/main/java/org/apache/cassandra/cdc/kafka/GenericRecordKafkaPublisher.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.cassandra.cdc.avro.AvroGenericRecordTransformer;
+import org.apache.cassandra.cdc.avro.CdcEventAvroEncoder;
+import org.apache.cassandra.cdc.msg.CdcEvent;
+import org.apache.cassandra.cdc.schemastore.SchemaStore;
+import org.apache.cassandra.spark.data.CqlField;
+import org.apache.kafka.clients.producer.KafkaProducer;
+
+import java.util.function.Function;
+
+public class GenericRecordKafkaPublisher extends KafkaPublisher<GenericData.Record> {
```
jmckenzie-dev: Much as it pains me, I believe braces need to be on newlines to be consistent w/the rest of the codebase.

cassandra-analytics-cdc-codec/src/main/java/org/apache/cassandra/cdc/kafka/ByteArrayKafkaPublisher.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * {@link KafkaPublisher} for the no-registry path where {@code value.serializer} is
+ * {@code org.apache.kafka.common.serialization.ByteArraySerializer}.
+ *
+ * <p>Overrides {@link #getPayload} to Avro-encode the transformed {@link GenericData.Record}
+ * to {@code byte[]} before handing it to the producer, since {@code ByteArraySerializer}
+ * is a pass-through and expects pre-encoded bytes.
+ */
+public class ByteArrayKafkaPublisher extends KafkaPublisher<byte[]> {
```
jmckenzie-dev: bracing style

cassandra-analytics-cdc-codec/src/main/java/org/apache/cassandra/cdc/kafka/KafkaPublisher.java
```diff
@@ -263,6 +276,7 @@ protected String getOrBuildKafkaPrefix(CdcEvent event)
                );
     }
 
+
```
jmckenzie-dev, line 283: Think this is an accidental newline.

cassandra-analytics-cdc-codec/src/main/java/org/apache/cassandra/cdc/kafka/GenericRecordKafkaPublisher.java
jmckenzie-dev: Could we get a javadoc explaining why this class exists and its purpose in the grand scheme of things?

cassandra-analytics-cdc-codec/src/main/java/org/apache/cassandra/cdc/kafka/RecordProducer.java
jmckenzie-dev, line 1: Could we remove some of the unused `buildRecords` methods in this file? It's cluttered with a bunch of them and it looks like we have 2 `buildRecords`, 1 `buildRecord`, and 2 `addHeader` that aren't in use, at least as far as a single project introspection is concerned.

I grepped through `cassandra-analytics` and didn't find any usages so think this is probably safe.

jmckenzie-dev (changes requested):

cassandra-analytics-cdc-codec/src/main/java/org/apache/cassandra/cdc/avro/AvroSerializer.java
jyothsnakonisa: `deserialize` method will be used by the consumer of CDC message. when they pass schemaUUID in the method, it will be used to get the schema from schemastore and the message will be deserialized.

jmckenzie-dev: My concern is about that 4th parameter to `deserialize`:
```
public CdcEnvelope deserialize(String keyspace, String table, byte[] data, String schemaUuid)
```
This goes through:
```
GenericRecord payload = deserializePayload(keyspace, table, schemaUuid, getPayload(header));
```
Then on to:
```
GenericDatumReader<GenericRecord> payloadReader = store.getReader(keyspace + "." + table, schemaUuid);
```
In that signature:
```
GenericDatumReader<GenericRecord> getReader(String namespace, String name);
```
What was `schemaUuid` before becomes `name`, and none of the 3 production implementations of that interface method make any use of that parameter. So I'm saying we just delete it entirely from the call stack; it looks like it's vestigial and right now it's just a `null` param getting passed around waiting for someone to try and use it in the future and NPE out.

jyothsnakonisa: Makes sense, I missed the NPE part earlier. I have updated the deserialize to use non null schema uuid.

Also, note that this class is not used anywhere now.

jmckenzie-dev (approved):
Looks good. Thanks!

cassandra-analytics-cdc-codec/src/main/java/org/apache/cassandra/cdc/avro/AvroSchemaUtils.java
```diff
@@ -0,0 +1,115 @@
...
+     * produce the same schema fingerprint.
+     *
+     * @param payloadSchema the table-specific Avro schema produced by the CQL-to-Avro converter
+     * @param name          the record name for the merged schema
+     * @param namespace     the namespace for the merged schema
+     * @return merged Avro schema
+     */
+    public static Schema buildMergedSchema(Schema payloadSchema, String name, String namespace)
+    {
+        InputStream is = AvroSchemaUtils.class.getClassLoader().getResourceAsStream(CDC_ENVELOPE_TEMPLATE);
```
skoppu22: Looks like inputstream is not closed. Can we do this in try( ... ) to ensure the closing?

cassandra-analytics-cdc-codec/src/test/java/org/apache/cassandra/cdc/kafka/ByteArrayKafkaPublisherTest.java
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.cassandra.cdc.msg.CdcEvent;
+import org.apache.cassandra.cdc.msg.CdcEventBuilder;
+import org.apache.cassandra.cdc.msg.Value;
+import org.apache.cassandra.cdc.schemastore.LocalTableSchemaStore;
+import org.apache.cassandra.spark.data.CqlField;
+
+import static org.apache.cassandra.spark.utils.ArrayUtils.listOf;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ByteArrayKafkaPublisherTest
```
skoppu22, line 49: Where do we test GenericRecordKafkaPublisher?

cassandra-analytics-cdc-codec/src/main/java/org/apache/cassandra/cdc/avro/AvroSchemaUtils.java
```diff
@@ -0,0 +1,114 @@
...
+     * produce the same schema fingerprint.
+     *
+     * @param payloadSchema the table-specific Avro schema produced by the CQL-to-Avro converter
+     * @param name          the record name for the merged schema
+     * @param namespace     the namespace for the merged schema
+     * @return merged Avro schema
+     */
+    public static Schema buildMergedSchema(Schema payloadSchema, String name, String namespace)
+    {
+        try (InputStream is = AvroSchemaUtils.class.getClassLoader().getResourceAsStream(CDC_ENVELOPE_TEMPLATE))
```
jmckenzie-dev: We're doing this for every message right? It kind of seems like we'd want to cache this value rather than reading it each time, even if the current impl is just a decompression from memory. It looks like we cache this value in `CdcEventAvroEncoder.java` already in the ctor calling `readSchema` - maybe we do something similar here?

skoppu22 (approved):

jyothsnakonisa:
CI : https://app.circleci.com/pipelines/github/jyothsnakonisa/cassandra-analytics?branch=schema-uuid

jmckenzie-dev:
👍🏻 Good to merge.

PR #206, opened 2026-05-12
CASSANALYTICS-42 Add S3-backed Cassandra batch reader

liucao-dd:
## Patch information

Jira: https://issues.apache.org/jira/browse/CASSANALYTICS-42
CEP: https://cwiki.apache.org/confluence/display/CASSANDRA/CEP-56%3A+Spark+Bulk+Reading+from+Cassandra+Backup+Uploaded+to+Object+Storage

## Summary

Adds an S3-backed Cassandra batch reader so Spark jobs can read SSTables directly from object-storage backups without going through a live cluster. Built on top of the existing `CassandraDataLayer` / bulk-reader abstractions and exposed through Spark SQL via a new `S3CassandraDataSource`.

This PR ships only the generic SPI and the S3 reference implementation. Concrete backup-provider implementations are out of scope and intended to live in downstream repos that plug into the `BackupReader` SPI introduced here.

## What's included

- **Backup reader SPI** (`cassandra-analytics-core/.../spark/data/backup/`): `BackupReader`, `BackupReaderConfig`, `BackupReaderFactory`, `BackupReaderRegistry` — pluggable abstraction so different backup providers can be wired in without changes to core.
- **S3 data layer** (`cassandra-analytics-core`):
  - `S3CassandraDataLayer` with token-aware partitioning and SSTable selection over S3-resident backups, including `SSTableTokenBounds` pruning that correctly handles the Murmur3 wrap-around.
  - `S3ClientCache`, `S3ClientConfig`, `S3DataSourceClientConfig`, `S3SizingFactory`, `S3TableSizeProvider`, and `S3SnapshotTimeProvider` (in `cassandra-analytics-common`).
- **Spark SQL integration**: `S3CassandraDataSource`, `S3CassandraPrebuiltReadContext(Registry)`, `S3CassandraTokenIndexPrebuilder`, plus refinements to `CassandraScanBuilder`, `CassandraPartitioning`, `CassandraTable`, and a new `CassandraSourceStatistics`.
- **Per-task metrics**: `SparkCustomMetricsStats` plus Spark `CustomTaskMetric` classes (`TaskTotal*` / `Total*`) for S3 GET/HEAD latency, summary read latency, skipped/corrupt SSTable counts, opened SSTable duration, mutable metadata drift, and head fallback counts. Threaded through `BackupReader` read paths via a per-task `Stats` argument.
- **Bridge plumbing**: minor changes to `CassandraBridgeImplementation` and `SummaryDbUtils` (both bridges) to thread the per-task `Stats` parameter; existing call sites get backwards-compatible overloads. `VersionRunner` is published via `java-test-fixtures` so the bridges can share it in tests.

## Tests

All new tests live in the modules where the code lives (`cassandra-analytics-core`, bridges):

- `S3ClientCacheTest`, `S3ClientConfigTest`, `S3DataSourceClientConfigTest`, `S3DataSourceClientConfigBufferTest`
- `S3SSTableLeakTests`, `SSTableTokenIndexTest`
- `BackupReaderFactorySerializationTest` plus a `FakeBackupReader` test fixture exercising the SPI without a real S3 endpoint
- `S3CassandraPrebuiltReadContextTest`
- Updated `ReaderUtilsTests`, `SSTableReaderTests`, `SummaryDbTests` for the new `Stats`-threaded signatures

The code is internally powering ~200 spark pipelines in production for ingesting data from s3 backup into data lake, ranging from 10MB to 300TB table size.

## Reviewer notes

The diff is large (~9.4k LOC added across ~80 files) because it introduces both the SPI and a complete S3 implementation plus the metrics surface. This is mostly to show the overall idea and actual shipping will likely come with smaller PRs if preferred.

Please also ignore my shadow jar build changes, those are artifacts of our internal build system that we can discuss cleanup later.

## Known gap currently deferred
1. s3 credential is not wired through sidecar
2. we use in memory buffer when reading from s3, and this works fine in prod when we have S3 integrated encryption (e.g. KMS), but may or may not work if other encryption mechanism exists that needs handling outside of S3 API.


jberragan (commented):
Thanks for the patch, this is a great addition! I will take a closer look, but a few high level questions:
- Did you consider putting this in a separate module (e.g. `cassandra-analytics-s3`). The Sidecar DataLayer should never had been put in `cassandra-analytics-core` and should eventually be moved to its own module.
- I suppose every user might have a slightly different backup path format, is this made pluggable by the `BackupReader` interface?

liucao-dd:
> Thanks for the patch, this is a great addition! I will take a closer look, but a few high level questions:
> 
> * Did you consider putting this in a separate module (e.g. `cassandra-analytics-s3`). The Sidecar DataLayer should never had been put in `cassandra-analytics-core` and should eventually be moved to its own module.

wasn't aware that we wanted the sidecar data layer to live outside of core. happy to discuss the appropriate modularization. A s3 specific one could work, or we can have a batch read specific module? the s3 implementation inherits a good amount of work from the sidecar one.

> * I suppose every user might have a slightly different backup path format, is this made pluggable by the `BackupReader` interface?

yes, i didn't include the internal concrete implementation of this interface on this PR because it is tied to a vendor we use. But the goal is to make it pluggable per needs of different organizations, using the `BackupReaderRegistry.register()`

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/FileType.java
```diff
@@ -83,4 +83,15 @@ public String getFileSuffix()
...
+
+    /**
+     * Whether the on-disk size of this component can drift from the value recorded in a backup
+     * manifest. Cassandra rewrites Summary/Filter/Statistics in place during compaction, so a
+     * stale manifest size for these components can produce truncated ranged-GETs against the
+     * backing store. The data layer treats these components specially when issuing reads.
+     */
+    public boolean isMutableMetadata()
+    {
+        return this == SUMMARY || this == FILTER || this == STATISTICS;
```
jberragan, line 95: Is this code/comment fully accurate? I'm aware Statistics.db can mutate the repair metadata (repairedAt, pendingRepair) to avoid recompaction, and also the compaction level to avoid a full rewrite.

Summary.db appears to only mutate if there is a change in the index sample size (redistributeSummaries).

Filter.db I can't find any place where it mutates. I think it is immutable from the first flush.

liucao-dd: Good catch, i think filter.db is theoretically immutable but there is a code path that re-writes it:

https://github.com/apache/cassandra/blob/45fa31b/src/java/org/apache/cassandra/io/sstable/format/SSTableReaderBuilder.java#L450
It should be very rare, and if I read the code correctly, it happens during live re-opens of an SSTable (startup, import, streaming receipt), and only when its Filter.db is missing or it has no ValidationMetadata.

So it is theoretically possible but most likely can be treated immutable.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/SummaryDbUtils.java
```diff
@@ -113,9 +119,17 @@ static Summary readSummary(InputStream summaryStream,
...
-            return new Summary(indexSummary, firstKey, lastKey);
+            try
+            {
+                DecoratedKey firstKey = partitioner.decorateKey(ByteBufferUtil.readWithLength(is));
+                DecoratedKey lastKey = partitioner.decorateKey(ByteBufferUtil.readWithLength(is));
+                return new Summary(indexSummary, firstKey, lastKey);
+            }
+            catch (IOException | RuntimeException exception)
+            {
+                indexSummary.close();
```
jberragan, line 130: This is a separate known issue, I think it is already being fixed in https://github.com/apache/cassandra-analytics/pull/138.

Please don't call `indexSummary.close()`, iirc it resulted in in seg. faults (SIGSEGV) due to the way Casandra references tracks.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/S3ClientCache.java
```diff
@@ -0,0 +1,555 @@
...
+/**
+ * Unified cache for S3 clients and transfer managers.
+ * <p>
+ * This class consolidates all S3 client creation across the codebase into a single
+ * cached location. Clients are keyed by region|endpoint|accessKeyId|secretHash for
+ * proper isolation across different configurations.
+ * <p>
+ * Key design decisions:
+ * <ul>
+ *   <li>No JVM shutdown hooks - problematic in Spark executors. Use explicit {@link #closeAll()}</li>
```
jberragan, line 53: 👍

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/S3ClientCache.java
```diff
@@ -0,0 +1,555 @@
...
+            {
+                manager.close();
+            }
+            catch (Exception e)
+            {
+                LOGGER.warn("Error closing S3TransferManager for key {}: {}", key, e.getMessage());
+            }
+        }
+
+        // Close async client
```
jberragan, line 459: Is it a problem if two threads call close at the same time? One thread can skip ahead and close asyncClient while S3TransferManager is still shutting down.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/S3SnapshotTimeProvider.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * A {@link TimeProvider} for S3-based reads that sets the reference epoch from autosnap timestamps.
+ * <p>
+ * The reference epoch drives TTL and tombstone evaluation in {@code CompactionIterator}'s
+ * {@code cell.isLive(nowInSec)} checks. To avoid silently dropping cells that were still alive
+ * at the time of an earlier node's snapshot, the reference epoch should be the <b>earliest</b>
+ * (minimum) autosnap epoch across all nodes. This ensures conservative TTL behavior: a cell
+ * is only expired if it was already expired at every node's snapshot time.
+ */
+public class S3SnapshotTimeProvider implements TimeProvider
```
jberragan, line 31: nit: there's very little difference vs `ReaderTimeProvider`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/backup/BackupReader.java
```diff
@@ -0,0 +1,302 @@
...
+     *
+     * @param clusterName logical cluster identity
+     * @param datacenter  datacenter to read from
+     * @param token       Cassandra token
+     * @param sstableKey  identifies the SSTable
+     * @param fileType    SSTable component
+     * @param stats       per-task stats sink
+     * @return {@code true} iff the component exists in the backup
+     */
+    boolean exists(String clusterName,
```
jberragan, line 209: Shouldn't this be a `CompletableFuture<Boolean>`? Checking for existence in S3 would still be blocking, right?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/DataLayer.java
```diff
@@ -234,6 +235,22 @@ public abstract SSTablesSupplier sstables(int partitionId,
...
+    /**
+     * Calculate the total size of all SSTable Data.db files across all partitions.
+     * This method provides a default implementation that returns empty. Specific DataLayer
+     * implementations should override this method if they have efficient ways to
+     * calculate the total size.
+     *
+     * @return OptionalLong containing the total size in bytes of all SSTable Data.db files,
+     * or empty if size information is not available
+     */
+    public OptionalLong calculateTotalSSTableSize()
```
jberragan, line 247: There's an existing interface `TableSizeProvider` to help with dynamically sizing the spark job, maybe you want to re-use here somehow?

cassandra-analytics-core/src/main/spark3/org/apache/cassandra/spark/sparksql/CassandraScanBuilder.java
```diff
@@ -146,4 +190,30 @@ private PartitionKeyFilter buildFilter(List<String> keys)
...
+        OptionalLong estimatedSizeInBytes = dataLayer.calculateTotalSSTableSize();
+        return new CassandraSourceStatistics(
+        estimatedSizeInBytes,
+        OptionalLong.empty(), // numRows - not calculated for now
+        Collections.emptyMap() // columnStats - empty for now
+        );
+    }
+
+    @Override
+    public CustomMetric[] supportedCustomMetrics()
```
jberragan, line 206: Nice!

skoppu22:
Nice work, interfaces are genuinely pluggable with vendor specific implementations and storage neutral.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/S3DataSourceClientConfig.java
```diff
@@ -0,0 +1,625 @@
...
+     * are not carried on the config object; instead callers pass the appropriate {@code Stats}
+     * instance per call into
+     * {@link org.apache.cassandra.spark.data.backup.BackupReader} read methods (executor-local
+     * {@code SparkCustomMetricsStats} for reads, {@code DoNothingStats.INSTANCE} for prebuild).
+     *
+     * @return a {@link BackupReaderConfig} populated from this config's S3 client settings
+     */
+    public BackupReaderConfig toBackupReaderConfig()
+    {
+        return BackupReaderConfig.of(s3Config);
```
skoppu22, line 434: custom properties are lost here, not sent to BackupReaderConfig

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/backup/BackupReader.java
```diff
@@ -0,0 +1,302 @@
...
+     * @return stable fingerprint string identifying the materialized manifest set
+     */
+    String getManifestFingerprint(String clusterName);
+
+    /**
+     * Returns the underlying {@link S3ClientConfig}. Used for intern-cache identity.
+     *
+     * @return the S3 client configuration this reader was constructed with
+     */
+    S3ClientConfig s3Config();
```
skoppu22, line 253: Organizations may use other types of Object stores as well. Instead of using S3 specific naming, would be helpful if we can use generic ObjectStore* namings.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/S3ClientCache.java
```diff
@@ -0,0 +1,555 @@
...
+    }
+
+    private static int clamp(long value, int floor, int ceiling)
+    {
+        return (int) Math.max(floor, Math.min(ceiling, value));
+    }
+
+    private static AwsCredentialsProvider getCredentialsProvider(S3ClientConfig config)
+    {
+        String accessKeyId = config.s3AccessKeyId();
```
skoppu22, line 429: Would be helpful to have pluggable credentials interfaces, like S3CredentialsProviderFactory, S3CredentialsProviderRegistry etc

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/backup/BackupReader.java
```diff
@@ -0,0 +1,302 @@
...
+     * @param datacenter  datacenter to read from
+     * @param token       Cassandra token (used for path resolution by some implementations)
+     * @param sstableKey  identifies the SSTable
+     * @param fileType    which SSTable component to read
+     * @param start       inclusive byte offset
+     * @param end         inclusive byte offset
+     * @param stats       per-task stats sink for S3 operation metrics
+     * @return future completing with the read bytes
+     */
+    CompletableFuture<byte[]> readAsync(String clusterName,
```
skoppu22, line 116: There is no sharable Decrypter interface. It would be helpful to have a BackupDecryptor interface so organizations can implement BackupDecryptor only to tune policies while sharing other parts of the logic.

readAsync ─► decryptor.decryptRange(...)
getAsync  ─► decryptor.decryptRange(...)

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/S3CassandraDataLayer.java
```diff
@@ -0,0 +1,1769 @@
...
+            Long actualSize = ssTable.actualComponentSizes.get(fileType);
+            this.actualSizeResolved = actualSize != null;
+            this.size = actualSize == null ? size : actualSize;
+        }
+
+        public void request(long start, long end, StreamConsumer consumer)
+        {
+            // Mutable metadata (Summary.db, Filter.db, Statistics.db) supports
+            // size-drift handling for stale autosnap manifests.
+            if (fileType.isMutableMetadata())
```
skoppu22, line 1380: Under client-side encryption ciphertext length ≠ plaintext length and cannot be inferred from the object size. So a client-side-encrypted Data.db/Index.db whose manifest size is the ciphertext size has no interface to resolve the real plaintext length

It would be helpful if you add a generic, per-component length hook to BackupReader and consult+cache it for all components, not just mutable ones.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/S3ClientConfig.java
```diff
@@ -0,0 +1,234 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import org.apache.cassandra.spark.utils.MapUtils;
+
+/**
+ * Minimal S3 client configuration for connecting to S3.
+ * This class contains only the essential settings needed to create an S3 client.
+ * It is shared across batch and streaming configurations.
+ */
+public class S3ClientConfig implements Serializable
```
skoppu22, line 38: S3ClientConfig carries no encryption identity, and buildS3Client/buildS3AsyncClient hardcode S3Client.builder() / S3AsyncClient.builder() with no injection point. So an org that wants an S3 Encryption Client cannot supply one.

getCacheKey hashes only region|endpoint|accessKeyId|secretHash — it has no encryption identity — so a plain client and an encrypting client built for the same bucket/credentials would collide on one cache entry: the first one to be created wins and is silently handed to the other caller

cassandra-analytics-core/src/main/spark3/org/apache/cassandra/spark/sparksql/S3CassandraDataSource.java
```diff
@@ -0,0 +1,96 @@
...
+     * Driver-local handle produced by {@link S3CassandraTokenIndexPrebuilder}. This is intentionally
+     * not a durable DataSource option: saved/replayed plans and remote client/server boundaries must
+     * rebuild the context in the JVM that plans the read.
+     */
+    public static final String READ_CONTEXT_ID_KEY = "s3CassandraReadContextId";
+
+    @Override
+    public String shortName()
+    {
+        return "s3CassandraBulkRead";
```
skoppu22, line 48: S3 is in the public API, even though the interfaces (BackupReader/BackupReaderRegistry) are storage-neutral and backupReaderType already selects the backend. Combined with the s3-region / s3-bucket / s3-* option keys, callers of a non-S3 backend still type s3CassandraBulkRead and s3- options.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/S3CassandraDataLayer.java
```diff
@@ -0,0 +1,1769 @@
...
+     */
+    protected void shutdownHook()
+    {
+        try
+        {
+            if (s3BackupReader != null)
+            {
+                s3BackupReader.close();
+            }
+            S3ClientCache.closeAll();
```
skoppu22, line 1154: A non-S3 reader that owns its own connection pool (e.g., a GCS/Azure client or an Apple Manager-proxy pool) has no hook to release it: the layer only knows how to tear down S3.

A generic call something like below would be helpful

`BackupReaderRegistry.factoryFor(backupReaderType).closeSharedResources();`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/S3CassandraDataLayer.java
```diff
@@ -0,0 +1,1769 @@
...
+
+        initializeS3BackupReader();
+
+        // list Cassandra instances in S3 bucket
+        final List<CassandraInstance> instances = s3BackupReader.instances(clusterName, config.keyspace(), config.table(), config.datacenter());
+        // build CassandraRing and TokenPartitioner. Prefer rack-aware authoritative replica
+        // placement from the BackupReader; fall back to the naive (rack-unaware) ring when
+        // none is available. Exceptions from the reader signal a genuine integrity issue and
+        // must surface — see BackupReader#buildRackAwareReplicas for the contract.
+        final Partitioner partitioner = Partitioner.Murmur3Partitioner;
```
skoppu22, line 511: Murmur3Partitioner is hardcoded here. Can we provide a hook to configure partitioners, something like 

`Partitioner partitioner = s3BackupReader.partitioner();`

skoppu22:
> Nice work, interfaces are genuinely pluggable with vendor specific implementations and storage neutral.

@liucao-dd Some of my comments are to make this feature support client encryption use cases or non-s3 object stores etc. I do not expect you to fix all those in this PR. For comments not resolved in this PR, please feel free to create JIRAs if required, so we can implement them later.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/data/backup/BackupReader.java
skoppu22: Also, would be good to have `KeyService` interface with  `byte[] unwrapDek(String keyId, byte[] wrappedDek) `method.

PR #209, opened 2026-05-18
CASSANALYTICS-52: Downgrade noisy TokenPartitioner partition/range maps to DEBUG

tejaslodayadd:
https://issues.apache.org/jira/browse/CASSANALYTICS-52

### DIFF
--- a/cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/partitioner/TokenPartitioner.java
+++ b/cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/partitioner/TokenPartitioner.java
@@ -116,9 +116,9 @@ public class TokenPartitioner implements Serializable
         validateCompleteRangeCoverage();
         validateRangesDoNotOverlap();
 
-        LOGGER.info("Number of partitions {}", reversePartitionMap.size());
-        LOGGER.info("Partition map " + partitionMap);
-        LOGGER.info("Reverse partition map " + reversePartitionMap);
+        LOGGER.debug("Number of partitions {}", reversePartitionMap.size());
+        LOGGER.debug("Partition map {}", partitionMap);
+        LOGGER.debug("Reverse partition map {}", reversePartitionMap);
     }
 
     private static int calculateSplits(CassandraRing ring, int defaultParallelism, Integer cores)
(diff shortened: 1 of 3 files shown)


cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/partitioner/TokenPartitioner.java
```diff
@@ -116,17 +116,17 @@ private void calculateTokenRangeMap()
         validateCompleteRangeCoverage();
         validateRangesDoNotOverlap();
 
-        LOGGER.info("Number of partitions {}", reversePartitionMap.size());
-        LOGGER.info("Partition map " + partitionMap);
-        LOGGER.info("Reverse partition map " + reversePartitionMap);
+        LOGGER.debug("Number of partitions {}", reversePartitionMap.size());
```
frankgh, line 119: I have found this information useful when debugging jobs in the past. I do agree that these log messages are not needed in the majority of case. I wonder if there's a better way to control when to log this information.

tejaslodayadd: probably a JavaDoc comment on calculateTokenRangeMap()?

tejaslodayadd: @frankgh is there a better way?

yifan-c: One option to reduce the verbosity is to only log at the driver instance. Currently the 3 messages are logged by every spark task. To remove the log messages from the executor nodes, we can skip logging in the code path of `org.apache.cassandra.spark.data.partitioner.TokenPartitioner.Serializer#read`

I'd strongly suggest keeping the log messages at the info level. They are useful when debugging.

frankgh: yeah, I second Yifan here. We should keep the log message. +1 to logging this from the driver

PR #211, opened 2026-05-27
CASSANALYTICS-167: Regenerate bloom filters for CQLSSTableWriter

lukasz-antoniak:
Fixes [CASSANALYTICS-167](https://issues.apache.org/jira/browse/CASSANALYTICS-167).

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
@@ -43,12 +43,14 @@ import org.apache.cassandra.bridge.CassandraVersionFeatures;
 import org.apache.cassandra.bridge.SSTableDescriptor;
 import org.apache.cassandra.spark.common.Digest;
 import org.apache.cassandra.spark.common.SSTables;
+import org.apache.cassandra.spark.data.FileSystemSSTable;
 import org.apache.cassandra.spark.data.FileType;
 import org.apache.cassandra.spark.data.LocalDataLayer;
 import org.apache.cassandra.spark.data.partitioner.Partitioner;
 import org.apache.cassandra.spark.reader.RowData;
 import org.apache.cassandra.spark.reader.StreamScanner;
 import org.apache.cassandra.spark.sparksql.filters.SSTableTimeRangeFilter;
+import org.apache.cassandra.spark.stats.BufferingInputStreamStats;
 import org.apache.cassandra.spark.utils.DigestAlgorithm;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -227,6 +229,9 @@ public class SortedSSTableWriter
         };
         Set<Path> dataFilePaths = new HashSet<>();
         Map<Path, Digest> fileDigests = new HashMap<>();
+        // FIXME: CQLSSTableWriter may produce incomplete Filter.db file, rebuilding it manually (see CASSANDRA-21423).
+        // rebuild Filter.db files before calculating their digest
+        rebuildFilterComponents(writerContext, sstableFilter);
         try (DirectoryStream<Path> stream = Files.newDirectoryStream(getOutDir(), sstableFilter))
         {
             for (Path path : stream)
@@ -281,20 +286,41 @@ public class SortedSSTableWriter
         // Only process new SSTables produced during final flush
         DirectoryStream.Filter<Path> unhashedFilter = path -> !hashedFiles.contains(path);
 
-        for (Path dataFile : getDataFileStream(unhashedFilter))
+        // FIXME: CQLSSTableWriter may produce incomplete Filter.db file, rebuilding it manually (see CASSANDRA-21423).
+        rebuildFilterComponents(writerContext, unhashedFilter);
+
+        try (DirectoryStream<Path> dataFileStream = getDataFileStream(unhashedFilter))
         {
-            // NOTE: We calculate file hashes before re-reading so that we know what we hashed
-            //       is what we validated. Then we send these along with the files and the
-            //       receiving end re-hashes the files to make sure they still match.
-            Map<Path, Digest> newFileDigests = calculateFileDigestMap(dataFile);
-            overallFileDigests.putAll(newFileDigests);
-            newlyHashedFiles.addAll(newFileDigests.keySet());
-            sstableCount += 1;
+            for (Path dataFile : dataFileStream)
+            {
+                // NOTE: We calculate file hashes before re-reading so that we know what we hashed
+                //       is what we validated. Then we send these along with the files and the
+                //       receiving end re-hashes the files to make sure they still match.
+                Map<Path, Digest> newFileDigests = calculateFileDigestMap(dataFile);
+                overallFileDigests.putAll(newFileDigests);
+                newlyHashedFiles.addAll(newFileDigests.keySet());
+                sstableCount += 1;
+            }
         }
         // Only calculate size for newly hashed files, not all files in overallFileDigests
         // (previously hashed files may have been deleted by prepareSStablesToSend)
         bytesWritten += calculatedTotalSize(newlyHashedFiles);
         validateSSTables(writerContext);
+    }
+
+    protected void rebuildFilterComponents(@NotNull BulkWriterContext writerContext,
+                                           @NotNull DirectoryStream.Filter<Path> filter) throws IOException
+    {
+        LocalDataLayer layer = buildLocalDataLayer(writerContext, getOutDir(), null);
+        try (DirectoryStream<Path> dataFileStream = getDataFileStream(filter))
+        {
+            for (Path dataFile : dataFileStream)
+            {
+                FileSystemSSTable ssTable = new FileSystemSSTable(dataFile, false, BufferingInputStreamStats::doNothingStats);
+                writerContext.bridge().rebuildBloomFilter(layer.partitioner(), layer.cqlTable(), ssTable, getOutDir());
+                LOGGER.debug("Rebuilt bloom filter for sstable {}", dataFile);
+            }
+        }
     }
 
     @VisibleForTesting
@@ -319,26 +345,7 @@ public class SortedSSTableWriter
         //       and then validate all of them in parallel threads
         try
         {
-            CassandraVersion version = CassandraBridgeFactory.getCassandraVersion(writerContext.cluster().getLowestCassandraVersion());
-            String keyspace = writerContext.job().qualifiedTableName().keyspace();
-            String schema = writerContext.schema().getTableSchema().createStatement;
-            Partitioner partitioner = writerContext.cluster().getPartitioner();
-            Set<String> udtStatements = writerContext.schema().getUserDefinedTypeStatements();
-            LocalDataLayer layer = new LocalDataLayer(version,
-                                                      partitioner,
-                                                      keyspace,
-                                                      schema,
-                                                      udtStatements,
-                                                      Collections.emptyList() /* requestedFeatures */,
-                                                      false /* useSSTableInputStream */,
-                                                      null /* statsClass */,
-                                                      SSTableTimeRangeFilter.ALL,
-                                                      outputDirectory.toString());
-            if (dataFilePaths != null)
-            {
-                layer.setDataFilePaths(dataFilePaths);
-            }
-
+            LocalDataLayer layer = buildLocalDataLayer(writerContext, outputDirectory, dataFilePaths);
             try (StreamScanner<RowData> scanner = layer.openCompactionScanner(partitionId, Collections.emptyList(), null))
             {
                 while (scanner.next())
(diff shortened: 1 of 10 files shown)


rustyrazorblade:
Solid patch.  Just a few things - nothing major.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -297,6 +311,25 @@ public synchronized void close(BulkWriterContext writerContext) throws IOExcepti
...
+                                         @NotNull DirectoryStream.Filter<Path> filter) throws IOException
+    {
+        LocalDataLayer layer = buildLocalDataLayer(writerContext, outputDirectory, null);
+        for (Path dataFile : getDataFileStream(filter))
+        {
+            try
+            {
+                FileSystemSSTable ssTable = new FileSystemSSTable(dataFile, false, BufferingInputStreamStats::doNothingStats);
+                writerContext.bridge().rebuildBloomFilter(layer.partitioner(), layer.cqlTable(), ssTable, outputDirectory);
+                LOGGER.error("Rebuilt bloom filter for sstable {}", dataFile);
```
rustyrazorblade: I think this should be INFO level.  It'll be confusing to users why every one of their jobs has errors.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/utils/FilterDbUtils.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public static IFilter buildBloomFilter(@NotNull Partitioner partitioner,
+                                           @NotNull CqlTable cqltable,
+                                           @NotNull SSTable ssTable) throws IOException
+    {
+        String keyspace = cqltable.keyspace();
+        String table = cqltable.table();
+
+        Map<MetadataType, MetadataComponent> componentMap = ReaderUtils.deserializeStatsMetadata(keyspace, table, ssTable, EnumSet.of(MetadataType.STATS));
+        StatsMetadata statsMetadata = (StatsMetadata) componentMap.get(MetadataType.STATS);
```
rustyrazorblade, line 53: Is there a possibility this could be NULL?

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
```diff
@@ -324,6 +333,44 @@ public List<ByteBuffer> encodePartitionKeys(Partitioner partitioner, String keys
...
+                    throw new IOException("Could not read Index.db file");
+                }
+                ReaderUtils.readPrimaryIndex(primaryIndex, tracker);
+            }
+
+            File filterFile = new File(directory.toFile(), descriptor.relativeFilenameFor(Component.FILTER));
+            try (FileOutputStream fos = new FileOutputStream(filterFile, false);
+                 DataOutputStreamPlus stream = new BufferedDataOutputStreamPlus(fos))
+            {
+                BloomFilterSerializer.serialize((BloomFilter) filter, stream);
```
rustyrazorblade, line 375: I'm not sure about this, but if the user has bloom filters disabled, I think this might throw an exception.  In 5.0 it uses IFilter.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriterTest.java
```diff
@@ -148,6 +162,71 @@ public void canCreateWriterForVersion(String version) throws IOException
...
+                                                         ssTable);
+
+        // second column is always set to 1 when inserting data
+        List<ByteBuffer> searchKeys = bridge.encodePartitionKeys(partitioner,
+                                                                 writerContext.qualifiedTableName().keyspace(),
+                                                                 writerContext.schema().getTableSchema().createStatement,
+                                                                 ImmutableList.of(ImmutableList.of("1", "1"), ImmutableList.of("7", "2")));
+
+        assertThat(bloomFilter.mightContain(searchKeys.get(0))).isTrue();
+        assertThat(bloomFilter.doesNotContain(searchKeys.get(1))).isTrue();
```
rustyrazorblade, line 230: Since we're using a limited set of data in here, it's very unlikely we'll ever hit this... so maybe just leave a comment.  If the test changes, we _could_ hit a bloom filter false positive, and it'll cause the test to fail.  I know it's really, really unlikely, so please just document it.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -297,6 +311,25 @@ public synchronized void close(BulkWriterContext writerContext) throws IOExcepti
...
+            }
+            catch (Exception e)
+            {
+                LOGGER.warn("Failed to rebuild bloom filter for sstable {}", dataFile, e);
```
rustyrazorblade: Is it safe to swallow this exception?  How would it fail?  Maybe make it an error?

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/TableSchemaTestCommon.java
```diff
@@ -31,6 +31,7 @@
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
+import org.apache.commons.lang.StringUtils;
```
rustyrazorblade: nit / question: Is this is commons lang3?

rustyrazorblade (commented):
Just a handful of small questions.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
lukasz-antoniak: For 5.0, we are using five-zero bridge and [this code](https://github.com/apache/cassandra-analytics/pull/211/changes#diff-fea9f2ebb75431ba0f30a7bf99ab10908e1fb730e09fd2b3fbc8b71a6fc6a201R366-R390).
Adding a condition to skip filter regeneration when `bloom_filter_fp_chance == 1.0`.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/utils/FilterDbUtils.java
lukasz-antoniak: From what I have reviewed in `CQLSSTableWriter`, all files (including statistics) have to be present once sstable is considered produced. Did I understand your concern correctly?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
lukasz-antoniak: I initially though to ignore failures, but producing unparseable file (and uploading it) may indeed cause an issue. Original 16B file is overwritten, so we do not publish the empty one in case of failure, but potentially broken one.

rustyrazorblade: Ok.  Mind just leaving a comment so it's clear that we're deliberately not uploading a bloom filter in case of a failure?  I don't expect we'll hit this path ever, I just want it to be clear of the intention.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/utils/FilterDbUtils.java
rustyrazorblade: Yep, just wanted to be sure.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -297,6 +311,18 @@ public synchronized void close(BulkWriterContext writerContext) throws IOExcepti
         validateSSTables(writerContext);
     }
 
+    protected void rebuildFilterComponents(@NotNull BulkWriterContext writerContext, @NotNull Path outputDirectory,
+                                         @NotNull DirectoryStream.Filter<Path> filter) throws IOException
```
yifan-c: nit: indentation is off by 2 spaces.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
```diff
@@ -324,6 +333,52 @@ public List<ByteBuffer> encodePartitionKeys(Partitioner partitioner, String keys
...
+                stream.flush();
+                SyncUtil.sync(fos);
+            }
```
yifan-c: Should we delete the `filterFile` if it fails to rebuild, instead of throwing exception? 

I think we are fine to upload without the filter component (after deleting the likely corrupted filter file). Server will rebuild the filter. Throwing exception fails the spark task, which causes a more expensive retry, delaying the completion of the spark job.

yifan-c (commented):
looks good overall.

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
rustyrazorblade: +1 to throwing it away.  Better than having a corrupt one, it'll rebuild on it's own this way.

lukasz-antoniak: Done, error handling updated.

lukasz-antoniak:
@rustyrazorblade, @yifan-c, I think I solved all comments. Can you do the final review?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -227,6 +229,15 @@ public synchronized Map<Path, Digest> prepareSStablesToSend(@NotNull BulkWriterC
         };
         Set<Path> dataFilePaths = new HashSet<>();
         Map<Path, Digest> fileDigests = new HashMap<>();
+        // FIXME: CQLSSTableWriter may produce incomplete Filter.db file, rebuilding it manually (see CASSANALYTICS-167).
```
skoppu22: If I understand correctly, we are rebuilding complete bloom filter, instead of fixing the actual cause (the cause of empty bloom filter), because the later requires changes in both C* 4.0 and 5.0 and new release/backporting etc in Cassandra, am I right? Can we create a Jira to fix this in upcoming C* versions so we can avoid completely rebuilding for upcoming C* versions, and add corresponding Jira here in comments, so we don't forget by any chance?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -297,6 +311,18 @@ public synchronized void close(BulkWriterContext writerContext) throws IOExcepti
         validateSSTables(writerContext);
     }
 
+    protected void rebuildFilterComponents(@NotNull BulkWriterContext writerContext, @NotNull Path outputDirectory,
+                                           @NotNull DirectoryStream.Filter<Path> filter) throws IOException
+    {
+        LocalDataLayer layer = buildLocalDataLayer(writerContext, outputDirectory, null);
```
skoppu22: [Minot] LocalDataLayer class's description says 'Mostly used for testing', but we are using for prod. We need to update this comment.

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
```diff
@@ -348,6 +352,64 @@ public org.apache.cassandra.bridge.BloomFilter openBloomFilter(Partitioner parti
...
+                                   @NotNull SSTable ssTable,
+                                   @NotNull Path directory) throws IOException
+    {
+        String keyspace = cqltable.keyspace();
+        String table = cqltable.table();
+        IPartitioner iPartitioner = getPartitioner(partitioner);
+        SchemaBuilder schemaBuilder = new SchemaBuilder(cqltable, partitioner);
+        TableMetadata tableMetadata = schemaBuilder.tableMetaData();
+
+        if (tableMetadata.params.bloomFilterFpChance == 1.0)
```
skoppu22, line 367: I guess here == comparison on doubles is safe as 1.0 doesn't involve rounding

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -227,6 +229,15 @@ public synchronized Map<Path, Digest> prepareSStablesToSend(@NotNull BulkWriterC
         };
         Set<Path> dataFilePaths = new HashSet<>();
         Map<Path, Digest> fileDigests = new HashMap<>();
+        // FIXME: CQLSSTableWriter may produce incomplete Filter.db file, rebuilding it manually (see CASSANALYTICS-167).
+        // rebuild Filter.db files before calculating their digest
+        try (DirectoryStream<Path> stream = getDataFileStream(sstableFilter))
+        {
+            for (Path path : stream)
```
skoppu22: Can we just pass sstableFilter to rebuildFilterComponents as it is already calling getDataFileStream and looping through it?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
lukasz-antoniak: Created [CASSANDRA-21423](https://issues.apache.org/jira/browse/CASSANDRA-21423).

skoppu22: Thanks for creating the Jira, can you please add CASSANDRA-21423 to the FIXME comments above and below

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -297,6 +305,18 @@ public synchronized void close(BulkWriterContext writerContext) throws IOExcepti
         validateSSTables(writerContext);
     }
 
+    protected void rebuildFilterComponents(@NotNull BulkWriterContext writerContext,
+                                           @NotNull DirectoryStream.Filter<Path> filter) throws IOException
+    {
+        LocalDataLayer layer = buildLocalDataLayer(writerContext, getOutDir(), null);
+        for (Path dataFile : getDataFileStream(filter))
```
skoppu22: Should we move getDataFileStream inside try-with-resources to autoclose? Same call above at line 292

lukasz-antoniak: Fixed both. Good catch.

skoppu22 (approved):

PR #213, opened 2026-05-30
CASSANALYTICS-171: Avoid Spark 4 partitioning warnings during reads

liucao-dd:
## Summary
- Report Spark 4 Cassandra scan partitioning as Spark's public `UnknownPartitioning` directly, instead of a custom `Partitioning` subclass.
- Preserve the input partition count when Spark asks `CassandraScanBuilder` for output partitioning.
- Add a Spark 4 unit test covering the reported partitioning contract.

## Why `UnknownPartitioning`
Spark 4's DataSource V2 partitioning contract expects connectors to report one of Spark's public partitioning types, such as `UnknownPartitioning` or `KeyGroupedPartitioning`, rather than implementing `Partitioning` directly. When Spark sees a custom `Partitioning`, `V2ScanPartitioningAndOrdering` ignores it and logs a warning.

Cassandra analytics input partitions are token ranges. Rows in one token range can contain many distinct Cassandra partition keys and many distinct token values, so they do not satisfy `KeyGroupedPartitioning`'s contract that every row in a Spark partition evaluates to the same partition value. Reporting `UnknownPartitioning` preserves the correct partition count without claiming keyed grouping semantics.

Jira: https://issues.apache.org/jira/browse/CASSANALYTICS-171

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 0.5.0
 -----
+ * Avoid Spark 4 partitioning warnings during bulk reads (CASSANALYTICS-171)
  * Spark 4.0 Support (CASSANALYTICS-34)
  * Add IAM credential support for S3 storage transport (CASSANALYTICS-155)
  * Make BulkWriterConfig extensible (CASSANALYTICS-168)
--- a/cassandra-analytics-core/src/main/spark4/org/apache/cassandra/spark/sparksql/CassandraScanBuilder.java
+++ b/cassandra-analytics-core/src/main/spark4/org/apache/cassandra/spark/sparksql/CassandraScanBuilder.java
@@ -43,6 +43,7 @@ import org.apache.spark.sql.connector.read.SupportsPushDownFilters;
 import org.apache.spark.sql.connector.read.SupportsPushDownRequiredColumns;
 import org.apache.spark.sql.connector.read.SupportsReportPartitioning;
 import org.apache.spark.sql.connector.read.partitioning.Partitioning;
+import org.apache.spark.sql.connector.read.partitioning.UnknownPartitioning;
 import org.apache.spark.sql.sources.Filter;
 import org.apache.spark.sql.types.StructType;
 import org.apache.spark.sql.util.CaseInsensitiveStringMap;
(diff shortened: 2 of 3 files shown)


cassandra-analytics-core/src/main/spark4/org/apache/cassandra/spark/sparksql/CassandraPartitioning.java
```diff
@@ -19,21 +19,16 @@
...
-import org.apache.cassandra.spark.data.DataLayer;
-import org.apache.spark.sql.connector.read.partitioning.Partitioning;
+import org.apache.spark.sql.connector.read.partitioning.UnknownPartitioning;
 
-class CassandraPartitioning implements Partitioning
+/**
+ * Cassandra input partitions are token ranges, not groups of rows sharing one partition key value,
+ * so the connector reports Spark's unknown partitioning rather than keyed-group partitioning.
+ */
+class CassandraPartitioning extends UnknownPartitioning
```
frankgh: do we even need the `CassandraPartitioning`? Can we directly instantiate an `UnknownPartitioning` instance from `CassandraScanBuilder`?

frankgh (commented):
Good find! I just have one small question

cassandra-analytics-core/src/main/spark4/org/apache/cassandra/spark/sparksql/CassandraPartitioning.java
liucao-dd: Good point. I checked Spark's `Partitioning` contract and several maintained Spark DSV2 connectors (Iceberg, Paimon, Lance). The common pattern is to instantiate `UnknownPartitioning` directly when the scan cannot guarantee keyed grouping, and reserve `KeyGroupedPartitioning` for cases where the connector can prove rows are grouped by the reported key expressions. Updated the patch to use `UnknownPartitioning` directly.

frankgh (approved):
+1

liucao-dd:
https://github.com/apache/cassandra-analytics/actions/runs/26692748197/job/78896333634?pr=213 this test timed out somehow. The patch didn't touch spark 3 related path though. Maybe it needs a retry?

sarankk (approved):
+1 Thanks, Lgtm

PR #214, opened 2026-06-03
CASSANALYTICS-160: Determine whether mutation tracking is enabled for keyspace for bulk writes

sarankk:


cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/CqlUtils.java
```diff
@@ -296,4 +299,33 @@ public static boolean isTimeRangeFilterSupported(String compactionStrategy)
     {
         return compactionStrategy == null || compactionStrategy.endsWith("TimeWindowCompactionStrategy");
     }
+
+    /**
+     * Extracts replication type from create schema statement
+     *
+     * @param schemaStr full cluster schema string as returned by Sidecar
+     * @param keyspace  name of the keyspace to check
+     * @return {@code true} if keyspace is tracked {@code false} otherwise
```
skoppu22, line 308: This is stale, extractReplicationType returning a string

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraDirectDataTransportContext.java
```diff
@@ -46,13 +46,24 @@ public CassandraDirectDataTransportContext(@NotNull BulkWriterContext bulkWriter
     }
 
     @Override
-    public DirectStreamSession createStreamSession(BulkWriterContext writerContext,
-                                                   String sessionId,
```
skoppu22, line 50: This indentation is better readable than the new one

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraDirectDataTransportContext.java
```diff
@@ -46,13 +46,24 @@ public CassandraDirectDataTransportContext(@NotNull BulkWriterContext bulkWriter
...
+            BulkWriterContext writerContext,
+            String sessionId,
+            SortedSSTableWriter sstableWriter,
+            Range<BigInteger> range,
+            ReplicaAwareFailureHandler<RingInstance> failureHandler,
+            ExecutorService executorService)
     {
+        if (bridge().isTracked(clusterInfo.getReplicationType()))
+        {
+            return new TrackedDirectStreamSession(writerContext,
```
skoppu22, line 59: As we plan to implement the feature in multiple PRs and phases, can we have a feature flag, set to disabled by default, so the existing code doesn't get impacted by this feature changes by any chance. And developers can turn it on when they like to try it out. CI can enable the feature flag to test the code. We can set it enabled by default when we reach a comfortable stage.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TrackedDirectStreamSession.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Stream session for bulk writes to keyspaces with mutation tracking enabled.
+ *
+ * <p>
+ * Tracked stream session uploads and triggers import on the <em>coordinator node only</em>. Cassandra's coordinated
+ * transfer then propagates the data to all other replicas, avoiding the duplicate-row updates that would occur if each
+ * replica independently streamed the data to its peers.
+ * <p>
+ */
+public class TrackedDirectStreamSession extends StreamSession<TransportContext.DirectDataBulkWriterContext>
```
skoppu22: Don't we need to implement similar one for CloudStorageStreamSession as well? In non-coordinated mode, it calls tryCreateRestoreSlicePerReplica. My understanding is, if the destination Cassandra keyspace is tracked, we will end up writing to every replica instead of writing to coordinator only. 

Incase of coordinated cloud storage mode, though analytics writes to one sidecar, we need to verify does Sidecar/Cassandra restore code is handling tracked keyspaces correctly or not.

skoppu22: We don't have to implement everything in this PR. But we may need to create similar class for these cases and follow up JIRAs not to miss them.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/utils/CqlUtilsTest.java
```diff
@@ -833,6 +833,43 @@ public void testTimeRangeFilterSupported()
...
 
+    @Test
+    public void testExtractReplicationType()
+    {
+        String schemaStrTracked
+        = "CREATE KEYSPACE k WITH REPLICATION "
+          + "= { 'class' : 'org.apache.cassandra.locator.NetworkTopologyStrategy', 'dc1': '3' }"
+          + " AND DURABLE_WRITES = true AND replication_type = 'tracked';";
+        assertThat(CqlUtils.extractReplicationType(schemaStrTracked, "k")).isEqualTo("tracked");
+        String schemaStrTrackedCaseInsensitive
```
skoppu22, line 844: For better readability, can you please either create separate tests for these or separate them with new lines.

CHANGES.txt
```diff
@@ -1,5 +1,6 @@
 0.5.0
 -----
+ * Analytics should identify if a keyspace is tracked to determine appropriate stream session to use for bulk writes (CASSANALYTICS-160)
```
skoppu22: [Minor] Prefer shorter and simpler statement

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -356,6 +356,17 @@ public ReplicationFactor replicationFactor()
         }
     }
 
+    @Override
+    public String getReplicationType()
```
skoppu22, line 360: CassandraClusterInfoGroup not overriding this, i.e, doesn't detect tracked keyspaces

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfoTest.java
```diff
@@ -75,6 +77,50 @@ public static CassandraClusterInfo mockClusterInfoForTimeSkewTest(int allowanceM
...
 
+    @Test
+    void testGetTrackedReplicationType()
+    {
+        String schema = "CREATE KEYSPACE mykeyspace "
+                        + "WITH replication = {'class': 'NetworkTopologyStrategy', 'dc1': '3'}"
+                        + "  AND durable_writes = true"
+                        + "  AND replication_type = 'tracked';";
+        CassandraClusterInfo ci = mockClusterInfoWithSchema("mykeyspace", schema);
+        assertThat(ci.getReplicationType())
```
skoppu22, line 88: These tests are not testing changes to CassandraClusterInfo, because MockClusterInfoForSchema overriding getReplicationType to call CqlUtils.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TrackedDirectStreamSession.java
sarankk: Yes we will create follow up JIRAs for S3 mode work

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraDirectDataTransportContext.java
sarankk: The tracked stream session should not be used for existing use cases since keyspace would not be tracked and we cannot use direct stream session when keyspace is tracked. Hence feature flag might not be necessary here

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraDirectDataTransportContext.java
```diff
@@ -71,7 +82,11 @@ public DirectDataTransferApi dataTransferApi()
     // only invoke in constructor
     protected DirectDataTransferApi createDirectDataTransferApi()
     {
-        CassandraBridge bridge = CassandraBridgeFactory.get(clusterInfo.getLowestCassandraVersion());
-        return new SidecarDataTransferApi(clusterInfo.getCassandraContext(), bridge, jobInfo);
+        return new SidecarDataTransferApi(clusterInfo.getCassandraContext(), bridge(), jobInfo);
+    }
+
+    private CassandraBridge bridge()
```
skoppu22, line 88: writerContext already has bridge() method. can you please reuse that

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfoTest.java
sarankk: Yes otherwise we need to mock sidecar client calls to get schema, avoiding that

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraDirectDataTransportContext.java
skoppu22: And we need to throw unsupported here instead of waiting until TrackedDirectStreamSession methods are called

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/CqlUtils.java
```diff
@@ -296,4 +299,33 @@ public static boolean isTimeRangeFilterSupported(String compactionStrategy)
...
+
+    /**
+     * Returns {@code true} if {@code replication_type = 'tracked'} in create statement otherwise {@code false}
+     *
+     * @param replicationType replication type extracted from create statement
+     * @return {@code true} if replication type is tracked {@code false} otherwise
+     */
+    public static boolean isTracked(String replicationType)
+    {
+        return TRACKED_REPLICATION_TYPE.equalsIgnoreCase(replicationType);
```
skoppu22, line 329: Here we are checking for replication_type value only. But the cluster need to have mutation_tracking.enabled to true for the feature to work at Cassandra side. Can you please explore what happens in Cassandra if a keyspace was created with replication_type = tracked, but then later admins changed mutation_tracking.enabled to false? Based on these findings we may need to check for mutation_tracking.enabled as well.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraDirectDataTransportContext.java
skoppu22: For any new feature, we should either develop them in a feature branch or disable it under a feature flag until it is complete and stable. For example, mutation tracking feature in C* is in a feature branch. 
Especially in this PR, we are introducing an incomplete feature to the upstream. So having a feature flag is very important to ensure existing use cases are not broken by any chance with our new code inadvertently.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfoTest.java
skoppu22: Can we have in-jvm tests to test CassandraClusterInfo code path then? Tests in CassandraClusterInfoTest.java‎ should test CassandraClusterInfo. If it is difficult, we can have in-jvm tests which go through this class.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraDirectDataTransportContext.java
sarankk: Thanks, moving the PR against a feature branch instead

CHANGES.txt
sarankk: Will keep this in mind for next time, since JIRA is already created, will use JIRA title for easier reference

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
sarankk: `CassandraClusterInfoGroup` is used for bulk writes through S3, for now we will not support write to tracked tables via S3. We will handle S3 code path changes in a separate JIRA

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfoTest.java
sarankk: @skoppu22 I plan to add in-jvm dtest in the follow up PR which adds the new stream session for write

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/CqlUtils.java
sarankk: Good catch Shailaja! thanks, looks like we need to check the cluster level flag as well `replication_type` = tracked does not guarantee mutation tracking is enabled at cluster level

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraDirectDataTransportContext.java
skoppu22: [minor] Code still has a new bridge() function. Please use writerContext's bridge function.

skoppu22 (approved):
Merge to feature branch is fine

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/CqlUtils.java
sarankk: @skoppu22 I looked into this more, in Cassandra side during imports through coordinated transfer Cassandra is only checking replication type so far, it is not checking the cluster level flag. If replication type is tracked, regardless of whether cluster level flag is enabled or disabled Cassandra takes the coordinated transfer route, we should be good to use the `TrackedStreamSession`.

The scenario of what if the cluster level mutation tracking flag is turned off after creating a tracked keyspace, needs further discussion and we can leave that out of this PR

skoppu22: OK

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfoTest.java
skoppu22: Fine, as we are merging to feature branch only

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
skoppu22: OK

CHANGES.txt
yifan-c: You can update the JIRA title.

yifan-c (approved):

PR #217, opened 2026-06-11
CASSANALYTICS-176 Upgrade sidecar version to 0.4.0

skoppu22:
Upgrading to latest Sidecar version. This unblocks C* 5.0 support features work in this repo.

Circle CI : https://app.circleci.com/pipelines/github/skoppu22/cassandra-analytics/138/workflows/dbad712d-e920-4e0f-ba01-dddd8bb2f488


build.gradle
```diff
@@ -223,6 +223,10 @@ subprojects {
 
   repositories {
     mavenCentral()
+    // Confluent repository for kafka-avro-serializer (transitive dep of sidecar-server)
```
frankgh: Can we instead exclude the transitive dependency since it is not required for Analytics to function ?

skoppu22: Integration tests are failing without this

frankgh: but maybe we should just exclude the dependency from the sidecar dependency?

skoppu22: Excluded. Circle CI passed after this https://app.circleci.com/pipelines/github/skoppu22/cassandra-analytics/140/workflows/86d76ed2-f1ab-4e4d-8756-3c0d69ff327e

frankgh (approved):
+1

jyothsnakonisa (approved):

PR #219, opened 2026-06-15
CASSANDRA-21464 trunk Add security-model discoverability

smiklosovic:


potiuk (approved):

smiklosovic:
@frankgh approved similar patch in Sidecar so I will just proceed.

PR #220, opened 2026-06-17
CASSANALYTICS-31 SAI index support in analytics

skoppu22:
Generate SAI index files as well while producing sstable files, so no need of async rebuilding of SAI indexes after bulk write.

Circle CI : https://app.circleci.com/pipelines/github/skoppu22/cassandra-analytics/148/workflows/5eaeec43-e43b-444c-badb-ef696a3828fc


cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/common/SSTables.java
jyothsnakonisa, line 1: Could you please add test for new methods that are being added specially for SAI file names?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/common/SSTables.java
```diff
@@ -22,14 +22,52 @@
...
+    /**
+     * Determine whether the given path is a primary SSTable data component ("&lt;descriptor&gt;-Data.db").
+     *
+     * @param path file path
+     * @return true if the path's file name is a primary data component
+     * @see #isDataComponent(String)
+     */
+    public static boolean isDataComponent(Path path)
+    {
+        return isDataComponent(path.getFileName().toString());
```
jyothsnakonisa: Should we do a null check?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/CqlUtils.java
```diff
@@ -59,6 +60,9 @@ public final class CqlUtils
     private static final Pattern ESCAPED_DOUBLE_BACKSLASH = Pattern.compile("\\\\");
     private static final Pattern COMPACTION_STRATEGY_PATTERN = Pattern.compile("compaction\\s*=\\s*\\{\\s*'class'\\s*:\\s*'([^']+)'");
 
+    private static final Pattern MULTI_WHITESPACE_PATTERN = Pattern.compile("\\s+");
+    private static final Pattern SAI_USING_PATTERN = Pattern.compile("USING '[^']*STORAGEATTACHEDINDEX'");
```
jyothsnakonisa: This pattern matches with something like "USING 'PrefixStorageAttachedIndex'" which is incorrect, how about using something like "Pattern.compile("USING '([^']*\\.)?STORAGEATTACHEDINDEX'")"

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
```diff
@@ -316,6 +349,45 @@ static void validateNoSecondaryIndexes(TableInfoProvider tableInfo)
         }
     }
 
+    public Set<String> getIndexStatements()
+    {
+        return indexStatements;
+    }
+
+    @VisibleForTesting
+    static boolean isCassandra5OrLater(String version)
```
jyothsnakonisa: Can u check the correctness of this method for a version like 3.11.x? May be worth adding tests for some of the methods you added here in this class

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
```diff
@@ -308,6 +300,47 @@ private static void validateDataframeFieldsInTable(TableInfoProvider tableInfo,
...
+     * SAI index components are generated alongside SSTables and are immediately queryable after import.
+     * <p>
+     * When any index is non-SAI (legacy 2i), the write is blocked unless SKIP_SECONDARY_INDEX_CHECK is set.
+     *
+     * @param tableInfo               the table info provider
+     * @param skipSecondaryIndexCheck  whether the user explicitly opted out of the check
+     * @param indexStatements          the CREATE INDEX statements for the table
+     * @param lowestCassandraVersion   the lowest Cassandra version in the cluster
+     */
+    static void validateSecondaryIndexes(TableInfoProvider tableInfo,
```
jyothsnakonisa: Can u add tests for this method?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
```diff
@@ -308,6 +300,47 @@ private static void validateDataframeFieldsInTable(TableInfoProvider tableInfo,
...
+
+        if (skipSecondaryIndexCheck)
+        {
+            LOGGER.warn("Bulk writing to tables with SecondaryIndexes will have an asynchronous index rebuild "
+                      + "take place automatically after writing. Reads against the index during this time "
+                      + "window will produce inconsistent or stale results until index rebuild is complete.");
+            return;
+        }
+
+        throw new UnsupportedAnalyticsOperationException("Bulkwriter doesn't support secondary indexes");
```
jyothsnakonisa: How about changing exception message to ""Bulkwriter doesn't support non-SAI  indexes."

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
jyothsnakonisa: Is `TableInfoProvider tableInfo,` parameter needed? can we may be do 

`if (indexStatements.isEmpty()) ` instead of `if (!tableInfo.hasSecondaryIndex())`

cassandra-five-zero-types/src/main/java/org/apache/cassandra/spark/reader/StorageAttachedIndexApplier.java
jyothsnakonisa, line 1: Since this is making important change to the schema, could you please add tests for this class?

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
```diff
@@ -281,10 +282,17 @@ public CqlTable buildSchema(String createStatement,
...
     {
-        return new SchemaBuilder(createStatement, keyspace, replicationFactor, partitioner, cassandraTypes -> udts, tableId, indexCount, enableCdc).build();
+        CqlTable cqlTable = new SchemaBuilder(createStatement, keyspace, replicationFactor, partitioner,
+                                              cassandraTypes -> udts, tableId, indexStatements, enableCdc).build();
+        // SAI is a Cassandra 5.0+ feature: register any Storage Attached Index definitions on the table metadata
+        // after the shared SchemaBuilder has built/registered the (index-less) table.
+        // buildSchema is invoked repeatedly within a JVM (the index-carrying per-partition RecordWriter included),
+        // so this re-applies the SAI definitions whenever a rebuild left the table without them.
+        StorageAttachedIndexApplier.maybeApply(keyspace, cqlTable.table(), indexStatements);
+        return cqlTable;
```
jyothsnakonisa, line 289: Schema Is created first and then SAI indexes are attached in the next step, I wondering about the brief moment between two statements where the schema doesn't have SAI indexes.

WDYT about merging both operations into a single CassandraSchema.apply in the 5.0 bridge? you might have to override CassandraSchema.apply call entirely from the 5.0 bridge and don't use SchemaBuilder's constructor at all for the schema update.

I am open to other approaches too

cassandra-four-zero-types/src/main/java/org/apache/cassandra/spark/reader/SchemaBuilder.java
```diff
@@ -222,6 +224,20 @@ private static Pair<KeyspaceMetadata, TableMetadata> updateSchema(Schema schema,
...
+        // the registered table.
+        if (maybeExistingTableMetadata != null
+            && !maybeExistingTableMetadata.indexes.isEmpty()
+            && tableMetadata.indexes.isEmpty())
+        {
+            tableMetadata = tableMetadata.unbuild()
+                                         .indexes(maybeExistingTableMetadata.indexes)
+                                         .build();
+        }
+
```
jyothsnakonisa: RecordWriter is using 5 arg buildSchema() method which is not attaching index statements, you could eliminate copying indexes here if you use the 8 arg buildSchema() method in RecordWriter and can eventually get rid of 5 arg buildSchema() method

 https://github.com/apache/cassandra-analytics/blob/0eb3a7f85dbef0500e947f4a8d9a1ae3682bdc60/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java#L119

```
cqlTable = writerContext.bridge()
                          .buildSchema(writerContext.schema().getTableSchema().createStatement,
                                       writerContext.job().qualifiedTableName().keyspace(),
                                       IGNORED_REPLICATION_FACTOR,
                                       writerContext.cluster().getPartitioner(),
                                       writerContext.schema().getUserDefinedTypeStatements(),
                                       null,
                                       writerContext.schema().getTableSchema().getIndexStatements(),
                                       false);
```

jyothsnakonisa (commented):
Some comments

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/common/SSTables.java
skoppu22: Ideally shouldn't happen for sstable files, but added null check for extra safety

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/CqlUtils.java
skoppu22: Done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
skoppu22: Done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
skoppu22: Done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
skoppu22: Done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TableSchema.java
skoppu22: Done

cassandra-five-zero-types/src/main/java/org/apache/cassandra/spark/reader/StorageAttachedIndexApplier.java
skoppu22: Done

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
skoppu22: Done using a afterRegister function

cassandra-four-zero-types/src/main/java/org/apache/cassandra/spark/reader/SchemaBuilder.java
skoppu22: I did restructure and moved SAI specific code to extended class. Now this comment not applicable I believe

jyothsnakonisa: @skoppu22  This block of code is moved to `FiveZeroSchemaBuilder.beforeTableRegistered` but the issue still exists. You can get rid of the logic to not drop indexes in beforeTableRegistered and even the method `beforeTableRegistered` if you use 8 argument constructor for schemaBuilder in Recordwriter passing ` writerContext.schema().getTableSchema().getIndexStatements()` in the constructor.

skoppu22: @jyothsnakonisa I have applied your recommended changes to pass index statements as part of cqlTable. We still need beforeTableRegistered. Because

Reason 1, Some builders only receive a CREATE TABLE string, with no CqlTable and no CREATE INDEX statements, so they can only build index‑less. These are public CassandraBridge APIs:

```
encodePartitionKeys(...) / toTokens(...) / encodePartitionKey(...) → new FiveZeroSchemaBuilder(createTableStmt, ks, rf, partitioner)

readPartitionKeys(...) → same createStmt‑only builder

```

We can't make these pass statements without changing the public bridge API, and their callers frequently don't have the CREATE INDEX statements to pass. If one of these is the first build of a table in a JVM, the table registers without its SAI index.

Reason 2, The same table is built more than once per JVM in normal flows:

```
Write path, multiple partitions/executor: RecordWriter is constructed per Spark partition (mapPartitions); the 2nd+ partition on an executor rebuilds the already‑registered table.

Write path, single partition: after writing, SortedSSTableWriter.validateSSTables(...) → buildLocalDataLayer → new LocalDataLayer → buildSchema rebuilds the same table in the same JVM.

Read path: getCompactionScanner, getPartitionSizeIterator, rebuildBloomFilter all build from a CqlTable, repeatedly within a JVM.
```

skoppu22: Looks like there are no callers for encodePartitionKeys and readPartitionKeys. If we can remove them, then we can switch to 8 arg constructor as you mentioned, and we can simplify beforeTableRegistered just to avoid duplicate registration

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/KryoSerializationTests.java
```diff
@@ -261,7 +261,9 @@ public void testCqlTable(CassandraBridge bridge)
                                       "create table test_keyspace.test_table"
                                       + " (a bigint, b bigint, c bigint, d bigint, e bigint, primary key((a, b), c));",
                                       replicationFactor,
-                                      fields);
+                                      fields,
+                                      java.util.Collections.emptySet(),
+                                      java.util.Collections.emptySet());
```
jyothsnakonisa: Please remove fully qualified class names.

cassandra-four-zero-types/src/main/java/org/apache/cassandra/spark/reader/SchemaBuilder.java
jyothsnakonisa: Yes encodePartitionKeys & readPartitionKeys are not used anywhere in production code hence removing them should be safe. I will leave it to you whether to remove those methods or not.

jyothsnakonisa (approved):
LGTM

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
```diff
@@ -86,11 +85,12 @@ protected void onSSTablesProduced(Set<SSTableDescriptor> sstables)
...
                 // 4. remove the sstables once sent
                 Map<Path, Digest> fileDigests = sstableWriter.prepareSStablesToSend(writerContext, sstables);
-                // retain only the SSTable data components
+                // retain only the primary SSTable data components ("<descriptor>-Data.db"); SAI per-index
+                // components such as "...+TermsData.db" (which also end with "Data.db") are excluded
                 IntWrapper sstableCounter = new IntWrapper();
                 fileDigests.keySet()
                            .stream()
-                           .filter(p -> p.getFileName().toString().endsWith(FileType.DATA.getFileSuffix()))
+                           .filter(SSTables::isDataComponent)
```
yifan-c, line 93: 👍

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
```diff
@@ -224,8 +225,9 @@ private void sendSSTableToOneReplica(Path dataFile,
         {
             for (Path componentFile : componentFileStream)
             {
-                // send data component the last
-                if (componentFile.getFileName().toString().endsWith("Data.db"))
+                // send the primary data component ("<descriptor>-Data.db") last; SAI per-index components
+                // such as "...+TermsData.db" are still streamed here rather than skipped
+                if (SSTables.isDataComponent(componentFile))
```
yifan-c, line 230: nit: Can you revert the comment?
This comment is already there in the javadoc of the method. 
Repeating here and several other call-sites creates noise. The call-sites only need to know the method can determine whether a file is data component or not.

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/CqlUtils.java
```diff
@@ -59,6 +60,9 @@ public final class CqlUtils
     private static final Pattern ESCAPED_DOUBLE_BACKSLASH = Pattern.compile("\\\\");
     private static final Pattern COMPACTION_STRATEGY_PATTERN = Pattern.compile("compaction\\s*=\\s*\\{\\s*'class'\\s*:\\s*'([^']+)'");
 
+    private static final Pattern MULTI_WHITESPACE_PATTERN = Pattern.compile("\\s+");
+    private static final Pattern SAI_USING_PATTERN = Pattern.compile("USING '([^']*\\.)?STORAGEATTACHEDINDEX'");
```
yifan-c: The pattern is not comprehensive. See the example below. cc: @maedhroz 

```
cqlsh> DESC TABLE ks.tstsai;

CREATE TABLE ks.tstsai (
    id int PRIMARY KEY,
    val1 int,
    val2 int,
    val3 int,
    val4 int,
    val5 int
);

CREATE CUSTOM INDEX val1_index ON ks.tstsai (val1) USING 'STORAGEATTACHEDINDEX';

CREATE CUSTOM INDEX val2_index ON ks.tstsai (val2) USING 'storageattachedindex';

CREATE CUSTOM INDEX val3_index ON ks.tstsai (val3) USING 'sai';

CREATE CUSTOM INDEX val4_index ON ks.tstsai (val4) USING 'StorageAttachedIndex';

CREATE CUSTOM INDEX val5_index ON ks.tstsai (val5) USING 'org.apache.cassandra.index.sai.StorageAttachedIndex';
```

yifan-c (changes requested):
partial review; spotted a change that is needed. submitting the comments so far.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BroadcastableTableSchema.java
```diff
@@ -47,7 +48,7 @@
  */
 public final class BroadcastableTableSchema implements Serializable
 {
-    private static final long serialVersionUID = 1L;
+    private static final long serialVersionUID = 2L;
```
lukasz-antoniak: I understand the rational, but is it really needed? In the past we have made changes in class fields without updating serial version ID.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CqlTableInfoProvider.java
```diff
@@ -157,7 +158,13 @@ public String getKeyspaceName()
     @Override
     public boolean hasSecondaryIndex()
```
lukasz-antoniak: I would suggest to update the method name, since it is not only about 2i indexes any more.

jmckenzie-dev: This is a somewhat nasty conflation. SAI is _technically_ a secondary index, we just always used that term too broadly to mean something very specific (and old...) in the project.

Short of us refactoring a bunch of old code just to rename something, I think your point stands though. Better to have this be `public boolean hasIndexes()` or something since it's not immediately clear from this method name if we're referring to the inappropriately broadly named legacy SecondaryIndex impl or the general concept of Secondary Indexes.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
```diff
@@ -153,7 +153,8 @@ protected StreamResult doFinalizeStream()
     @Override
     protected void sendRemainingSSTables()
     {
-        try (DirectoryStream<Path> dataFileStream = Files.newDirectoryStream(sstableWriter.getOutDir(), "*Data.db"))
+        try (DirectoryStream<Path> dataFileStream = Files.newDirectoryStream(sstableWriter.getOutDir(),
+                                                                             SSTables.DATA_COMPONENT_GLOB))
```
jmckenzie-dev, line 157: Strictly speaking this is a change in behavior. One I think is correct, but worth calling out so we can think Extra Hard.

Before we matched `*Data.db`, and now we're matching `*-Data.db`. Given we can have files that match the former in SAI incorrectly, I'm confident this is a correct change. I don't _think_ it's worth adding a comment here to discuss how closely the two file names will potentially be to one another (`+Data.db` vs. `-Data.db`). I think the upstream contract in the core DB is pretty firm so I don't see this drifting, and I don't think there's anything to change here. Just calling it out.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
jmckenzie-dev: This comment actually confused me a bit. It wasn't clear what "here" meant; I read it as "are still streamed here" then saw the immediate "if match: continue" logic which looks like the opposite of streaming here. :)

So if you drop the comment entirely, no worries. If you decide to keep it, I'd clarify something like "SAI per-index components such as "..." are still streamed below; we're only skipping core SSTable `-Data.db` files here."

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
```diff
@@ -116,12 +116,17 @@ public RecordWriter(BulkWriterConfig config, String[] columnNames)
...
         cqlTable = writerContext.bridge()
                                 .buildSchema(writerContext.schema().getTableSchema().createStatement,
                                              writerContext.job().qualifiedTableName().keyspace(),
                                              IGNORED_REPLICATION_FACTOR,
                                              writerContext.cluster().getPartitioner(),
-                                             writerContext.schema().getUserDefinedTypeStatements());
+                                             writerContext.schema().getUserDefinedTypeStatements(),
+                                             null,
+                                             writerContext.schema().getTableSchema().getIndexStatements(),
+                                             false);
```
jmckenzie-dev: It may help clarity here to comment on why a null UUID and cdc always false are the correct choices for args here. From this elevation I have no way of really knowing whether those 2 choices are correct.

Another option would be an extension to the bridge w/a method signature that excludes those 2 arguments, thus making it a responsibility of the bridge and `buildSchema` side of the call, which would make reviewing this at this point a lot easier to have confidence in the correctness of. Right now it's a method signature w/some internals effectively leaking where we need to know if this null / false combination is correct here. Which I unfortunately do not. 😀

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SidecarDataTransferApi.java
```diff
@@ -107,6 +110,14 @@ public RemoteCommitResult commitSSTables(CassandraInstance instance,
         // Always verify SSTables on import
         importOptions.verifySSTables(true).extendedVerify(!job.skipExtendedVerify());
 
+        // When SAI index components were generated alongside SSTables, enable SAI validation on import.
+        // failOnMissingIndex makes the import fail (instead of silently rebuilding) when the uploaded
+        // SSTables are missing their SAI components, and validateIndexChecksum verifies their checksums.
+        if (hasSaiIndexes)
+        {
+            importOptions.failOnMissingIndex(true).validateIndexChecksum(true);
```
jmckenzie-dev, line 118: Is there any world where someone would want the old behavior of generating the SAI indexes on-node? I can't think of any reason to still _ever_ support that, but hard-coding these checks and this flow throughout this patch is definitely communicating "you will do things the new way and you will like it".

Which, for the record, I'm ok with. Again: can't think of why someone would ever want the old lazy stale index way, but just to check - did we consider this and decide on this path?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -387,11 +389,10 @@ private LocalDataLayer buildLocalDataLayer(@NotNull BulkWriterContext writerCont
 
     private DirectoryStream<Path> getDataFileStream(DirectoryStream.Filter<Path> filter) throws IOException
     {
-        // Combine the data file filter with the provided filter
-        DirectoryStream.Filter<Path> combinedFilter = path -> {
-            String fileName = path.getFileName().toString();
-            return fileName.endsWith("Data.db") && filter.accept(path);
-        };
+        // Combine the data file filter with the provided filter.
+        // Match only the primary SSTable data component ("<descriptor>-Data.db"); the leading '-' excludes
```
jmckenzie-dev: I'd comment `SSTables.isDataComponent` with this context, remove this comment here and in all the places it's proliferated in this patch, and just call the method here. This leaves implementation details and notes on how that disambiguation and filtering occurs in the class that owns that behavior. This is implementation details leaking out of the called method which could go stale in the future if we change or augment the logic in `isDataComponent(String n)`

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriterTest.java
```diff
@@ -176,7 +176,10 @@ public void testBloomFilterRebuild(String version) throws IOException
                                                new ReplicationFactor(ReplicationFactor.ReplicationStrategy.SimpleStrategy,
                                                                      ImmutableMap.of("replication_factor", 1)),
                                                partitioner,
-                                               Collections.emptySet());
+                                               Collections.emptySet(),
+                                               null,
```
jmckenzie-dev: Hm. Seeing this in a few spots (null UUID + false cdc). See earlier comment re: new method in `CassandraBridge`.

jmckenzie-dev (commented):
Most of the way through review; will try and knock out the rest of it tomorrow.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BroadcastableTableSchema.java
yifan-c: The serialVersionUID change is unnecessary. 
The serialization is not used in any long running service and no mixed-mode is expected.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SidecarDataTransferApi.java
skoppu22: - I do not see a usecase where user doesn't want to generate SAI indexes for the table being written
- If we come across any perf or other issues, we can introduce a feature flag to disable this feature in analytics. failOnMissingIndex and validateIndexChecksum are only to tell Cassandra to verify index components and fail if mismatch. withIndexes() method of CqlSSTableWriter is the actual method asking Cassandra to generate SAI components.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
skoppu22: I prefer to not have a constructor without requiring index statements. Someone changin code in the future may miss passing index statements and analytics will miss generating SAI components in such cases. 

I will write a comment instead.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriterTest.java
skoppu22: Same as above, I would like to avoid a constructor without index statements as a param, so someone may miss passing them

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/KryoSerializationTests.java
skoppu22: Done

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BroadcastableTableSchema.java
skoppu22: Reverted

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CqlTableInfoProvider.java
skoppu22: Renamed to hasIndexes

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
skoppu22: Reworded as suggested by Josh

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
skoppu22: Removed the comment

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
skoppu22: It shouldn't impact, as other than data files, only other sstable files ending with Data.db are SAI and these exists only from 5.0

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/CqlUtils.java
skoppu22: Done, also added tests for all these cases

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RecordWriter.java
skoppu22: Added comment

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriterTest.java
skoppu22: Added comment

cassandra-four-zero-types/src/main/java/org/apache/cassandra/spark/reader/SchemaBuilder.java
skoppu22: Marked them deprecated

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriterTest.java
jmckenzie-dev: Can we have a variant that takes index statements but the user doesn't have to provide UUID and cdc, pushing that responsibility into the Bridge instead?

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteSAIIndexTest.java
```diff
@@ -0,0 +1,305 @@
...
+import static org.apache.cassandra.testing.TestUtils.ROW_COUNT;
+import static org.apache.cassandra.testing.TestUtils.TEST_KEYSPACE;
+
+/**
+ * Integration test for bulk write and read operations on tables with Storage Attached Indexes (SAI).
+ * Covers an all-SAI table and a mixed SAI + legacy 2i table.
+ */
+class BulkWriteSAIIndexTest extends SharedClusterSparkIntegrationTestBase
+{
+    private static final QualifiedName TABLE_SAI =
```
jmckenzie-dev: nit: no need for newline here and on the declaration of TABLE_MIXED

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteSAIIndexTest.java
```diff
@@ -0,0 +1,305 @@
...
+        SparkSession spark = getOrCreateSparkSession();
+        Dataset<Row> dfWrite = DataGenerationUtils.generateCourseData(spark, ROW_COUNT);
+
+        // 1. Bulk write to table with SAI indexes on both course and marks
+        bulkWriterDataFrameWriter(dfWrite, TABLE_SAI).save();
+
+        // 2. Flush to ensure SSTable components (including SAI) are written to disk
+        cluster.getFirstRunningInstance().flush(TEST_KEYSPACE);
+
+        // 3. Verify SAI index SSTable components exist on the filesystem
```
jmckenzie-dev: This comment is redundant w/the code right below it. Doesn't add anything new beyond what the code clearly communicates.

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteSAIIndexTest.java
```diff
@@ -0,0 +1,305 @@
...
+        assertThat(hasSaiIndexFiles())
+            .as("SAI index files should exist on disk after bulk write and flush")
+            .isTrue();
+
+        // 4. Bulk read the data back and verify equality
```
jmckenzie-dev: Same here; redundant comment.

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
```diff
@@ -281,10 +282,15 @@ public CqlTable buildSchema(String createStatement,
                                 Partitioner partitioner,
                                 Set<String> udts,
                                 @Nullable UUID tableId,
-                                int indexCount,
+                                Set<String> indexStatements,
                                 boolean enableCdc)
     {
-        return new SchemaBuilder(createStatement, keyspace, replicationFactor, partitioner, cassandraTypes -> udts, tableId, indexCount, enableCdc).build();
+        // FiveZeroSchemaBuilder attaches SAI definitions (and preserves them across index-less rebuilds) within the
```
jmckenzie-dev: How hard would it be for us to push this contract up to the root `SchemaBuilder` - that anything that has multiple components in a schema guarantees atomic visibility of all components? I _believe_ that's true for the base `SchemaBuilder` excluding legacy SecondaryIndex implementations, but other than that one exception, I think it could be a valuable contract to push that expectation to the root, javadoc it there on `SchemaBuilder` (w/the one exception on legacy 2i), and then we don't have to have comments scattered around at call sites outside the class and instead can have a mental model about the correctness and mutation guarantees of the class we're relying on.

This probably translates into no code changes and just comment changes; calling out the lack of atomicity on legacy 2i (hence the disabled by default) but otherwise communicating the guarantee of atomicity across all other components and their visibility. That'll help provide guidance for future developers as they maintain and extend these classes.

cassandra-five-zero-types/src/main/java/org/apache/cassandra/spark/reader/SchemaBuilder.java
jmckenzie-dev, line 1: Am I correct in understanding that `SchemaBuilder.java` now serves as `PreFiveZeroSchemaBuilder.java`? If so, we should javadoc this class at least to indicate that role it now plays and the object hierarchy split, if not go so far as to rename the class to something like "BasicSchemaBuilder.java" or something. Names are hard, so maybe just a javadoc comment indicating the role it now plays with a {@link} callout to `FiveZeroschemaBuilder` to help future maintainers know the context.

cassandra-five-zero-bridge/src/test/java/org/apache/cassandra/spark/reader/StorageAttachedIndexApplierTest.java
```diff
@@ -0,0 +1,217 @@
...
+    @Test
+    public void testApplyToIsIdempotent()
+    {
+        String keyspace = uniqueKeyspace("idempotent");
+        registerIndexlessTable(keyspace, "tbl");
+
+        Set<String> sai = ImmutableSet.of(
+                "CREATE CUSTOM INDEX tbl_b_idx ON " + keyspace + ".tbl (b) USING 'StorageAttachedIndex';");
+        applyIndexes(keyspace, "tbl", sai);
+        // Second call is a no-op because the registered table already carries indexes.
```
jmckenzie-dev, line 75: Why are we calling this a 2nd time? What are we trying to exercise here?

cassandra-five-zero-types/src/main/java/org/apache/cassandra/spark/reader/FiveZeroSchemaBuilder.java
```diff
@@ -0,0 +1,100 @@
...
+                                 boolean enableCdc)
+    {
+        super(createStmt, keyspace, replicationFactor, partitioner, udtStatementsProvider, tableId, indexStatements, enableCdc);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected TableMetadata beforeTableRegistered(TableMetadata tableMetadata, @Nullable TableMetadata previousTable)
```
jmckenzie-dev: The name `previousTable` - does this mean `beforeTable`? i.e. this method mutates whatever table you pass in in a "before and after" pattern right?

The name "previousTable" had me initially trying to figure out what the previous table was and what its relationship to this new table was. The word "previous" tends to refer to things in a chain of ordering with multiple entrants vs. a "before and after" which speaks to mutating one thing.

Which is a ton of words to say: I think we should rename this variable to `beforeTable` or `existingTableMetadata` or something. Ultimately it's just decorating the existing table metadata with the indices of the passed in metadata.

cassandra-five-zero-types/src/main/java/org/apache/cassandra/spark/reader/StorageAttachedIndexApplier.java
```diff
@@ -0,0 +1,116 @@
...
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(StorageAttachedIndexApplier.class);
+
+    private StorageAttachedIndexApplier()
+    {
+    }
+
+    /**
+     * Applies any SAI definitions in {@code indexStatements} to {@code table} within the supplied {@code schema}.
+     * Operates on a caller-provided schema so it runs inside an existing atomic schema update (the same update that
```
jmckenzie-dev: Is this comment here talking about the context of the calling methods? Because we should avoid doing this. Any subsequent changes to the outer context could render this obsolete and then we have a comment here making claims about some external contracts. While it's true and in context at the time of this patch, comments should speak to what they locally have ownership over (the why of things and non-obvious implementation details) without speaking about the context of external usage of a method.

jmckenzie-dev (changes requested):

cassandra-five-zero-bridge/src/test/java/org/apache/cassandra/spark/reader/StorageAttachedIndexApplierTest.java
skoppu22: As StorageAttachedIndexApplier.applyTo claims 'idempotent if the table already carries indexes', we are testing the same

cassandra-five-zero-types/src/main/java/org/apache/cassandra/spark/reader/SchemaBuilder.java
skoppu22: This is a merge problem. 5.0 SchemaBuilder was not existing and came into upstream yesterday through https://github.com/apache/cassandra-analytics/commit/39ce05672cb371bccc479d8e8c85d468fe5944dc#diff-ace84ae8527e72fa5543611f49a9634856d03ae1cbad5f1cf66a26b0fe2d6ae8. Before that I added 5.0 SchemaBuilder as FiveZeroSchemaBuilder. Now we have two SchemaBuilders for 5.0. I am merging FiveZeroSchemaBuilder into 5.0 SchemaBuilder. Thanks for noticing this.

cassandra-five-zero-types/src/main/java/org/apache/cassandra/spark/reader/StorageAttachedIndexApplier.java
skoppu22: My AI agent doing this, I have updated it to remember this guideline. Updated the javadoc

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteSAIIndexTest.java
skoppu22: Done

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteSAIIndexTest.java
skoppu22: Done

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/BulkWriteSAIIndexTest.java
skoppu22: Done

cassandra-five-zero-types/src/main/java/org/apache/cassandra/spark/reader/FiveZeroSchemaBuilder.java
skoppu22: Renamed to existingTableMetadata

cassandra-five-zero-types/src/main/java/org/apache/cassandra/spark/reader/SchemaBuilder.java
skoppu22: Merged FiveZeroSchemaBuilder into 5.0's SchemaBuilder

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriterTest.java
skoppu22: Done

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
skoppu22: Updated comments

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlTable.java
```diff
@@ -265,9 +256,9 @@ public boolean cdc()
         return cdc;
     }
 
-    public int indexCount()
+    public Set<String> indexStatements()
     {
-        return indexCount;
+        return indexStatements;
```
jmckenzie-dev, line 261: If our goal is to keep external consumers from mutating the contents of this set we should do something like:
```java
public Set<String> indexStatements() {
    return Collections.unmodifiableSet(indexStatements);
}
```

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/CqlUtils.java
```diff
@@ -259,16 +264,88 @@ public static Set<String> extractUdts(@NotNull String schemaStr, @NotNull String
...
+
+    /**
+     * Extracts CREATE INDEX statements for the given table from the schema string.
+     *
+     * @param schemaStr full cluster schema text
+     * @param keyspace  the keyspace name
+     * @param table     the table name
+     * @return set of CREATE INDEX statements for the table
+     */
+    public static Set<String> extractIndexStatements(@NotNull String schemaStr,
```
jmckenzie-dev, line 279: This caught my eye. I think there's a couple regex specific issues here; had a couple models take a couple passes at it and this is where we landed:

>   1. Stray ? after the keyspace makes its last character optional. The format string "\"?%s?\"?" expands to "?<keyspace>?"?, so the trailing ? binds the final character of the keyspace — asking for keyspace orders also matches order. Note the table side (correctly) has no ?; that asymmetry is the tell. This is  copy-pasted from extractCleanedTableSchema, so it's pre-existing, but it's still wrong.

> 2. Identifiers are interpolated raw into the pattern. A keyspace/table name containing a regex metacharacter (legal in quoted CQL identifiers) misbehaves or throws PatternSyntaxException. Wrap both in Pattern.quote(...).

> 2. ⚠  These two are separate fixes. Pattern.quote alone does not fix bug #1 — the ? lives in the format string outside the %s, so it survives quoting. We need to delete the ? and quote the value.

> 3. Cosmetic: \.{1} — the {1} is dead; it's just \..

Whenever I see big blocks of regex like this it always makes my "spidey senses tingle". Regexes are super powerful but are effectively embedded functions and edge-cases and "sanitization" of input become quite challenging when there's this much density expressed in one place.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BroadcastableTableSchema.java
```diff
@@ -166,6 +171,11 @@ public boolean isQuoteIdentifiers()
         return quoteIdentifiers;
     }
 
+    public Set<String> getIndexStatements()
```
jmckenzie-dev, line 175: Same here; returning a collection an external user could mutate under us. 😬

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/reader/SchemaTests.java
```diff
@@ -376,7 +378,7 @@ public void testUdts(CassandraBridge bridge)
                                                             + "  nationality text,\n"
                                                             + "  weight float,\n"
                                                             + "  height int\n"
-                                                            + ");"));
+                                                            + ");"), null, Collections.emptySet(), false);
```
jmckenzie-dev: Can update various call sites to simplify them now w/out the vestigial UUID and boolean now:
```java
    public CqlTable buildSchema(String createStatement,
                                String keyspace,
                                ReplicationFactor replicationFactor,
                                Partitioner partitioner,
                                Set<String> udts,
                                Set<String> indexStatements)
```

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/bridge/SSTableWriterImplementation.java
```diff
@@ -0,0 +1,187 @@
...
+        {
+            builder.withType(udt);
+        }
+
+        builder.inDirectory(inDirectory)
+               .forTable(createStatement)
+               .withPartitioner(cassPartitioner)
+               .using(insertStatement)
+               // The data frame to write is always sorted,
+               // see org.apache.cassandra.spark.bulkwriter.CassandraBulkSourceRelation.insert
```
jmckenzie-dev: Is this a contractual promise from the other class or could this drift? There's no javadoc on `CassandraBulkSourceRelation#insert` and no javadoc on the parent class indicating that this is a durable "external API" commitment here.

jmckenzie-dev (changes requested):
There are a lot of places where we call the old `CassandraBridge#buildSchema` where we can simplify to the signature assuming null UUID and false on enableCdc which would tidy up the patch a bit.

cassandra-five-zero-types/src/main/java/org/apache/cassandra/spark/reader/StorageAttachedIndexApplier.java
jmckenzie-dev: Yep. I've seen the same from both humans and LLM's. When you're working in cross-cutting domains it's easy to let the context of what's happening on the other side of an API boundary leak over in your comments on why you're doing what you're doing.

cassandra-five-zero-bridge/src/main/java/org/apache/cassandra/bridge/SSTableWriterImplementation.java
skoppu22: Added javadoc for why we need `sorted`

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/CqlTable.java
skoppu22: indexStatements is already created using Collections.unmodifiableSet above in constructor

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BroadcastableTableSchema.java
skoppu22: Done

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/reader/SchemaTests.java
skoppu22: Done

PR #222, opened 2026-07-09
CASSANALYTICS-177: Add instanceId query parameter to fix HTTP 421 err…

bianca-stanciu29:
…ors when Sidecar is behind a load balancer

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
@@ -114,9 +114,10 @@ public class BulkSparkConf implements Serializable
     //       which will throw a configuration exception for each setting with that prefix it does not recognize
     public static final String SETTING_PREFIX = "spark.cassandra_analytics.";
 
-    // Cassandra version of target cluster. Configuration parameter is exposed to be able to correctly initialize static
-    // components, before cluster version is discovered via Sidecar.
-    public static final String CASSANDRA_VERSION                       = SETTING_PREFIX + "cassandra.version";
+    // Disable SSTable version-based bridge determination. When true, falls back to legacy mode,
+    // which selects the bridge based on the Cassandra version returned by the Sidecar.
+    // This provides a safety fallback mechanism if SSTable version detection fails or encounters issues.
+    public static final String DISABLE_SSTABLE_VERSION_BASED_BRIDGE    = SETTING_PREFIX + "bridge.disable_sstable_version_based";
     public static final String HTTP_MAX_CONNECTIONS                    = SETTING_PREFIX + "request.max_connections";
     public static final String HTTP_RESPONSE_TIMEOUT                   = SETTING_PREFIX + "request.response_timeout";
     public static final String HTTP_CONNECTION_TIMEOUT                 = SETTING_PREFIX + "request.connection_timeout";
@@ -170,6 +171,7 @@ public class BulkSparkConf implements Serializable
     protected final String configuredJobId;
     protected boolean useOpenSsl;
     protected int ringRetryCount;
+    protected final boolean disableSSTableVersionBasedBridge;
     // create sidecarInstances from sidecarContactPointsValue and effectiveSidecarPort
     private final String sidecarContactPointsValue; // It takes comma separated values
     private transient Set<SidecarInstance> sidecarContactPoints; // not serialized
@@ -222,6 +224,7 @@ public class BulkSparkConf implements Serializable
         // else fall back to props, and then default if neither specified
         this.useOpenSsl = getBoolean(USE_OPENSSL, true);
         this.ringRetryCount = getInt(RING_RETRY_COUNT, DEFAULT_RING_RETRY_COUNT);
+        this.disableSSTableVersionBasedBridge = getBoolean(DISABLE_SSTABLE_VERSION_BASED_BRIDGE, false);
         this.importCoordinatorTimeoutMultiplier = getDouble(IMPORT_COORDINATOR_TIMEOUT_MULTIPLIER, 0.5);
         this.ttl = MapUtils.getOrDefault(options, WriterOptions.TTL.name(), null);
         this.timestamp = MapUtils.getOrDefault(options, WriterOptions.TIMESTAMP.name(), null);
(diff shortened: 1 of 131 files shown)


skoppu22 (changes requested):
This PR stamps the same instanceId on every request — so if the job writes to more than one instance, all but one get mislabeled.

Example:
Setup (a normal 3-node bulk write)
Cassandra ring:   instance 1  |  instance 2  |  instance 3
Sidecar (one config knows all 3, ids 1/2/3), fronted by a load balancer at  sidecar-lb:9043
Operator sets:    spark.cassandra_analytics.sidecar.instance.id = 2
The writer builds ONE client over all three (CassandraContext.java:90):

AnalyticsSidecarClient.from(new SimpleSidecarInstancesProvider(clusterConfig /* 3 instances */), conf)
Copy
What happens per request (with this PR)
request meant for instance 1  ──▶  ...?instanceId=2  ──▶  server: instanceFromId(2) ──▶ instance 2   ❌ WRONG
request meant for instance 2  ──▶  ...?instanceId=2  ──▶  server: instanceFromId(2) ──▶ instance 2   ✅ right
request meant for instance 3  ──▶  ...?instanceId=2  ──▶  server: instanceFromId(2) ──▶ instance 2   ❌ WRONG
The ?instanceId=2 overrides the Host header on the server, so the server confidently routes the instance-1 and instance-3 work to instance 2. Because config.instanceId() is a single job-global value, vertxRequest(...) applies it uniformly even though it already knows the real target (sidecarInstance) and could distinguish them.

Net: a fixed id is only correct when the job touches exactly one instance. For any real multi-node ring it silently misroutes → the data-correctness risk behind the WARNING.

bianca-stanciu29 (commented):
hi @skoppu22 
Thanks for flagging this, it's addressed now.

vertxRequest prefers the target instance's own resolved id over the job-level one, for
any ring size (not just 3): if every instance is individually addressable and tagged via
`host[:port]=<id>`, each request gets the right id regardless of instance count. If any
instance isn't resolved, validateSidecarInstanceIdCoverage fails fast with a clear error
instead of silently misrouting.

To be upfront: your exact topology, multiple instances behind one shared address, still
can't work with this PR alone, at any scale. A hostname-keyed lookup can't disambiguate targets sharing one address; it now fails fast with "Duplicate key" instead of resolving(see testSidecarInstanceIdsByHostnameThrowsWhenSharedHostnameHasDifferentIds). 
Making that work needs Sidecar to report each instance's own id in the ring response, tracking that as a follow-up.

This is the same gap you flagged on #223 (CASSANALYTICS-181), single-cluster jobs behind a load balancer hit it too, and a real fix should cover both paths generically rather than being patched per-PR. 
 Appreciate a re-review when you get a chance.

bianca-stanciu29 (commented):
Hi @skoppu22 
Follow-up is up: [apache/cassandra-sidecar#380 (CASSSIDECAR-495)](https://github.com/apache/cassandra-sidecar/pull/380).

It adds `sidecarInstanceId` to both the ring and token-range-replicas
responses, reported per node/replica by the Sidecar itself (via its
local instances config), instead of the client trying to derive it
from a hostname it observes.

This covers the shared-address case you flagged here: multiple
Cassandra instances fronted by one Sidecar address (e.g. behind a
load balancer). The client no longer needs a hostname-keyed lookup at
all, it reads the correct `instanceId` straight off the ring/
token-range response for each node

skoppu22:
> Hi @skoppu22 Follow-up is up: [apache/cassandra-sidecar#380 (CASSSIDECAR-495)](https://github.com/apache/cassandra-sidecar/pull/380).
> 
> It adds `sidecarInstanceId` to both the ring and token-range-replicas responses, reported per node/replica by the Sidecar itself (via its local instances config), instead of the client trying to derive it from a hostname it observes.
> 
> This covers the shared-address case you flagged here: multiple Cassandra instances fronted by one Sidecar address (e.g. behind a load balancer). The client no longer needs a hostname-keyed lookup at all, it reads the correct `instanceId` straight off the ring/ token-range response for each node

Sorry for the delay, I am taking a look now

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/RingInstance.java
```diff
@@ -39,10 +39,22 @@ public class RingInstance implements CassandraInstance, Serializable
     private static final long serialVersionUID = 4399143234683369652L;
     private RingEntry ringEntry;
     private @Nullable String clusterId;
+    private @Nullable Integer sidecarInstanceId;
```
skoppu22, line 42: [minor] Need to update serialVersionUID as we are adding a new param

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -126,6 +126,7 @@ public class BulkSparkConf implements Serializable
     public static final String SIDECAR_REQUEST_RETRY_DELAY_MILLIS      = SETTING_PREFIX + "sidecar.request.retries.delay.milliseconds";
     public static final String SIDECAR_REQUEST_MAX_RETRY_DELAY_MILLIS  = SETTING_PREFIX + "sidecar.request.retries.max.delay.milliseconds";
     public static final String SIDECAR_REQUEST_TIMEOUT_SECONDS         = SETTING_PREFIX + "sidecar.request.timeout.seconds";
+    public static final String SIDECAR_INSTANCE_ID                     = SETTING_PREFIX + "sidecar.instance.id";
```
skoppu22, line 129: Why do we need global instance ID? Why don't we drop sidecar.instance.id, rely on dynamic ids and the host:port=<id> suffix ? That deletes the possibility of incorrectly using global instance ID.

bianca-stanciu29: Yes, you're right, I missed this. Good catch!

I removed sidecar.instance.id and the whole global ID path, including instanceId from HttpClientConfig and validateSidecarInstanceIdCoverage.

Routing now only uses the per-instance ID reported by Sidecar ([#380](https://github.com/apache/cassandra-sidecar/pull/380)) or the host:port=<id> contact-point suffix as a fallback for older Sidecars.

The behavioral change is that a single instance behind a load balancer now needs the ID on the contact point or a Sidecar that reports it dynamically.

Tests are updated accordingly. Ready for another look, thanks!

PR #223, opened 2026-07-13
[CASSANALYTICS-181] - Support Sidecar behind a load balancer for time…

bianca-stanciu29:
…-skew validation in coordinated writes

 ## Problem                                                                                                            
                  
  Time-skew validation before a coordinated bulk write fans out HTTP calls directly to per-replica FQDNs resolved from the Cassandra token map. In deployments where Sidecar is fronted by a load balancer, those FQDNs are not routable from Spark executors, only the load balancer address is. This causes  time-skew validation to fail at the start of every coordinated write job,  blocking bulk writes entirely in load-balanced Sidecar deployments.                                                                 
  ## Solution                                                                                                           
                  
  Add a SIDECAR_BEHIND_LOAD_BALANCER writer option (default: false). When true, time-skew validation skips the per-replica fan-out and routes through the preconfigured load balancer contact points instead

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
@@ -146,6 +146,7 @@ public class BulkSparkConf implements Serializable
     public final double importCoordinatorTimeoutMultiplier;
     public boolean quoteIdentifiers;
     public final boolean skipSecondaryIndexCheck;
+    public final boolean sidecarBehindLoadBalancer;
     protected final String keystorePassword;
     protected final String keystorePath;
     protected final String keystoreBase64Encoded;
@@ -229,6 +230,7 @@ public class BulkSparkConf implements Serializable
         this.timestamp = MapUtils.getOrDefault(options, WriterOptions.TIMESTAMP.name(), null);
         this.quoteIdentifiers = MapUtils.getBoolean(options, WriterOptions.QUOTE_IDENTIFIERS.name(), false, "quote identifiers");
         this.skipSecondaryIndexCheck = MapUtils.getBoolean(options, WriterOptions.SKIP_SECONDARY_INDEX_CHECK.name(), false, "skip secondary index check");
+        this.sidecarBehindLoadBalancer = MapUtils.getBoolean(options, WriterOptions.SIDECAR_BEHIND_LOAD_BALANCER.name(), false, "sidecar behind load balancer");
         int storageClientConcurrency = MapUtils.getInt(options, WriterOptions.STORAGE_CLIENT_CONCURRENCY.name(),
                                                        DEFAULT_STORAGE_CLIENT_CONCURRENCY, "storage client concurrency");
         long storageClientKeepAliveSeconds = MapUtils.getLong(options, WriterOptions.STORAGE_CLIENT_THREAD_KEEP_ALIVE_SECONDS.name(),
@@ -276,6 +278,11 @@ public class BulkSparkConf implements Serializable
         this.configuredJobId = MapUtils.getOrDefault(options, WriterOptions.JOB_ID.name(), null);
         this.coordinatedWriteConfJson = MapUtils.getOrDefault(options, WriterOptions.COORDINATED_WRITE_CONFIG.name(), null);
         this.coordinatedWriteConf = buildCoordinatedWriteConf(dataTransportInfo.getTransport(), logger);
+        if (this.sidecarBehindLoadBalancer && this.coordinatedWriteConf == null && logger != null)
+        {
+            logger.warn("{} is set but {} is not configured; the flag is only honored for coordinated writes and will be ignored on the single-cluster path.",
+                        WriterOptions.SIDECAR_BEHIND_LOAD_BALANCER, WriterOptions.COORDINATED_WRITE_CONFIG);
+        }
         this.digestAlgorithmSupplier = digestAlgorithmSupplierFromOptions(dataTransport, options);
         validateEnvironment();
     }
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
@@ -278,11 +278,6 @@ public class BulkSparkConf implements Serializable
         this.configuredJobId = MapUtils.getOrDefault(options, WriterOptions.JOB_ID.name(), null);
         this.coordinatedWriteConfJson = MapUtils.getOrDefault(options, WriterOptions.COORDINATED_WRITE_CONFIG.name(), null);
         this.coordinatedWriteConf = buildCoordinatedWriteConf(dataTransportInfo.getTransport(), logger);
-        if (this.sidecarBehindLoadBalancer && this.coordinatedWriteConf == null && logger != null)
-        {
-            logger.warn("{} is set but {} is not configured; the flag is only honored for coordinated writes and will be ignored on the single-cluster path.",
-                        WriterOptions.SIDECAR_BEHIND_LOAD_BALANCER, WriterOptions.COORDINATED_WRITE_CONFIG);
-        }
         this.digestAlgorithmSupplier = digestAlgorithmSupplierFromOptions(dataTransport, options);
         validateEnvironment();
     }
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
@@ -246,16 +246,7 @@ public class CassandraClusterInfo implements ClusterInfo, Closeable
         TimeSkewResponse timeSkew;
         try
         {
-            TokenRangeMapping<RingInstance> topology = getTokenRangeMapping(true);
-            List<SidecarInstance> instances = topology.getSubRanges(range)
-                                                      .asMapOfRanges()
-                                                      .values()
-                                                      .stream()
-                                                      .flatMap(Collection::stream)
-                                                      .distinct() // remove duplications
-                                                      .map(replica -> new SidecarInstanceImpl(replica.nodeName(), getCassandraContext().sidecarPort()))
-                                                      .collect(Collectors.toList());
-            timeSkew = getCassandraContext().getSidecarClient().timeSkew(instances).get();
+            timeSkew = fetchTimeSkew(range).get();
         }
         catch (InterruptedException | ExecutionException exception)
         {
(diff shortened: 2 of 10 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
```diff
@@ -95,7 +95,29 @@ public class CassandraClusterInfoGroup implements ClusterInfo, MultiClusterSuppo
...
+     * Selects the {@link CassandraClusterInfo} concrete type based on the
+     * {@link BulkSparkConf#sidecarBehindLoadBalancer} flag. Kept centralized so the
+     * driver-side factory and the executor-side broadcast reconstruction stay in lockstep.
+     */
+    private static CassandraClusterInfo createClusterInfo(BulkSparkConf conf, String clusterId)
+    {
+        if (conf.sidecarBehindLoadBalancer)
+        {
+            LOGGER.info("Using CoordinatedCassandraClusterInfo for load-balanced Sidecar. clusterId={}", clusterId);
+            return new CoordinatedCassandraClusterInfo(conf, clusterId);
```
skoppu22: This name CoordinatedCassandraClusterInfo is misleading. For the reader, this naming sounds like Coordinated work is happening only when sidecarBehindLoadBalancer is true, and else case below doing non-coordinated work. Need to pick appropriate name instead of CoordinatedCassandraClusterInfo

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
```diff
@@ -95,7 +95,29 @@ public class CassandraClusterInfoGroup implements ClusterInfo, MultiClusterSuppo
...
+        }
+        return new CassandraClusterInfo(conf, clusterId);
+    }
+
+    private static CassandraClusterInfo createClusterInfo(BroadcastableClusterInfo bci)
+    {
+        return bci.getConf().sidecarBehindLoadBalancer
+               ? new CoordinatedCassandraClusterInfo(bci)
```
skoppu22: Same as above

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/LoadBalancedCassandraClusterInfo.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public CoordinatedCassandraClusterInfo(BroadcastableClusterInfo broadcastable)
+    {
+        super(broadcastable);
+    }
+
+    @Override
+    protected CompletableFuture<TimeSkewResponse> fetchTimeSkew(Range<BigInteger> range)
+    {
+        return getCassandraContext().getSidecarClient().timeSkew();
```
skoppu22, line 58: The same problem exists for Single cluster behind a load balancer. Hence instead of using 'Coordinated' word, good to develop this as a generic framework for handling cluster/clusters behind a load balancer, then we can invoke it from anywhere needed

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
```diff
@@ -276,6 +278,11 @@ public BulkSparkConf(SparkConf conf, Map<String, String> options, @Nullable Logg
         this.configuredJobId = MapUtils.getOrDefault(options, WriterOptions.JOB_ID.name(), null);
         this.coordinatedWriteConfJson = MapUtils.getOrDefault(options, WriterOptions.COORDINATED_WRITE_CONFIG.name(), null);
         this.coordinatedWriteConf = buildCoordinatedWriteConf(dataTransportInfo.getTransport(), logger);
+        if (this.sidecarBehindLoadBalancer && this.coordinatedWriteConf == null && logger != null)
+        {
+            logger.warn("{} is set but {} is not configured; the flag is only honored for coordinated writes and will be ignored on the single-cluster path.",
```
skoppu22: What about single cluster behind a load balancer? Can you create a Jira for supporting this case and add Jira here as a TODO ?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CoordinatedCassandraClusterInfo.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Variant of {@link CassandraClusterInfo} used when Sidecars are fronted by a load balancer.
+ * Replica FQDNs from the token map are not routable from Spark executors in that topology,
+ * so time-skew validation must query the configured contact points (load balancer endpoints)
+ * rather than per-range replicas. Selected when
+ * {@link org.apache.cassandra.spark.bulkwriter.WriterOptions#SIDECAR_BEHIND_LOAD_BALANCER}
+ * is set.
+ */
+public class CoordinatedCassandraClusterInfo extends CassandraClusterInfo
```
skoppu22: As I mentioned above, this naming is incorrect, this class is for Cluster behind a LB, nothing specific to Corodinated

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/LoadBalancedCassandraClusterInfo.java
skoppu22: Also, range param is unused, hence mention a comment something like 'range is irrelevant; contact points are queried directly'

skoppu22 (changes requested):

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
skoppu22: I would recommend making LB handling as generic framework and invoke it from both coordinated and single cluster case, then we will have uniformity and support for both cases.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
bianca-stanciu29: Renamed to LoadBalancedCassandraClusterInfo and moved it out of the cloudstorage.coordinated package into bulkwriter, since it isn't specific to coordinated writes. The concrete type is now selected via CassandraClusterInfo.create(...).

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CassandraClusterInfoGroup.java
bianca-stanciu29: Done, same rename applied here, this now calls CassandraClusterInfo.create(...).

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/LoadBalancedCassandraClusterInfo.java
bianca-stanciu29: Made it generic: the load-balanced variant is now selected by CassandraClusterInfo.create(...), which is invoked from both the single-cluster path (CassandraBulkWriterContext / BroadcastableClusterInfo.reconstruct) and the coordinated path (CassandraClusterInfoGroup), so it can be reused wherever needed.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/BulkSparkConf.java
bianca-stanciu29: Went with your recommendation to make it generic rather than defer it: SIDECAR_BEHIND_LOAD_BALANCER is now honored on both the single-cluster and coordinated write paths, so single-cluster behind a load balancer is supported directly for time-skew validation. Removed this warning since the flag is no longer ignored on the single-cluster path, no follow-up Jira needed for this case.

For context: per-instance request routing behind a shared load-balancer address is a separate concern, enabled on the Sidecar side by [CASSSIDECAR-495](https://issues.apache.org/jira/browse/CASSSIDECAR-495)

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/coordinated/CoordinatedCassandraClusterInfo.java
bianca-stanciu29: Agreed, renamed to LoadBalancedCassandraClusterInfo and moved out of the coordinated package, since it's about a cluster behind a load balancer, not coordinated writes specifically.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfoTest.java
```diff
@@ -63,6 +71,75 @@ void testTimeSkewAcceptable()
...
+
+        assertThatNoException()
+        .describedAs("Load-balanced cluster info must use timeSkew() so load balancer contact points stay reachable")
+        .isThrownBy(() -> ci.validateTimeSkewWithLocalNow(range(10, 20), localNow));
+        verify(sidecarClient).timeSkew();
+        verify(sidecarClient, never()).timeSkew(anyList());
+    }
+
+    @Test
+    void testCreateSelectsLoadBalancedClusterInfoWhenSidecarBehindLoadBalancer()
```
skoppu22, line 109: Can you please write test cases to ensure
- Coordinated case creates LoadBalancedCassandraClusterInfo for each cluster when LB is enabled
- BroadcastableClusterInfo creates LoadBalancedCassandraClusterInfo on executors

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/LoadBalancedCassandraClusterInfo.java
skoppu22: Nice work

skoppu22 (approved):
Looks good, please add test cases mentioned

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfoTest.java
bianca-stanciu29: Done, added the tests. The coordinated one checks `fromBulkSparkConf(...)` builds a `LoadBalancedCassandraClusterInfo` per cluster when the flag is on, and the executor one checks `BroadcastableClusterInfo.reconstruct()` returns one too. Also added the flag-off cases on both paths to catch the inverse regression.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -138,6 +138,57 @@ public CassandraClusterInfo(BroadcastableClusterInfo broadcastable)
...
 
+    /**
+     * Creates a {@link CassandraClusterInfo} for a single cluster, selecting the concrete type based on the
+     * {@link BulkSparkConf#sidecarBehindLoadBalancer} flag.
+     *
+     * @param conf bulk write conf
+     * @return {@link LoadBalancedCassandraClusterInfo} when Sidecar is behind a load balancer, otherwise a plain
+     *         {@link CassandraClusterInfo}
+     */
+    public static CassandraClusterInfo create(BulkSparkConf conf)
```
bbotella, line 149: Maybe a nit, but at least I'd like for us to consider.

We are adding a new class `LoadBalancedCassandraClusterInfo` that extends `CassandraClusterInfo`, which is fine, but:
`CassandraClusterInfo` may return an instance of itself or an instance of the new class, forcing it to "know" (depend) on `LoadBalancedCassandraClusterInfo`. This feels like a factory pattern with a potential abstract `CassandraClusterInfo` class and two classes extending it: `LoadBalancedCassandraClusterInfo` and maybe `StandaloneCassandraClusterInfo`, and the factory static create method returning one or the other? I think that approach would be cleaner. Opinions?

bianca-stanciu29: Good point, thanks.
I agree that having the base class reference its subclass in `create()` isn’t ideal.

I think the abstract CassandraClusterInfo + StandaloneCassandraClusterInfo / LoadBalancedCassandraClusterInfo approach makes sense. My only hesitation is the scope. CassandraClusterInfo is quite central, and making it abstract would mean introducing a separate StandaloneCassandraClusterInfo and updating the places that currently instantiate or mock CassandraClusterInfo directly. That would likely touch a number of tests and mocks as well.

Since the PR is otherwise ready, I’d prefer to keep this change smaller for now and revisit the full split separately. But happy to go with the cleaner approach if you think it’s worth doing as part of this PR.

PR #224, opened 2026-07-14
CASSANALYTICS-182 : CDC batch-write mixing a CDC-enabled and CDC-disabled table drops the CDC table's mutation

jyothsnakonisa:


cassandra-analytics-cdc/src/main/java/org/apache/cassandra/cdc/Cdc.java
```diff
@@ -415,22 +415,33 @@ protected void refreshSchema()
...
-                // update Schema instance with latest schema
-                cdcBridge().updateCdcSchema(tables, cdcOptions.partitioner(), tableIdLookup);
+                // Filter CDC-enabled tables for publishing decisions
+                Set<CqlTable> cdcTables = allTables.stream()
+                                                   .filter(CqlTable::cdc)
+                                                   .collect(Collectors.toSet());
+                this.cdcEnabledTables = cdcTables;
+                if (cdcTables.isEmpty())
+                {
+                    LOGGER.warn("No CDC-enabled tables found");
```
jmckenzie-dev: In this case, we're proceeding to update the schema we have cached to effectively wipe any prior state in the "no cdc enabled" case. What do you think about us adding this context to the log? Because my initial naive assumption was "if we hit this point why don't we fast-fail out?" and then realized "Ah yes. We need to have our schema reflect this".

Concretely, something like:
"No CDC-enabled tables found; no cdc updates will be processed until CDC is enabled on a table's schema"?

cassandra-analytics-cdc-codec/src/main/java/org/apache/cassandra/cdc/schemastore/CachingSchemaStore.java
```diff
@@ -118,11 +122,16 @@ public void onSchemaChange()
                     }
                     return value;
                 });
+            }
+            // publishSchemas() re-fetches and republishes every CDC table's schema, so it only
+            // needs to run once after the cache has been updated for all changed tables above —
+            // calling it per-table would redundantly re-fetch the schema and republish every
+            // table N times over.
+            if (!refreshedCdcTables.isEmpty())
```
jmckenzie-dev, line 125: What happens in the case where we had CDC tables then the DDL was changed and they were disabled? Wouldn't we skip the publish here so we never get the memo that there's no more CDC enabled?

cassandra-analytics-cdc-codec/src/main/java/org/apache/cassandra/cdc/schemastore/CachingSchemaStore.java
```diff
@@ -160,8 +169,11 @@ protected CqlToAvroSchemaConverter schemaConverter()
     private void publishSchemas()
     {
         schemaSupplier
-        .getCdcEnabledTables()
-        .thenAccept(refreshedCdcTables -> {
+        .getTables()
+        .thenAccept(allTables -> {
+            Set<CqlTable> refreshedCdcTables = allTables.stream()
```
jmckenzie-dev: We have this pattern of filtering down to just the CDC enabled tables in multiple places. What do you think about augmenting the `SchemaSupplier` to provide both:
```java
    CompletableFuture<Set<CqlTable>> getTables();
    CompletableFuture<Set<CqlTable>> getCDCEnabledTables();
```

That'd tidy up a lot of this duplicated stream code and give us a single inflection point to make modifications if this pattern needs to change (i.e. `CqlTable::cdc` becomes something else, or needs more data to resolve, for instance)

cassandra-analytics-cdc-sidecar/src/main/java/org/apache/cassandra/cdc/sidecar/SidecarCdc.java
```diff
@@ -92,11 +92,12 @@ public static SidecarCdcBuilder builder(@NotNull String jobId,
 
     public void initSchema()
     {
-        Set<CqlTable> tables = FutureUtils.get(schemaSupplier.getCdcEnabledTables());
-        Optional<ReplicationFactor> rfOp = tables.stream()
-                                                 .map(CqlTable::replicationFactor)
-                                                 .filter(rf -> rf.getOptions().containsKey(dc()))
-                                                 .max(Comparator.comparingInt(rf -> rf.getOptions().get(dc())));
+        Set<CqlTable> allTables = FutureUtils.get(schemaSupplier.getTables());
+        Optional<ReplicationFactor> rfOp = allTables.stream()
```
jmckenzie-dev: Same thing here; the pattern of:
1. `schemaSupplier.getTables()`
2. `.filter(CqlTable::cdc)`

cassandra-analytics-cdc/src/main/java/org/apache/cassandra/cdc/Cdc.java
```diff
@@ -415,22 +415,33 @@ protected void refreshSchema()
...
                 {
                     LOGGER.warn("Error refreshing schema", throwable);
                     return null;
                 }
-                this.cdcEnabledTables = tables;
-                if (tables == null || tables.isEmpty())
+                if (allTables == null || allTables.isEmpty())
                 {
-                    LOGGER.warn("No CQL enabled tables");
+                    LOGGER.warn("No tables returned from schema supplier");
```
jmckenzie-dev: Is this a warn condition or ERROR? Is there ever a non-error case where a schema supplier would return 0 tables at all for the "getAllTables" shaped case?

cassandra-analytics-cdc/src/test/java/org/apache/cassandra/cdc/CdcTests.java
```diff
@@ -266,6 +278,171 @@ public List<CdcState> loadState(String jobId, int partitionId, @Nullable TokenRa
...
+                                                    tablesPassedToBridge))
+        {
+            // start() flips isRunning to true and calls refreshSchema() synchronously (the
+            // schemaSupplier future is already completed, so the .handle()/.whenComplete()
+            // chain runs inline). scheduleRun()/scheduleMonitorSchema() are overridden to
+            // no-ops in RecordingCdc, so this only exercises the schema refresh path.
+            cdc.start();
+
+            assertThat(tablesPassedToBridge.get())
+            .as("Schema.instance must be updated with ALL tables (CDC and non-CDC) so a "
```
jmckenzie-dev: Do we normally have . continuations indented on the same line as their root? I find that a little hard / surprising to read (i.e. it's not immediately clear these are coupled / dependent operations from code structure, have to infer from leading .).

Vs for example:
```java
            assertThat(tablesPassedToBridge.get())
                .as("Schema.instance must be updated with ALL tables (CDC and non-CDC) so a "...
                .containsExactlyInAnyOrder(cdcTable, nonCdcTable);
```
I won't push back if our project norm is . on the same line as the lines above, but I personally pretty strongly prefer the indentation.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/KryoSerializationTests.java
```diff
@@ -269,6 +270,38 @@ public void testCqlTable(CassandraBridge bridge)
         assertThat(deserialized).isEqualTo(table);
     }
 
+    @ParameterizedTest
+    @MethodSource("org.apache.cassandra.bridge.VersionRunner#bridges")
+    public void testCqlTableCdcFlagSurvivesSerialization(CassandraBridge bridge)
+    {
+        // CqlTable.equals()/hashCode() deliberately exclude the cdc field (see CqlTable.java),
```
jmckenzie-dev: Why don't we have the `cdc` flag as part of our equality comparison and hashcode impl?

cassandra-four-zero-types/src/main/java/org/apache/cassandra/bridge/CassandraSchema.java
```diff
@@ -304,31 +305,27 @@ public static void updateCdcSchema(@NotNull Schema schema,
             UUID tableId = tableIdLookup.lookup(table.keyspace(), table.table());
             if (cdcEnabledTables.containsKey(table.keyspace()) && cdcEnabledTables.get(table.keyspace()).contains(table.table()))
             {
-                // table has cdc enabled already, update schema if it has changed
-                LOGGER.debug("CDC already enabled keyspace={} table={}", table.keyspace(), table.table());
+                // table already registered in Schema.instance — update if schema changed
+                LOGGER.debug("Table already in Schema.instance, updating if changed keyspace={} table={}", table.keyspace(), table.table());
```
jmckenzie-dev: Since this can noop presumably on a lot of cases, I think we should consider logging in the `maybeUpdateSchema` call in the case where we have an update. The prior logging here made sense as it was a dupe process; here it's just telling us that we already knew a table existed but not much else.

cassandra-four-zero-types/src/main/java/org/apache/cassandra/bridge/CassandraSchema.java
```diff
@@ -304,31 +305,27 @@ public static void updateCdcSchema(@NotNull Schema schema,
...
                 cdcEnabledTables.get(table.keyspace()).remove(table.table());
-                CassandraSchema.maybeUpdateSchema(schema, partitioner, table, tableId, true);
-                Preconditions.checkArgument(CassandraSchema.isCdcEnabled(schema, table),
-                                            "CDC not enabled for table: " + table.keyspace() + "." + table.table());
+                CassandraSchema.maybeUpdateSchema(schema, partitioner, table, tableId, table.cdc());
                 continue;
```
jmckenzie-dev: Why the if / continue structure (which is arguably glorified modern-day `goto`) instead of an if / else if / else structure?

cassandra-four-zero-types/src/main/java/org/apache/cassandra/bridge/CassandraSchema.java
jmckenzie-dev: This confused me. The name `cdcEnabledTables` followed by:
```java
        cdcEnabledTables.forEach((ks, tables) -> tables.forEach(table -> {
            LOGGER.warn("Disabling CDC on table keyspace={} table={}", ks, table);
            CassandraSchema.disableCdc(schema, ks, table);
        }));
```
Read that we were removing cdcEnabledTables from our schema.

Renaming this variable to `remainingCdcEnabledTables` or `schemaCdcEnabledTables` plus a comment would help disambiguate this so it's clear we're not unregistering cdc enabled tables there at the end of the loop. The logic looks right but this naming threw me for a loop.

cassandra-four-zero-types/src/main/java/org/apache/cassandra/bridge/CassandraSchema.java
```diff
@@ -337,13 +334,92 @@ public static void updateCdcSchema(@NotNull Schema schema,
...
+                // Never let one misbehaving table abort unregistration of the rest of the
+                // batch — the caller (CassandraClusterSchemaMonitor) treats this call as
+                // all-or-nothing for bookkeeping purposes, so a partial failure here must not
+                // prevent the other (unrelated) tables in the same batch from being cleaned up.
+                LOGGER.warn("Failed to unregister table keyspace={} table={}", keyspace, table, e);
+            }
+        }
+    }
+
+    private static void unregisterTable(@NotNull Schema schema, @NotNull String keyspace, @NotNull String table)
```
jmckenzie-dev: I think we should rename this. `CassandraSchema#unregisterTable` implies it's for just straight "unregistered your tables", vs. a name like `CassandraSchema#unregisterNonCdcTables`, which is where we see the "Refusing to unregistered CDC-enabled table" log message making a lot more sense.

jmckenzie-dev (changes requested):

cassandra-analytics-cdc/src/main/java/org/apache/cassandra/cdc/Cdc.java
jyothsnakonisa: Also note that this class is not used anywhere, looks like it was added as a reference class. There is a separate implementation of the interface with this same class name in Sidecar.

cassandra-analytics-cdc/src/main/java/org/apache/cassandra/cdc/api/SchemaSupplier.java
```diff
@@ -21,13 +21,32 @@
...
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.stream.Collectors;
 
 import org.apache.cassandra.spark.data.CqlTable;
 
 /**
- * Supplies all CDC enabled tables
+ * Supplies schema for all tables relevant to CDC processing.
+ * Returns ALL tables (CDC-enabled and CDC-disabled) so that the bridge's Schema.instance
```
jmckenzie-dev: Should move this javadoc down to the method below and update this since the interface now can do either.

cassandra-analytics-cdc/src/main/java/org/apache/cassandra/cdc/Cdc.java
jmckenzie-dev: Ah. Does this mean we need to drop it?

cassandra-analytics-cdc-codec/src/main/java/org/apache/cassandra/cdc/schemastore/CachingSchemaStore.java
jmckenzie-dev: Ah - I was confused about this comment being down here as I was looking at it in isolation with the if check and thinking it was describing what the if check did, not the choice of placement of said check outside the loop. Maybe augment this comment with something like "We call this outside the loop because it needs..." to make that more clear.

jmckenzie-dev (changes requested):
Just some comment changes remaining but otherwise LGTM

cassandra-analytics-cdc/src/main/java/org/apache/cassandra/cdc/Cdc.java
jyothsnakonisa: In the documentation, it says this class is for reference. I don't see any harm in dropping it, I will leave it to you

jyothsnakonisa:
Circle CI: https://app.circleci.com/pipelines/gh/jyothsnakonisa/cassandra-analytics/101/details?useNewPipelines=true&workflowId=4f0b90ac-6f53-4eb6-9426-8d832a3dd46d

cassandra-analytics-cdc/src/main/java/org/apache/cassandra/cdc/Cdc.java
jmckenzie-dev: Eh, it's fine. Keeping it for reference doesn't hurt anything really IMO.

jmckenzie-dev (approved):
Looks good. Thanks!

PR #226, opened 2026-07-22
[CASSANALYTICS-183] - Pass quoted identifiers to sidecar API calls for mixed-case keyspace/table restore

bianca-stanciu29:


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java
```diff
@@ -289,11 +289,12 @@ protected String getCurrentKeyspaceSchema() throws Exception
     private TokenRangeReplicasResponse getTokenRangesAndReplicaSets()
     {
         CassandraContext context = getCassandraContext();
+        String quotedKeyspace = maybeQuotedIdentifier(bridge(), conf.quoteIdentifiers, conf.keyspace);
```
bbotella, line 292: nit: change name to maybeQuotedKeyspace?

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/bulkwriter/cloudstorage/CloudStorageDataTransferApiImplTest.java
```diff
@@ -0,0 +1,120 @@
...
+        sidecarClient = mock(SidecarClient.class);
+        jobInfo = mock(JobInfo.class);
+        when(jobInfo.qualifiedTableName()).thenReturn(new QualifiedTableName("MyKeyspace", "MyTable", true));
+        when(jobInfo.getRestoreJobId()).thenReturn(JOB_ID);
+        when(jobInfo.getRestoreJobId(null)).thenReturn(JOB_ID);
+        api = new CloudStorageDataTransferApiImpl(jobInfo, sidecarClient, mock(StorageClient.class), null);
+    }
+
+    @Test
+    void testCreateRestoreJobUsesQuotedIdentifiers() throws Exception
```
bbotella, line 64: What do you think about adding quoted identifiers tests to the other modified APIs?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/QualifiedTableName.java
```diff
@@ -65,6 +65,14 @@ public String keyspace()
         return keyspace;
     }
 
+    /**
+     * @return the keyspace name, quoted with double quotes when {@code quoteIdentifiers} is set
+     */
+    public String maybeQuotedKeyspace()
```
bbotella, line 71: Wondering here if we really need wrappers with the "maybe*" prefix, or we can just use this as default for keyspace() and table(). What do you think @yifan-c / @frankgh ?

PR #227, opened 2026-07-28
CASSANALYTICS-186: AbstractSchemaBuilder.build() drops the cdc flag, silently disabling CDC on all tables

jyothsnakonisa:


cassandra-four-zero-types/src/main/java/org/apache/cassandra/spark/reader/AbstractSchemaBuilder.java
```diff
@@ -479,7 +479,8 @@ public CqlTable build()
                             replicationFactor,
                             fields,
                             new HashSet<>(udts.values()),
-                            indexCount);
+                            indexCount,
+                            enableCdc);
```
yifan-c, line 483: Not an action item for this patch – we should consider applying the builder pattern to `CqlTable`. The constructor overloads are error prone.

yifan-c (approved):

jyothsnakonisa:
CI : https://app.circleci.com/pipelines/gh/jyothsnakonisa/cassandra-analytics/103/details?useNewPipelines=true&workflowId=55348d24-2416-47b6-8bde-a4e79d1b59dc#step-108-

jmckenzie-dev (approved):

PR #229, opened 2026-08-04
CASSANALYTICS-188: Expand list of architecture that supports unaligned access in FastByteOperations 

sarankk:


yifan-c (approved):

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/util/FastByteOperations.java
```diff
@@ -81,7 +81,13 @@ private static class BestHolder
         static ByteOperations getBest()
         {
             String arch = System.getProperty("os.arch");
-            boolean unaligned = arch.equals("i386") || arch.equals("x86") || arch.equals("x86_64") || arch.equals("amd64");
+            boolean unaligned = arch.equals("i386")
```
yifan-c, line 86: Please add the same disclaimer as in Cassandra source code. 

```
    // Note that s390x, aarch64, & ppc64le architectures are not officially supported and adding them here is only done out
    // of convenience for those that want to run C* on these architectures at their own risk (see #11214, #13326, & #13615)
```
#11214 refers to CASSANDRA-11214

frankgh (approved):
+1

PR #234, opened 2026-08-19
CASSANALYTICS-180: TokenPartitioner fails to detect range gap in reader

mansikhara:


cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/utils/RangeUtils.java
```diff
@@ -70,6 +74,54 @@ public static boolean isOpenClosedRange(Range<?> range)
...
+    static List<Range<BigInteger>> findUncoveredRanges(Range<BigInteger> fullRange,
+                                                       Collection<Range<BigInteger>> coveringRanges)
+    {
+        // An empty fullRange has nothing to cover, so every input would trivially look fully covered.
+        // Reject it rather than report a false "no gaps".
+        Preconditions.checkArgument(!fullRange.isEmpty(), "fullRange must not be empty");
+
+        RangeSet<BigInteger> uncovered = TreeRangeSet.create();
+        uncovered.add(fullRange);
+        coveringRanges.forEach(uncovered::remove);
```
jyothsnakonisa:
```suggestion
        uncovered.removeAll(coveringRanges); 
```
You could use removeAll instead of removing each entry in for loop

jyothsnakonisa (approved):
Looks good! Please add Changes.txt entry for this patch

yifan-c:
Sorry for the belated review. 

First of all, thanks for adding more test cases. 

I think the fix can be much smaller than what's here. The actual bug is just that `.filter(Range::isEmpty)` in `validateCompleteRangeCoverage()` is a no-op: Guava's TreeRangeSet never materializes empty ranges in `asRanges()`, so that filter can never keep anything, and the missing-range check silently always passes.

The minimal and correct fix is:
- Drop the `.filter(Range::isEmpty)` filter
- Change `Range.closed(minToken, maxToken)` to `Range.openClosed(minToken, maxToken)`, since sub-ranges everywhere else are open-closed and closed would falsely flag minToken as a gap on an otherwise healthy ring

Would you mind trimming the patch down to that?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/TokenPartitioner.java
yifan-c, line 1: We should really consolidate the two `TokenPartitioner`s, some day... Not in scope for this patch, I am just mentioning.

yifan-c (approved):

PR #236, opened 2026-08-23
Add Cassandra 6.0 support

michaelsembwever:
https://issues.apache.org/jira/browse/CASSANALYTICS-37

Cassandra Analytics reads, bulk writes and consumes CDC through one bridge jar per major version. This adds the SIXZERO bridge, pinned to 6.0-alpha2 from Maven Central, in five new modules that copy and override the five-zero and four-zero sources.

Four 6.0 changes needed work:

 - SSTable versions big-pa and bti-ea, both added for compression dictionary metadata in the CompressionInfo component.
 - A trailing dictionary section in CompressionInfo. The six-zero CompressionMetadata deserializes it, interns one dictionary per dictId so that readers of one SSTable share the zstd compressor that ZstdDictionaryCompressor keys by dictId, and releases the references it holds through close(). SSTableCache gained a removal listener to call close(), and SSTableReader holds a second reference for the life of a stream.
 - CEP-21 Transactional Cluster Metadata retypes Schema.instance to SchemaProvider and replaces transform() with submit(), which notifies no listener offline, so SchemaUpdater initializes the keyspace instances by hand after ClusterMetadataService.initializeForClients().
 - Commit log version 9 and messaging version 14, read through the shaded 6.0 deserializers.

Two defects in the existing bridges surfaced while testing the copies and are fixed here. CassandraSchema.unregisterNonCdcTable called SchemaUpdater.load, which on 5.0 only adds a keyspace and throws AlreadyExistsException for one that exists, so the table stayed registered; removeTables replaces the keyspace metadata instead. On 5.0 that commit reaches Keyspace.dropCf, which needs machinery a client-mode process never started, so removeTables clears the initialized flag for the commit only.

BridgeVersionConsistencyTest asserts that every type a bridge hands out reports the bridge's own version, which catches a six-zero module that silently inherits a version-bearing class from five-zero through the Gradle copy-forward.


cassandra-six-zero-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
```diff
@@ -0,0 +1,850 @@
...
+    @Override
+    @VisibleForTesting
+    public void writeTombstoneSSTable(Partitioner partitioner,
+                                      Path directory,
+                                      String createStatement,
+                                      String deleteStatement,
+                                      Consumer<Writer> consumer)
+    {
+        File cassFile = new File(directory.toFile());
+        try (SSTableTombstoneWriter writer = SSTableTombstoneWriter.builder()
```
sarankk: Shall we use `CQLSSTableWriter` here, since 6.0 version's `CQLSSTableWriter` takes any modification statement. Changes in `SSTableTombstoneWriter` are very similar to one in `CQLSSTableWriter`. Also with this, we can remove `SSTableTombstoneWriter` in 6.0 bridge.

cassandra-six-zero-bridge/src/main/java/org/apache/cassandra/bridge/SSTableWriterImplementation.java
```diff
@@ -0,0 +1,159 @@
...
+                                       this::onSSTablesProduced,
+                                       partitioner)
+                      .build();
+    }
+
+    private static IPartitioner determineSupportedPartitioner(String partitioner)
+    {
+        return partitioner.toLowerCase().contains("random")
+               ? RandomPartitioner.instance
+               : Murmur3Partitioner.instance;
```
sarankk: `Murmur3Partitioner.instance` variable seems to be available in 4.0 version as well. Shall we update this in 4.0 bridge and remove the class override here.

cassandra-six-zero-bridge/src/test/java/org/apache/cassandra/bridge/SSTableWriterImplementationTest.java
```diff
@@ -0,0 +1,245 @@
...
+import org.apache.cassandra.io.sstable.format.SSTableFormat;
+import org.apache.cassandra.spark.TestUtils;
+import org.apache.cassandra.util.ReflectionUtils;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Unit tests for configuring {@link SSTableWriterImplementation}
+ */
+class SSTableWriterImplementationTest
```
sarankk, line 52: We can remove this test class as well if we update in 4.0 version

cassandra-six-zero-types/src/main/java/org/apache/cassandra/spark/data/complex/AbstractCqlList.java
```diff
@@ -0,0 +1,130 @@
...
+                        int ttl,
+                        long now,
+                        Object value)
+    {
+        for (Object o : (List<?>) value)
+        {
+            if (ttl != NO_TTL)
+            {
+                rowBuilder.addCell(CqlType.expiring(cd, timestamp, ttl, now, type().serialize(o),
+                                                    CellPath.create(TimeUUID.Generator.nextTimeUUID().toBytes())));
```
sarankk: How about we use `randomCellPath()` instead in 4.0 types module? we can remove override in 6.0 types
```suggestion
                rowBuilder.addCell(CqlType.expiring(cd, timestamp, ttl, now, type().serialize(o), randomCellPath()));
```

cassandra-six-zero-types/src/main/java/org/apache/cassandra/spark/data/types/AbstractTimeUUID.java
```diff
@@ -0,0 +1,56 @@
...
+
+package org.apache.cassandra.spark.data.types;
+
+import org.apache.cassandra.cql3.functions.types.DataType;
+import org.apache.cassandra.db.marshal.AbstractType;
+import org.apache.cassandra.db.marshal.TimeUUIDType;
+import org.apache.cassandra.utils.TimeUUID;
+
+/**
+ * A copy of the four-zero class. Cassandra 5.0 removed {@code UUIDGen.getTimeUUID()}, so the inherited
```
sarankk: `TimeUUID` in 5.0 types already overrides `randomValue` method, we can remove this class in 6.0

sarankk (commented):
Thanks for this Mick! changes look good to me, left for few comments

cassandra-six-zero-types/src/main/java/org/apache/cassandra/spark/data/complex/AbstractCqlList.java
michaelsembwever: yes, and fixes a 5.0 defect

cassandra-six-zero-bridge/src/main/java/org/apache/cassandra/bridge/CassandraBridgeImplementation.java
michaelsembwever: addressed in https://github.com/apache/cassandra-analytics/pull/236/changes/fc607e72500b64816f068db01e70e1116c461966

cassandra-six-zero-bridge/src/main/java/org/apache/cassandra/bridge/SSTableWriterImplementation.java
michaelsembwever: addressed in https://github.com/apache/cassandra-analytics/pull/236/changes/fc607e72500b64816f068db01e70e1116c461966

cassandra-six-zero-bridge/src/test/java/org/apache/cassandra/bridge/SSTableWriterImplementationTest.java
michaelsembwever: not sure i understand.

the six-zero copy replaces the five-zero one in the source set. Deleting it compiles the five-zero copy against 6.0, which asserts da/oa sstable versions and then calls the removed getRawConfig().sstable.selected_format.

cassandra-six-zero-types/src/main/java/org/apache/cassandra/spark/data/complex/AbstractCqlList.java
michaelsembwever: addressed in https://github.com/apache/cassandra-analytics/pull/236/changes/fc607e72500b64816f068db01e70e1116c461966

cassandra-six-zero-types/src/main/java/org/apache/cassandra/spark/data/types/AbstractTimeUUID.java
michaelsembwever: addressed in https://github.com/apache/cassandra-analytics/pull/236/changes/fc607e72500b64816f068db01e70e1116c461966

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/ResiliencyTestBase.java
```diff
@@ -59,6 +61,33 @@ public abstract class ResiliencyTestBase extends SharedClusterSparkIntegrationTe
...
+    /**
+     * Skips the test class when the version under test declares none of the internals that the ByteBuddy hooks of a
+     * topology-change test intercept: {@code StorageService.bootstrap(Collection, long)},
+     * {@code StorageService.unbootstrap()} and {@code RangeRelocator.stream()}. CEP-21 Transactional Cluster Metadata
+     * removed all three in Cassandra 6.0, in favour of {@code org.apache.cassandra.tcm.sequences}. A hook that fails
+     * to install is silent, so the test would instead wait on a latch that never counts down.
+     *
+     * <p>Call this from {@link #beforeClusterProvisioning()}, which runs before the cluster starts.</p>
+     */
+    protected void assumeTopologyChangeHooksSupported()
```
jyothsnakonisa, line 73: This will skip several topology change related tests for 6.0 version permanently. WDYT about re writing the tests with 6.0 equivalent way of removed methods.

If you think that is too involved, we should consider doing a followup PR to fix these tests. As a starting point, BootstrapAndJoin.bootstrap(...) and UnbootstrapStreams.execute(...) look like reasonably direct replacements for the old StorageService.bootstrap()/unbootstrap() hooks

michaelsembwever:
squashed and rebased off trunk

cassandra-six-zero-types/src/main/java/org/apache/cassandra/bridge/SchemaUpdater.java
```diff
@@ -0,0 +1,129 @@
...
+     * transformation of its own. {@link #submit} is also the wrong follow-up here: it reports the table in
+     * {@code Keyspaces.diff().altered}, which makes {@code DistributedSchema.initializeKeyspaceInstances} call
+     * {@code Keyspace.dropCf} and so initialize {@code CompactionManager}, which throws in client mode where
+     * concurrent_compactors is zero. Rebuild the instances from the committed metadata instead, which leaves the
+     * removed table without a column family store and touches no compaction machinery.
+     */
+    public static void removeTables(SchemaProvider schema, KeyspaceMetadata keyspaceMetadata)
+    {
+        schema.submit(replace(keyspaceMetadata));
+        openKeyspaceInstances();
```
jyothsnakonisa: openKeyspaceInstances() diffing against DistributedSchema.empty() means it creates a keyspace instance for every keyspace, not just the one being changed. But removeTables() does have a real before schema, so instead of using openKeyspaceInstances(), can it diff against before with just the target keyspace excluded? That way only the modified keyspace gets rebuilt instead of all of them.
                                                       
```
 public static void openKeyspaceInstances()
    {
        ClusterMetadata.current().schema.initializeKeyspaceInstances(DistributedSchema.empty(), false);
    }
```

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/db/commitlog/BufferingCommitLogReader.java
```diff
@@ -553,19 +553,21 @@ private void readMutationInternal(byte[] inputBuffer,
                                                        messagingVersion,
                                                        DeserializationHelper.Flag.LOCAL);
         }
-        catch (UnknownTableException ex)
+        catch (Throwable t)
```
jyothsnakonisa, line 556: Thanks for fixing this, I saw an issue with this earlier.

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/CassandraVersionFeatures.java
```diff
@@ -32,7 +34,14 @@ public class CassandraVersionFeatures implements Comparable<CassandraVersionFeat
 {
     private static final Pattern VERSION_PATTERN_3        = Pattern.compile("(?:.+-)?([0-9]+)\\.([0-9]+)\\.([0-9]+)([a-zA-Z0-9-]*)");
     private static final Pattern VERSION_PATTERN_4        = Pattern.compile("(?:.+-)?([0-9]+)\\.([0-9]+)\\.([0-9]+)\\.([0-9]+)[a-zA-Z0-9-]*");
-    private static final Pattern VERSION_PATTERN_SNAPSHOT = Pattern.compile("(?:.+-)?([0-9]+)\\.([0-9]+)-(SNAPSHOT)$");
+    // A major version carries no patch number until its first release: cassandra-4.0-SNAPSHOT, cassandra-6.0-alpha2,
```
jyothsnakonisa, line 37: Since version pattern is changing can you please add tests for "major.minor-SNAPSHOT" with no alpha/rc tag " versions?

cassandra-six-zero-bridge/src/main/java/org/apache/cassandra/io/sstable/format/bti/BtiReaderUtils.java
jyothsnakonisa, line 1: can you please add tests for this class? there are tests for this 5.0 class, may be you can copy them for 6.0 class

cassandra-six-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/SSTableCache.java
```diff
@@ -0,0 +1,240 @@
...
+        return CacheBuilder.newBuilder()
+                           .expireAfterAccess(expireAfterMins, TimeUnit.MINUTES)
+                           .maximumSize(size)
+                           .<SSTable, Optional<CompressionMetadata>>removalListener(notification -> {
+                               Optional<CompressionMetadata> metadata = notification.getValue();
+                               if (metadata != null)
+                               {
+                                   metadata.ifPresent(CompressionMetadata::close);
+                               }
+                           })
```
jyothsnakonisa, line 132: nit: needs a test, can you force an entry out of this cache and verify the dictionary ref actually got released.

cassandra-six-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/SSTableReader.java
```diff
@@ -0,0 +1,987 @@
...
+        private final Ref<? extends CompressionDictionary> dictionaryRef;
+        private long lastTimeNanos = System.nanoTime();
+
+        SSTableStreamReader() throws IOException
+        {
+            lastToken = sparkRangeFilter != null ? sparkRangeFilter.tokenRange().upperEndpoint() : null;
+            @Nullable CompressionMetadata compressionMetadata = SSTableCache.INSTANCE.compressionMetadata(ssTable,
+                                                                                                          version.hasMaxCompressedLength(),
+                                                                                                          metadata.params.crcCheckChance);
+            dictionaryRef = compressionMetadata != null ? compressionMetadata.acquireDictionaryRef() : null;
```
jyothsnakonisa, line 721: dictionaryRef is acquired here, but openDataStream(), CompressedRawInputStream.from(...), and skipFully(...) can still throw after the reference is acquired, leaking it since no one calls close() on a constructor that never returned.                                                                                                                                                                               
                                                                                                                                                                                                                 
Since dictionaryRef is only used later in close(), can we move its acquisition to the end of the constructor, after everything that can throw has already succeeded?

jyothsnakonisa (commented):
Hey Mick! looks good overall leaving a few comments, also please rebase this PR.

cassandra-analytics-common/src/main/java/org/apache/cassandra/bridge/CassandraVersionFeatures.java
michaelsembwever: added in `CassandraVersionFeaturesTest`

see https://github.com/apache/cassandra-analytics/pull/236/changes/16d6c8fceb6fb3e535a66857e2a10f12175da2c9

cassandra-analytics-integration-tests/src/test/java/org/apache/cassandra/analytics/ResiliencyTestBase.java
michaelsembwever: yes, see CASSANALYTICS-112  (i'll offer a separate PR for it)

cassandra-six-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/SSTableCache.java
michaelsembwever: added `CompressionDictionaryTests.testCacheEvictionReleasesDictionaryReference()`

it needed package-private `evictCompressionMetadata(SSTable)` (the cache field is private and nothing could force an entry out), and `CompressionDictionaryTests.testCacheEvictionReleasesDictionaryReference`. 



see https://github.com/apache/cassandra-analytics/pull/236/changes/16d6c8fceb6fb3e535a66857e2a10f12175da2c9

cassandra-six-zero-bridge/src/main/java/org/apache/cassandra/spark/reader/SSTableReader.java
michaelsembwever: The dictionary itself is used twice before the constructor returns: CompressedRawInputStream.from(...) allocates through CompressionMetadata.compressor(), and skipFully(...) decompresses every chunk it skips. 



see https://github.com/apache/cassandra-analytics/pull/236/changes/16d6c8fceb6fb3e535a66857e2a10f12175da2c9

cassandra-six-zero-bridge/src/main/java/org/apache/cassandra/io/sstable/format/bti/BtiReaderUtils.java
michaelsembwever: is it needed…?

The five-zero test directory is included in the six-zero module.

see locally after a run `build/test-results/cassandra-six-zero-bridge/testBti/TEST-org.apache.cassandra.io.sstable.format.bti.BtiReaderUtilsTest.xml`

sarankk (approved):
Thanks Mick for adding 6.0 support!, looks good.

jyothsnakonisa (approved):
Thanks for resolving comments, looks good to me!

michaelsembwever:
Committed https://github.com/apache/cassandra-analytics/commit/9a2355882596793d106180309ae80db3b71450cb

PR #238, opened 2026-08-28
CASSANALYTICS-194: Parse <replicas>/<transient> replication factor fo…

mansikhara:
…r witness-enabled keyspaces

Cassandra accepts a replication factor of the form <replicas>/<transient>, which witness replicas under mutation tracking reuse, so a witness-enabled keyspace declares 'datacenter1': '3/1' meaning three replicas of which one is a witness.

CqlUtils.extractReplicationFactor() passed each datacenter value to Integer.parseInt, so a bulk read against any such keyspace failed during job setup with an uncaught NumberFormatException.

Transient counts are now tracked per datacenter alongside the existing totals and exposed through getFullReplicationFactor(), getTransientReplicationFactor(), getFullReplicas(dc), getTransientReplicas(dc) and hasTransientReplicas(). getTotalReplicationFactor() keeps its existing meaning of all replicas including witnesses, so behaviour for untracked keyspaces is unchanged. Parsing applies the same constraints Cassandra enforces in locator.ReplicationFactor.validate. A new parseStrict factory reports an unparseable or empty replication map at parse time rather than dropping the datacenter and failing later with a misleading "DC not found in replication factor"; the lenient constructor is retained unchanged for CDC callers. The Kryo serializer and CassandraRing's hand-rolled JDK readObject/writeObject are updated so the new field survives serialization to Spark executors.

Prerequisite for CASSANALYTICS-164.


cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
```diff
@@ -50,6 +51,11 @@
  *             "replication_factor" : 1
  *         }
  *     }
+ * <p>
+ * Replica counts may also use the {@code <replicas>/<transient>} form, e.g. {@code "DC1" : "3/1"}, meaning three
+ * replicas of which one is transient. Witness replicas under mutation tracking (CEP-45/CEP-46) reuse this form, so
```
sarankk: Nit: Transient replication was added before MT, it was updated to work on top of MT later. Also why do we use term reuse here? 
```suggestion
 * replicas of which one is transient. Witness replicas (CEP-46) reuse this form, so
```

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
```diff
@@ -108,11 +114,44 @@ public static ReplicationFactor simpleStrategy(int rf)
...
+    /**
+     * Per-datacenter count of transient (witness) replicas, parsed from the {@code <replicas>/<transient>} form.
+     * A datacenter absent from this map has no transient replicas. Always empty for untracked keyspaces using the
+     * plain {@code <replicas>} form, which keeps behaviour identical for those keyspaces.
+     */
+    @NotNull
+    private final Map<String, Integer> transientOptions;
 
+    /**
+     * Lenient parse: a replication value that cannot be parsed is logged and its datacenter omitted. Retained for
```
sarankk: In what scenario do we anticipate using lenient parse?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
```diff
@@ -108,11 +114,44 @@ public static ReplicationFactor simpleStrategy(int rf)
     private final ReplicationStrategy replicationStrategy;
     @NotNull
     private final Map<String, Integer> options;
+    /**
+     * Per-datacenter count of transient (witness) replicas, parsed from the {@code <replicas>/<transient>} form.
+     * A datacenter absent from this map has no transient replicas. Always empty for untracked keyspaces using the
+     * plain {@code <replicas>} form, which keeps behaviour identical for those keyspaces.
+     */
+    @NotNull
+    private final Map<String, Integer> transientOptions;
```
sarankk: Options hold `dc : 3` and transient options hold `dc : 1` for transient replication `dc: 3/1`. The maps could diverge. Shall we combine the 2 maps into 1 using `ReplicaCounts`?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
```diff
@@ -108,11 +114,44 @@ public static ReplicationFactor simpleStrategy(int rf)
...
+     */
+    public static ReplicationFactor parseStrict(@NotNull Map<String, String> options)
+    {
+        return new ReplicationFactor(options, true);
+    }
+
+    private ReplicationFactor(@NotNull Map<String, String> options, boolean strict)
     {
         this.replicationStrategy = ReplicationFactor.ReplicationStrategy.getEnum(options.get("class"));
         this.options = new LinkedHashMap<>(options.size());
```
sarankk: If we plan to merge the maps, shall we make this immutable ?

cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/data/ReplicationFactorTests.java
```diff
@@ -111,4 +118,296 @@ public void testEquality()
...
+    }
+
+    @Test
+    public void testTransientEqualToTotalIsRejected()
+    {
+        // Cassandra requires at least one full replica, so 3/3 is invalid
+        ReplicationFactor replicationFactor = new ReplicationFactor(ImmutableMap.of(
+        "class", "NetworkTopologyStrategy",
+        "datacenter1", "3/3"));
+        assertThat(replicationFactor.getOptions()).doesNotContainKey("datacenter1");
```
sarankk: We should throw in this scenario instead, if Cassandra does not allow it, Sidecar should fail as well.

cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/data/ReplicationFactorTests.java
```diff
@@ -111,4 +118,296 @@ public void testEquality()
...
+        assertThat(full.hashCode()).isNotEqualTo(withTransient.hashCode());
+    }
+
+    @Test
+    public void testNegativeReplicationFactorIsRejected()
+    {
+        ReplicationFactor replicationFactor = new ReplicationFactor(ImmutableMap.of(
+        "class", "NetworkTopologyStrategy",
+        "datacenter1", "-3"));
+        assertThat(replicationFactor.getOptions()).doesNotContainKey("datacenter1");
```
sarankk: Should we throw in this scenario as well?

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/utils/CqlUtilsTest.java
```diff
@@ -142,6 +142,84 @@ public void testExtractReplicationFactor(CassandraBridge bridge)
         assertThat(systemSchemaRf.getOptions()).isEqualTo(ImmutableMap.of());
     }
 
+    @Test
+    public void testExtractReplicationFactorWithWitnessReplicas()
+    {
+        // Witness-enabled keyspace as created by Cassandra's WitnessAlwaysReadsFullReplicaTest on the
+        // cep-45-mutation-tracking branch: the <replicas>/<transient> form plus replication_type = 'tracked'
+        String schema = "CREATE KEYSPACE witnessks WITH REPLICATION = {'class': 'NetworkTopologyStrategy', "
+                        + "'datacenter1': '3/1', 'datacenter2': '3/1'} AND replication_type = 'tracked' "
```
sarankk: `replication_type = 'tracked'` this field is not yet supported in Sidecar schema response, shall we remove this from example?

sarankk (commented):
Thanks Mansi, left few comments

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
mansikhara: Good catch, you are right that transient replication came first. Applied your suggestion, thanks!

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
mansikhara: None, as it turns out. I checked the callers and there are only two, ReplicationFactorSupplier and CdcOptions, and both build hardcoded maps that cannot fail to parse. So I have removed the lenient path entirely and made strict the only behaviour! Thanks for the find

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
mansikhara: Agreed and done, thanks!

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
mansikhara: Done!

cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/data/ReplicationFactorTests.java
mansikhara: Great suggestions! Done.

cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/data/ReplicationFactorTests.java
mansikhara: Done, thanks!

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/utils/CqlUtilsTest.java
mansikhara: You are right, I confirmed it. Removed it from these fixtures since it is irrelevant to replication factor parsing anyway.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/utils/CqlUtilsTest.java
```diff
@@ -142,6 +142,83 @@ public void testExtractReplicationFactor(CassandraBridge bridge)
         assertThat(systemSchemaRf.getOptions()).isEqualTo(ImmutableMap.of());
     }
 
+    @Test
+    public void testExtractReplicationFactorWithWitnessReplicas()
+    {
+        // The <replicas>/<transient> form, as Cassandra's WitnessAlwaysReadsFullReplicaTest creates it on the
+        // cep-45-mutation-tracking branch. replication_type is deliberately omitted: Sidecar builds its schema
+        // response from the driver's exportAsString(), which does not emit that property, and it is irrelevant to
+        // replication factor parsing in any case. extractReplicationType is covered separately.
```
sarankk: Nit: remove comment, `<replicas>/<transient>` existed prior to cep-45 branch and rest is not necessary

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
```diff
@@ -106,62 +110,204 @@ public static ReplicationFactor simpleStrategy(int rf)
...
+    /**
+     * Canonical constructor. {@code validate} exists only for {@link #readResolve()}, which must not reject an
+     * instance that an earlier version was willing to create.
+     */
+    private ReplicationFactor(@NotNull ReplicationStrategy replicationStrategy,
+                              @NotNull Map<String, ReplicaCounts> replication,
+                              boolean validate)
+    {
+        this.replicationStrategy = replicationStrategy;
+        if (validate)
```
sarankk: Can we delete this flag, it is always false

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
```diff
@@ -106,62 +110,204 @@ public static ReplicationFactor simpleStrategy(int rf)
...
+        this(replicationStrategy, options, Collections.emptyMap());
+    }
 
+    public ReplicationFactor(@NotNull ReplicationStrategy replicationStrategy,
+                             @NotNull Map<String, Integer> options,
+                             @NotNull Map<String, Integer> transientOptions)
+    {
+        this.replicationStrategy = replicationStrategy;
+        Map<String, ReplicaCounts> merged = new LinkedHashMap<>(options.size());
         for (Map.Entry<String, Integer> entry : options.entrySet())
```
sarankk: Nit: We seem to repeat these bits among constructors, existing code seems to have 3 constructors, with transient options we seem to have 4, could we restructure these?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/partitioner/CassandraRing.java
```diff
@@ -295,6 +322,13 @@ private void writeObject(ObjectOutputStream out) throws IOException, ClassNotFou
             out.writeUTF(option.getKey());
             out.writeByte(option.getValue());
         }
+        Map<String, Integer> transientOptions = this.replicationFactor.getTransientOptions();
```
sarankk: Why are we writing options and transient options separately? could we write `ReplicationFactor` itself since it is serializable?

sarankk (commented):
Thanks Mansi!, few more comments.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/utils/CqlUtilsTest.java
mansikhara: Done!

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
mansikhara: Done!

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
mansikhara: Done! Good eye, thanks!

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/partitioner/CassandraRing.java
mansikhara: Done, thanks!

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
```diff
@@ -106,62 +110,209 @@ public static ReplicationFactor simpleStrategy(int rf)
...
+    }
 
-        for (Map.Entry<String, Integer> entry : options.entrySet())
+    /**
+     * @return the number of replicas across all datacenters that hold the full data set, i.e. the total
+     *         replication factor minus transient (witness) replicas
+     */
+    public Integer getFullReplicationFactor()
+    {
+        return replication.values().stream().mapToInt(ReplicaCounts::fullReplicas).sum();
```
sarankk: Nit: Shall we precompute these values in constructor and return here

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
```diff
@@ -106,62 +110,209 @@ public static ReplicationFactor simpleStrategy(int rf)
...
+    {
+        ReplicaCounts counts = replication.get(datacenter);
+        if (counts == null)
         {
-            if ("class".equals(entry.getKey()))
-            {
-                continue;
-            }
-            this.options.put(entry.getKey(), entry.getValue());
+            throw new IllegalArgumentException(String.format("Datacenter %s not found in replication factor %s",
```
sarankk, line 304: We seem to throw this only for full replica count extraction and not for transient replica, any reason why?

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
```diff
@@ -106,62 +110,209 @@ public static ReplicationFactor simpleStrategy(int rf)
...
 
+    /**
+     * @return per-datacenter total replica counts, including transient (witness) replicas. Unmodifiable, and derived
+     *         from {@link #getReplication()}.
+     */
     @NotNull
     public Map<String, Integer> getOptions()
     {
-        return options;
+        Map<String, Integer> options = new LinkedHashMap<>(replication.size());
```
sarankk: Similarly shall we precompute these and return here, instead of compute on every call

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
```diff
@@ -106,62 +110,209 @@ public static ReplicationFactor simpleStrategy(int rf)
...
+    }
+
+    /**
+     * @return per-datacenter transient (witness) replica counts, unmodifiable. Datacenters without transient
+     *         replicas are absent.
+     */
+    @NotNull
+    public Map<String, Integer> getTransientOptions()
+    {
+        Map<String, Integer> transientOptions = new LinkedHashMap<>();
```
sarankk: Same as above

sarankk (commented):
Changes look good

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
mansikhara: Done!

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
mansikhara: No reason, was an oversight, thanks for catching it :)

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
mansikhara: Done!

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
mansikhara: Done!

sarankk (approved):
+1 Thanks Mansi, LGTM

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
```diff
@@ -188,13 +359,173 @@ public boolean equals(Object other)
...
+
+        /**
+         * Mirrors the constraints Cassandra enforces in {@code ReplicationFactor.validate}: transient replicas must
+         * be non-negative and strictly fewer than the total, so at least one full replica always exists.
+         *
+         * @param datacenter        datacenter name for the error message, may be {@code null}
+         * @param allReplicas       total replicas
+         * @param transientReplicas transient (witness) replicas
+         */
+        static void validate(String datacenter, int allReplicas, int transientReplicas)
```
jyothsnakonisa: ReplicaCounts class doesn't need `datacenter` at all and the datacenter argument fit here in this method. I see that you are passing this for logging can u try moving logging out of this class

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
```diff
@@ -188,13 +359,173 @@ public boolean equals(Object other)
...
+        }
+
+        /**
+         * @param datacenter        datacenter name, used only in error messages, may be {@code null}
+         * @param allReplicas       total replicas
+         * @param transientReplicas transient (witness) replicas
+         * @return validated counts
+         * @throws IllegalArgumentException when the counts are inconsistent
+         */
+        public static ReplicaCounts of(String datacenter, int allReplicas, int transientReplicas)
```
jyothsnakonisa: ReplicaCounts class doesn't need datacenter at all and the datacenter argument fit here in this method. I see that you are passing this for logging can u try moving logging out of this class

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
```diff
@@ -203,25 +534,29 @@ public static class Serializer extends com.esotericsoftware.kryo.Serializer<Repl
...
             out.writeByte(replicationFactor.replicationStrategy.value);
-            out.writeByte(replicationFactor.options.size());
-            for (Map.Entry<String, Integer> entry : replicationFactor.options.entrySet())
+            out.writeByte(replicationFactor.replication.size());
+            for (Map.Entry<String, ReplicaCounts> entry : replicationFactor.replication.entrySet())
             {
                 out.writeString(entry.getKey());
-                out.writeByte(entry.getValue());
+                out.writeByte(entry.getValue().allReplicas());
+                out.writeByte(entry.getValue().transientReplicas());
```
jyothsnakonisa: This adds a new field to the Kryo wire format (1 byte/DC → 2 bytes/DC) with no version marker, so an old-format ReplicationFactor read by this new code would crash or silently produce wrong RF values. 

CassandraRing solves this exact problem for its own serialization with a version byte, and this is actually the primary path Spark uses (its JDK fallback is just a fallback), so it seems worth the same protection here.

bbotella (approved):
+1

cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/data/ReplicationFactorTests.java
```diff
@@ -111,4 +118,357 @@ public void testEquality()
...
+        ReplicationFactor replicationFactor = new ReplicationFactor(ImmutableMap.of(
+        "class", "NetworkTopologyStrategy",
+        "datacenter1", "3",
+        "datacenter2", "0"));
+        assertThat(replicationFactor.getOptions().get("datacenter2")).isEqualTo(Integer.valueOf(0));
+        assertThat(replicationFactor.getTotalReplicationFactor()).isEqualTo(3);
+    }
+
+    @Test
+    public void testKryoSerializationRoundTripWithTransientReplicas() throws Exception
```
jyothsnakonisa: This test is almost identical to testPrecomputedValuesSurviveKryoSerialization, WDYT of merging them?

cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/data/ReplicationFactorTests.java
```diff
@@ -111,4 +118,357 @@ public void testEquality()
...
+
+        assertThat(deserialized).isEqualTo(original);
+        assertThat(deserialized.getTotalReplicationFactor()).isEqualTo(6);
+        assertThat(deserialized.getFullReplicationFactor()).isEqualTo(5);
+        assertThat(deserialized.getTransientReplicas("datacenter1")).isEqualTo(1);
+        assertThat(deserialized.getTransientReplicas("datacenter2")).isEqualTo(0);
+    }
+
+    @Test
+    public void testJdkSerializationRoundTripWithTransientReplicas() throws Exception
```
jyothsnakonisa: Same here, this test Is identical to testPrecomputedValuesSurviveJdkSerialization, WDYT of merging them

jyothsnakonisa (commented):
Looks good, left some minor comments

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
mansikhara: Agreed, done!

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
mansikhara: Done!

cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/data/ReplicationFactorTests.java
mansikhara: Merged, thanks!

cassandra-analytics-common/src/test/java/org/apache/cassandra/spark/data/ReplicationFactorTests.java
mansikhara: Merged, thanks!

cassandra-analytics-common/src/main/java/org/apache/cassandra/spark/data/ReplicationFactor.java
mansikhara: Done, thanks!
But, practically the exposure is narrow: driver and executors always load the same jar, so old and new bytes can't meet inside a job, and the RF isn't persisted between jobs. But it's one byte and it makes the next format change safe, so I've added it here!

jyothsnakonisa (approved):
LGTM!

PR #239, opened 2026-08-31
CASSANALYTICS-104: Eliminate redundant filesystem lookups in SSTable streaming

lukasz-antoniak:
Fixes [CASSANALYTICS-104](https://issues.apache.org/jira/browse/CASSANALYTICS-104).

### DIFF
--- a/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
+++ b/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
@@ -21,13 +21,12 @@ package org.apache.cassandra.spark.bulkwriter;
 
 import java.io.IOException;
 import java.math.BigInteger;
-import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
@@ -42,8 +41,6 @@ import org.slf4j.LoggerFactory;
 import org.apache.cassandra.bridge.SSTableDescriptor;
 import org.apache.cassandra.spark.bulkwriter.token.ReplicaAwareFailureHandler;
 import org.apache.cassandra.spark.common.Digest;
-import org.apache.cassandra.spark.common.SSTables;
-import org.apache.cassandra.spark.data.FileType;
 import org.apache.cassandra.util.IntWrapper;
 
 public class DirectStreamSession extends StreamSession<TransportContext.DirectDataBulkWriterContext>
@@ -85,20 +82,18 @@ public class DirectStreamSession extends StreamSession<TransportContext.DirectDa
                 // 2. validate the sstables
                 // 3. send the sstables to all replicas
                 // 4. remove the sstables once sent
-                Map<Path, Digest> fileDigests = sstableWriter.prepareSStablesToSend(writerContext, sstables);
-                // retain only the SSTable data components
+                SortedSSTableWriter.PreparedSSTables preparedSSTables = sstableWriter.prepareSStablesToSend(writerContext, sstables);
                 IntWrapper sstableCounter = new IntWrapper();
-                fileDigests.keySet()
-                           .stream()
-                           .filter(p -> p.getFileName().toString().endsWith(FileType.DATA.getFileSuffix()))
-                           .forEach(sstable -> {
-                               sstableCounter.value++;
-                               sendSStableToReplicas(sstable);
-                           });
+                preparedSSTables.sstables()
+                                .forEach(preparedSSTable -> {
+                                    sstableCounter.value++;
+                                    sendSStableToReplicas(preparedSSTable);
+                                });
 
                 LOGGER.info("[{}]: Sent newly produced SSTables. sstables={}", sessionID, sstableCounter.value);
-                LOGGER.info("[{}]: Removing temporary files after streaming. files={}", sessionID, fileDigests);
-                fileDigests.keySet().forEach(path -> {
+                Set<Path> allSSTableFiles = preparedSSTables.files();
+                LOGGER.info("[{}]: Removing temporary files after streaming. files={}", sessionID, allSSTableFiles);
+                allSSTableFiles.forEach(path -> {
                     try
                     {
                         Files.deleteIfExists(path);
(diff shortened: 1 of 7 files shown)


cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
```diff
@@ -182,57 +171,59 @@ protected void sendRemainingSSTables()
...
-            failureHandler.addFailure(this.tokenRange, replica, exception.getMessage());
-            errors.add(new StreamError(this.tokenRange, replica, exception.getMessage()));
+            // Sometimes error can contain just file name (e.g. when it is missing).
+            // Let us return 3 latest stacktrace lines for easier troubleshooting.
+            String stackTrace = Arrays.stream(exception.getStackTrace())
+                                      .limit(3)
+                                      .map(StackTraceElement::toString)
+                                      .collect(Collectors.joining("\n"));
+            String errorMessage = exception.getClass().getName() + ": " + exception.getMessage()
+                                  + "\n" + String.join("\n", stackTrace);
```
yifan-c: `String.join("\n", stackTrace)`  is equivalent to just `stackTrace`, right? 
The stacktrace are already aggregated in the statement above.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
```diff
@@ -85,20 +82,19 @@ protected void onSSTablesProduced(Set<SSTableDescriptor> sstables)
                 // 2. validate the sstables
                 // 3. send the sstables to all replicas
                 // 4. remove the sstables once sent
-                Map<Path, Digest> fileDigests = sstableWriter.prepareSStablesToSend(writerContext, sstables);
+                SortedSSTableWriter.PreparedSSTables preparedSSTables = sstableWriter.prepareSStablesToSend(writerContext, sstables);
                 // retain only the SSTable data components
```
yifan-c: Please remove this comment, it is stale with the removal of `.filter`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
```diff
@@ -153,22 +149,15 @@ protected StreamResult doFinalizeStream()
     @Override
     protected void sendRemainingSSTables()
     {
-        try (DirectoryStream<Path> dataFileStream = Files.newDirectoryStream(sstableWriter.getOutDir(), "*Data.db"))
+        try
         {
-            for (Path dataFile : dataFileStream)
-            {
-                if (isFileStreamed(dataFile))
```
yifan-c, line 160: `StreamSession.isFileStreamed` is now dead. Can you remove the code in the base class?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
yifan-c: The effect of this method has no use. It can be deleted, along with `isFileStreamed`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
```diff
@@ -182,57 +171,59 @@ protected void sendRemainingSSTables()
...
-                                         Map<Path, Digest> fileHashes) throws IOException
+                                         RingInstance instance) throws IOException
     {
-        try (DirectoryStream<Path> componentFileStream = Files.newDirectoryStream(dataFile.getParent(),
-                                                                                  SSTables.getSSTableBaseName(dataFile) + "*"))
+        for (Path componentFile : preparedSSTable.files())
         {
-            for (Path componentFile : componentFileStream)
+            // send data component the last
+            if (preparedSSTable.dataFile().equals(componentFile))
```
yifan-c: Can `dataFile()` return null?

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -438,4 +437,107 @@ public Map<Path, Digest> fileDigestMap()
...
+    {
+        return remainingSSTablesAfterClose;
+    }
+
+    /**
+     * Helper class representing list of newly generated sstables.
+     */
+    public static class PreparedSSTables
+    {
+        private static final PreparedSSTables EMPTY = new PreparedSSTables();
```
yifan-c: `EMPTY` is mutable. Caller can call `addIfAbsent` on it and mess up the state.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -438,4 +437,107 @@ public Map<Path, Digest> fileDigestMap()
...
+
+        private final Map<String, PreparedSSTable> sstables = new HashMap<>(); // indexed by base file name
+
+        public PreparedSSTable addIfAbsent(Path path)
+        {
+            String baseName = SSTables.getSSTableDescriptor(path).baseFilename;
+            return sstables.computeIfAbsent(baseName, (__) -> new PreparedSSTable());
```
yifan-c: Based on the impl, `getOrPrepareSSTable` method name fits better.

yifan-c (commented):
Please update  the `CHANGES.txt`

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
lukasz-antoniak: Can you recheck now? I did not think of an easier way to fix this issue.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/DirectStreamSession.java
lukasz-antoniak: Theoretically it could, but only if data file was not generated during sstable creation (e.g. `remainingSSTablesAfterClose` contained some sstable files, but without `Data.db`). I have replaced the operands in equality check, and added preconditions check.

cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/SortedSSTableWriter.java
```diff
@@ -289,19 +295,31 @@ public synchronized void close(BulkWriterContext writerContext) throws IOExcepti
...
+                overallFileDigests.put(path, digest);
+                newlyHashedFiles.add(path);
+
+                prepared.getOrPrepareSSTable(path)
+                        .addComponent(path, digest);
+
+                if (isDataFile(path))
+                {
+                    sstableCount += 1;
+                }
```
yifan-c: nit: how about adding the check outside of the for-loop to assert that all `PreparedSSTable` have non-null data file in `prepared`. 
Then, the if statement can be replaced with `sstableCount += prepared.sstables().size();`. 

The same applies to `prepareSStablesToSend`

yifan-c (approved):

PR #240, opened 2026-09-02
CASSANALYTICS-195: Make QuickTheories tests reproducible from a single seed

yifan-c:
Introduces QTRandom to bridge QT's RandomnessSource into Java Random API

Patch by Yifan Cai; Reviewed by TBD for CASSANALYTICS-195

### DIFF
--- a/cassandra-analytics-cdc/src/test/java/org/apache/cassandra/cdc/test/CdcTester.java
+++ b/cassandra-analytics-cdc/src/test/java/org/apache/cassandra/cdc/test/CdcTester.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Random;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
@@ -102,6 +103,7 @@ public class CdcTester
     Partitioner partitioner = Partitioner.Murmur3Partitioner;
     CdcOptions cdcOptions;
     CassandraSource cassandraSource;
+    public final Random random;
 
     CdcTester(CassandraBridge bridge,
               CdcBridge cdcBridge,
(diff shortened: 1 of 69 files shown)


cassandra-bridge/src/testFixtures/java/org/apache/cassandra/spark/utils/test/QTRandom.java
```diff
@@ -0,0 +1,48 @@
...
+    {
+        this.source = source;
+    }
+
+    @Override
+    protected int next(int bits)
+    {
+        return (int) source.next(Constraint.between(0, (1L << bits) - 1));
+    }
+}
```
yifan-c, line 39: This class is the core of the patch. The other test code are modified to take `QTRandom` instead of `Random`.

cassandra-analytics-cdc/src/test/java/org/apache/cassandra/cdc/test/CdcTester.java
```diff
@@ -153,6 +157,7 @@ public static class Builder
         private boolean shouldCdcEventWriterFailOnProcessing = false;
         private CdcOptions cdcOptions;
         private CassandraSource cassandraSource = CassandraSource.DEFAULT;
+        private Random random = new Random();
```
bbotella, line 160: Nit: Maybe moving this inside the builder to avoid instantiating Random if it is not needed? Just like we do with CdcOptions.

cassandra-analytics-core/src/test/java/org/apache/cassandra/spark/utils/test/QTRandomTest.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    @Test
+    public void differentSeedDrawsProduceDifferentOutput()
+    {
+        Random first = new QTRandom(new FixedSequenceSource(1L));
+        Random second = new QTRandom(new FixedSequenceSource(2L));
+
+        assertThat(first.nextLong()).isNotEqualTo(second.nextLong());
```
bbotella, line 54: This is "technically" flaky :-P

But let's keep it as is. It would be REAL bad luck to hit it!

PR #241, opened 2026-09-10
CASSANALYTICS-196: Skip CDC-irrelevant mutations on deserialization failure

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSANALYTICS-196

### DIFF
--- a/cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/db/commitlog/BufferingCommitLogReader.java
+++ b/cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/db/commitlog/BufferingCommitLogReader.java
@@ -51,6 +51,9 @@ import org.apache.cassandra.io.util.DataInputBuffer;
 import org.apache.cassandra.io.util.FileDataInput;
 import org.apache.cassandra.io.util.RandomAccessReader;
 import org.apache.cassandra.io.util.RebufferingInputStream;
+import org.apache.cassandra.schema.Schema;
+import org.apache.cassandra.schema.TableId;
+import org.apache.cassandra.schema.TableMetadata;
 import org.apache.cassandra.spark.exceptions.TransportFailureException;
 import org.apache.cassandra.spark.utils.AsyncExecutor;
 import org.apache.cassandra.spark.utils.LoggerHelper;
@@ -567,6 +570,16 @@ public class BufferingCommitLogReader implements CommitLogReadHandler,
         catch (Throwable t)
         {
             JVMStabilityInspector.inspectThrowable(t);
+
+            if (failedMutationHasNoCdcTable(inputBuffer, size))
+            {
+                // No CDC-enabled table is involved – safe to skip and move on
+                logger.trace("Ignoring mutation that failed to deserialize; no CDC-enabled table involved", "error", t);
+                stats.mutationsDeserializeFailedNonCdcCount(1);
+
+                return;
+            }
+
             Path p = Files.createTempFile("mutation", "dat");
 
             try (DataOutputStream out = new DataOutputStream(Files.newOutputStream(p)))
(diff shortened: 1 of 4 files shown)


cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/db/commitlog/BufferingCommitLogReader.java
```diff
@@ -604,6 +617,45 @@ private void readMutationInternal(byte[] inputBuffer,
...
+        {
+            int updateCount = (int) in.readUnsignedVInt();
+            TableId tableId = TableId.deserialize(in);
+            TableMetadata metadata = Schema.instance.getTableMetadata(tableId);
+            if (metadata != null && metadata.params.cdc)
+            {
+                return false;
+            }
+
+            if (updateCount > 1)
```
bbotella, line 645: Can't we check this *before* we need to extract the tableId and the metadata? (just at line 638)

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/db/commitlog/BufferingCommitLogReader.java
```diff
@@ -567,6 +570,16 @@ private void readMutationInternal(byte[] inputBuffer,
         catch (Throwable t)
         {
             JVMStabilityInspector.inspectThrowable(t);
+
+            if (failedMutationHasNoCdcTable(inputBuffer, size))
```
bbotella, line 574: It is scary that this new behavior doesn't break a pre-existing test :-(

cassandra-four-zero-bridge/src/test/java/org/apache/cassandra/db/commitlog/BufferingCommitLogReaderTableContextTests.java
```diff
@@ -0,0 +1,129 @@
...
+                                                                                                     corruptedMutation.length);
+
+        assertThat(failedMutationHasNoCdcTable).isFalse();
+    }
+
+    @Test
+    public void testFailedMutationHasNoCdcTableWhenTableIdUnresolvable() throws Exception
+    {
+        CassandraBridgeImplementation.setup();
+        TableId unknownTableId = TableId.fromUUID(UUID.randomUUID());
```
bbotella, line 70: nit: Maybe move these two to a helper getUnregisteredTable method?

cassandra-four-zero-bridge/src/main/java/org/apache/cassandra/db/commitlog/BufferingCommitLogReader.java
```diff
@@ -604,6 +617,45 @@ private void readMutationInternal(byte[] inputBuffer,
...
+     * If a mutation has more than one partition update, it conservatively assumes a CDC-enabled table is involved.
+     *
+     * @param inputBuffer raw byte array w/Mutation data
+     * @param size        serialized size of the mutation
+     * @return false if the first table touched by the mutation is confirmed CDC-enabled, or the mutation has more
+     *         than one {@link PartitionUpdate} (so the rest can't be safely checked); true otherwise, including
+     *         when the table's metadata can't be found, or even the first {@link TableId} can't be read
+     */
+    @VisibleForTesting
+    static boolean failedMutationHasNoCdcTable(byte[] inputBuffer, int size)
```
bbotella, line 633: Maybe it is just me, but I tend to not like the negative on method names. HasNo + false is what we are looking for here -> that double negative always messes my head :-P

What do you think on having a failedMutationHasCdcTable method instead that returns true if it needs to be handled, and false if it needs to be ignored?

Obviously not a blocker.

jyothsnakonisa (approved):
Looks good to me!

PR #245, opened 2026-09-17
CASSANALYTICS-198: Add publish script and workflow

bbotella:
CASSANALYTICS-198


jyothsnakonisa:
Thanks @bbotella for the patch, does each snapshot publishing after a commit overrides existing artifact for that version in the repository or does it create a new artifact?

bbotella:
The only thing that should be overwritten is the -SNAPSHOT pointer, but the artifact itself will not be overwritten. That should allow others to point to snapshot artifact and always get the latest commit to trunk.

jyothsnakonisa:
Since each commit creates a new artifact, I am wondering if it creates any storage space issues in the artifactory

bbotella:
I've seen a similar pattern on other Apache projects. We can start with this and fix it if it becomes an issue

michaelsembwever:
you can use SNAPSHOT or the timestamped snapshot version. 

> I am wondering if it creates any storage space issues in the artifactory 

we will get pinged if it becomes a problem

CHANGES.txt
```diff
@@ -1,5 +1,6 @@
 0.5.0
 -----
+ * Publish a snapshot artifact for every trunk commit (CASSANALYTICS-198)
```
michaelsembwever: nit:  this isn't user-facing and shouldn't be in CHANGES.txt

michaelsembwever (approved):

PR #246, opened 2026-09-18
CASSANALYTICS-112 Support integration tests after Transactional Cluster Metadata

michaelsembwever:
Select topology-change hooks for legacy Cassandra and Transactional Cluster Metadata. Create schema before supported transitions and configure three metadata replicas on Cassandra 6.0 and later.

Retain single-node success and failure scenarios, including multi-datacenter coverage, across supported versions. Preserve concurrent joins and leaves and cluster-doubling/halving scenarios for Cassandra older than 6.0, with their required range-movement settings scoped to those tests and restored during teardown.

Exercise concurrent TCM joins and leaves with Sidecar's CASSSIDECAR-277 disjoint token placement, including a join in an unreplicated datacenter. Preserve the existing consistency-level and per-node data checks.

Wait for joining and replacement startup before shutdown, require successful joins to reach Normal, and await decommission results before validating writes. Inject terminal leave failures and verify the expected error. Add completion regression tests and raise the integration JVM datagram-socket limit for multi-node tests.


bbotella (commented):
Quick (probably dumb) question. Should we have Cassandra 6.0 entries here?
https://github.com/apache/cassandra-analytics/blob/trunk/.github/workflows/test.yaml#L247

